PR #86, opened 2017-03-10
Add paging to user preference database access code to avoid postgres …

jsharkey13:
…errors, see ucam-cl-dtg/isaac-app#823

Some sanity checking of the paging would be good!

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/dos/PgUserPreferenceManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/dos/PgUserPreferenceManager.java
@@ -32,6 +32,8 @@ import java.sql.SQLException;
 import java.util.List;
 import java.util.Map;
 
+import static java.lang.Math.min;
+
 /**
  *  A postgres specific User Preference Manager
  */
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dos/PgUserPreferenceManager.java
```diff
@@ -86,37 +88,48 @@ public UserPreference getUserPreference(String preferenceType, String preference
...
+                pst.setString(pagedUsers.size() + 1, preferenceType);
+                pst.setString(pagedUsers.size() + 2, preferenceName);
 
-        } catch (SQLException e) {
-            throw new SegueDatabaseException("Postgres exception", e);
+                ResultSet results = pst.executeQuery();
+
+                while (results.next()) {
+                    Long userId = results.getLong("user_id");
+                    UserPreference pref = new UserPreference(userId, results.getString("preference_type"),
```
sacummins: Maybe the construction of these objects could be refactored out in to a common method? Just all these repeated string literals could cause some problems later if they need to change.

PR #88, opened 2017-04-28
Stop the log_events table from storing null valued fields in its event_details JSON.

mlt47:
Issue https://github.com/ucam-cl-dtg/isaac-app/issues/790.

This would stop the log_events table from storing null valued fields in event_details **for every event_type, not just 'Answer Question'**.
From skimming through non-'Answer Question' log events, it seems like the other event_types which store null values are QUESTION_ATTEMPT_RATE_LIMITED and USER_SCHOOL_CHANGE.

The only event_type which seems to me that might want to usefully store null values is the EQN_EDITOR_LOG @Morpheu5. It does not seem to do so, so I think this change is fine.


jsharkey13:
This is strictly an improvement, and so I'm happy to merge it - but it doesn't address the unnecessary `tags` and `published` JSON keys. Perhaps they're not worth it!

PR #92, opened 2017-05-16
backend password validation checking for min password length

dunderwood88:
Simple backend validation for minimum password length enforcement.


jsharkey13:
Does this cover changing passwords whilst resetting your password? That needs to be addressed too!

PR #94, opened 2017-05-25
Log Events Cleanup

jsharkey13:
Make log events log valid JSON, contain actually useful information, and use a consistent set of JSON keys when referring to the same global property.

### DIFF
--- a/src/main/resources/db_scripts/log_cleanup.sql
+++ b/src/main/resources/db_scripts/log_cleanup.sql
@@ -1,0 +1,23 @@
+﻿/*
+  List all top-level keys used in logs:
+*/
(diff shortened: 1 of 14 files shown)


src/main/resources/db_scripts/log_cleanup.sql
```diff
@@ -0,0 +1,23 @@
+﻿/*
+  List all top-level keys used in logs:
+*/
+--SELECT DISTINCT jsonb_object_keys(event_details) as keys FROM logged_events --WHERE event_type='VIEW_PAGE';--event_details::TEXT LIKE '{%';
```
sacummins: Do you still need this line? If not may as well remove.

PR #96, opened 2017-06-01
Feature/return question part status information

mlt47:
First medium sized pull request to the API, so please talk me through any recommendations that you have so that I can match the codebase/Java's conventions and design decisions in the future.

Physicists want this information so that we can display question part progress as an extra column in the  Assignment Progress. This change sends the necessary information to the front-end and, as a bonus, calculates a better progress percentage for the My Boards page which uses question part progress rather than just question progress.


mlt47:
In trying to display the information I noticed that the Assignment Progress uses another endpoint to get information so I need to alter that before this pull request is ready to be merged.

PR #102, opened 2017-06-30
generalised user preference fetching

dunderwood88:
What do people think of this? (Not quite finished)

Assuming the email preferences are migrated to the user_preference table...

I've changed the "subject interest" end point to a "user preference" endpoint, which fetches all user preferences and returns them all in a json blob. Therefore, instead of having two separate endpoints for email and subject preferences, they are returned together and filtered on the front end. This removes the Isaac specific element of the subject interests.

Previously updates to subject preferences where filtered by a hard-coded "acceptedSubjects" list. Perhaps we could filter *all* accepted user preferences by having them in a config file maybe?

This would solve both issues 845 and 846.


dunderwood88:
p.s. see this branch on the front end to see changes to handling:

https://github.com/ucam-cl-dtg/isaac-app/commit/093ba21ce9d2f6209c37064b07d6b9d6fd1b2c8b

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -614,16 +614,15 @@ public Response getAllSchoolOtherResponses(@Context final Request request) {
...
+     * @param userPreferences
+     * 			  - the preferences for this user
      * @return the updated user object.
      * @throws NoCredentialsAvailableException
      * @throws IncorrectCredentialsProvidedException
      */
     private Response updateUserObject(final HttpServletRequest request, final HttpServletResponse response,
                                       final RegisteredUser userObjectFromClient, final String passwordCurrent,
-                                      final List<IEmailPreference> emailPreferences, final Map<String, Boolean> subjectInterests)
+                                      final Map<String, Map<String, Boolean>> userPreferences)
```
sacummins: This looks good for now, but I would like to see the updateUserObject method simplified in the future. As it isn't just the user object now. Perhaps that will be my next mission..

jsharkey13: I agree, and https://github.com/ucam-cl-dtg/isaac-app/issues/813 is related too.

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -709,30 +708,23 @@ private Response updateUserObject(final HttpServletRequest request, final HttpSe
                 }
             }
 
-            // Now update the email preferences
-            emailPreferenceManager.saveEmailPreferences(userObjectFromClient.getId(), emailPreferences);
+            if (userPreferences != null) {
 
-            //----------------------------------------------------------------------------------------------------------
-            // FIXME - the code between the dashed lines should be refactored out of this class; it does not belong here!
```
jsharkey13: Where has the validation code gone to? This was the only thing preventing people from adding random junk to the database directly . . .

dunderwood88: My solution to this would be to have user preferences with their allowed values to be held in a config file somewhere, rather than hardcoding them in. I would still need to implement this if we all agree its the best solution.

dunderwood88:
I've added the necessary validation code to this now. Sorry this took so long, Kafka took over my life...

jsharkey13:
If the files look weird, try ignoring whitespace changes: https://github.com/ucam-cl-dtg/isaac-api/pull/102/files?w=1

jsharkey13:
@sacummins - could you review this again; it didn't actually do what it was supposed to last time, and I have made drastic changes to ensure that user preferences are used everywhere and email preferences are fully purged.

jsharkey13:
See the app pull too: https://github.com/ucam-cl-dtg/isaac-app/pull/907

config-templates/windows--local-dev-segue-config.properties
```diff
@@ -92,3 +92,6 @@ TWITTER_CALLBACK_URI=http://localhost:8000/auth/twitter/callback
 
 # IPAddress GeoCoding
 IP_INFO_DB_API_KEY=[enter_key_here]
+
+# User Preferences
+SUBJECT_INTEREST=PHYSICS_UNI,PHYSICS_ALEVEL,PHYSICS_GCSE,CHEMISTRY_UNI,CHEMISTRY_ALEVEL,CHEMISTRY_GCSE,MATHS_UNI,MATHS_ALEVEL,MATHS_GCSE
```
sacummins: I wonder if it would be sensible to do the same thing as above with the other user preferences e.g. emails? We could take a hybrid approach of listing email preferences in a config file and hard coding the mandatory ones e.g. SYSTEM.

Just a suggestion, but I realise that the email stuff is still a bit horrid and probably not a massive priority to fix.

src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
```diff
@@ -422,10 +382,17 @@ public Response generateEmailVerificationToken(@PathParam("email") final String
     @GZIP
     public final Response sendEmails(@Context final HttpServletRequest request,
 		    		@PathParam("contentid") final String contentId, 
-		    		@PathParam("emailtype") final Integer emailTypeInt, 
+		    		@PathParam("emailtype") final String emailTypeString,
```
sacummins: Can we just stick the enum type in instead of a String. Not sure if RestEasy will map it for us (or what happens if they provide something that isn't a valid enum type - presumably a standard error?) Happy either way, just a point of interest

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -281,27 +276,28 @@ public Response createOrUpdateUserSettings(@Context final HttpServletRequest req
...
+     * @return user preference map
      */
     @GET
-    @Path("users/subject_interests")
+    @Path("users/user_preferences")
     @Produces(MediaType.APPLICATION_JSON)
     @GZIP
-    public Response getSubjectInterests(@Context final HttpServletRequest httpServletRequest) {
-        // FIXME - this endpoint does not belong here, and should probably return actual user preferences eventually.
+    public Response getUserPreferences(@Context final HttpServletRequest httpServletRequest) {
```
sacummins: How often do we want to request all user preferences in one go as opposed to just those of a certain type? Do we want to allow for a filter param?

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -852,4 +826,65 @@ private Response createUserObjectAndLogIn(final HttpServletRequest request, fina
...
+    private List<UserPreference> userPreferenceObjectToList(final Map<String, Map<String, Boolean>> userPreferenceObject, final long userId) {
+        List<UserPreference> userPreferences = Lists.newArrayList();
+        if (null == userPreferenceObject) {
+            return userPreferences;
+        }
+        // FIXME: This entire method is horrible, but required to sanitise what is stored in the database . . .
+        for (String preferenceType: userPreferenceObject.keySet()) {
+
+            // Check if the given preference type is one we support:
+            if (!EnumUtils.isValidEnum(IsaacUserPreferences.class, preferenceType)
```
sacummins: We could make this slightly nicer by dependency injecting all the enums we should support into a single list. This will also remove the need for specifying isaac specific classes in the code. This is fine for now though.

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -852,4 +826,65 @@ private Response createUserObjectAndLogIn(final HttpServletRequest request, fina
...
+                    && !EnumUtils.isValidEnum(SegueUserPreferences.class, preferenceType)) {
+                log.warn("Unknown user preference type '" + preferenceType + "' provided. Skipping.");
```
sacummins: Arguably, if they provide us an invalid preference we should probably tell them and even throw an exception?

src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
jsharkey13: You get really awful unhandled (and unhandleable) errors if it doesn't map exactly, which is why I ended up with it this way round. I did try it that way at first!

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
jsharkey13: Doing so breaks the whole atomicity thing: https://github.com/ucam-cl-dtg/isaac-app/issues/813 and really makes a mess when things go wrong. Failing silently isn't great, but at least it logs to the server logs, so we can ensure the front-end is doing the right thing.

PR #103, opened 2017-07-14
Feature/question parts for assignment progress and board completion

mlt47:
Uses question parts to calculate board completion % everywhere (my boards, my assignments, assignment progress, CSV downloads).

AssignmentFacade.getAssignmentProgress(...) now also returns a list of the number of question parts correctly answered as well as the pass status of each question. This change is backwards compatible, but in the future it might be nice to change it into one list of objects.


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -958,102 +956,93 @@ private void permanentlyStoreGameboard(final GameboardDTO gameboardToStore) thro
...
-                // for this question to determine if there is a
-                // correct answer somewhere.
-                for (int i = questionAttempts.size() - 1; i >= 0; i--) {
-                    if (questionAttempts.get(i).isCorrect() != null && questionAttempts.get(i).isCorrect()) {
-                        foundCorrectForThisQuestion = true;
-                        break;
+        Validate.notNull(gameItem, "gameItem cannot be null");
+        Validate.notNull(questionAttemptsFromUser, "questionAttemptsFromUser cannot be null");
+
+        int questionPartsCorrect = 0, questionPartsIncorrect = 0, questionPartsNotAttempted = 0;
```
jsharkey13: I think this would fail Steve's checkstyle!

jsharkey13:
But I still would like to fix the checkstyle one before merging!

PR #112, opened 2017-09-07
Move logic out of the DTO and into a manager

mlt47:


sacummins:
Seems sensible to me as it is more consistent with the way we manage everything else.

PR #116, opened 2017-09-28
Add IP based rate limiting

jsharkey13:
After seeing a rise in misuse of our API by scripts which ignore cookies, add rate limiting on a per-IP address basis to prevent abuse of our services.
This will not affect a malicious or dedicated attacker, but will prevent basic scripts from attempting to guess answers. 
For further enhancements, we could treat users with verified email addresses differently and reduce the per IP limits too. Adding [rate limiting](https://www.nginx.com/blog/rate-limiting-nginx/) to our [nginx router](https://github.com/ucam-cl-dtg/isaac-router) would be the best method for a dedicated attack or persistent abuse.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
@@ -29,6 +29,7 @@ import uk.ac.cam.cl.dtg.segue.api.managers.SegueResourceMisuseException;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
 import uk.ac.cam.cl.dtg.segue.api.monitors.AnonQuestionAttemptMisuseHandler;
 import uk.ac.cam.cl.dtg.segue.api.monitors.IMisuseMonitor;
+import uk.ac.cam.cl.dtg.segue.api.monitors.IPQuestionAttemptMisuseHandler;
 import uk.ac.cam.cl.dtg.segue.api.monitors.QuestionAttemptMisuseHandler;
 import uk.ac.cam.cl.dtg.segue.dao.ILogManager;
 import uk.ac.cam.cl.dtg.segue.dao.content.ContentManagerException;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
```diff
@@ -196,13 +199,26 @@ public Response answerQuestion(@Context final HttpServletRequest request,
...
                     this.getLogManager().logEvent(currentUser, request, QUESTION_ATTEMPT_RATE_LIMITED, response.getEntity());
                     String message = "You have made too many attempts at this question part. Please log in or try again later.";
                     return SegueErrorResponse.getRateThrottledResponse(message);
                 }
+                try {
+                    // And monitor on a blanket per IP Address basis for non-logged in users.
+                    // If we see serious misuse, this could be moved to *before* the attempt validation and checking,
+                    // to save server load. Since this occurs after the anon user notify event, that will catch most
+                    // misuse and this will catch misuse ignoring cookies or with repeated new anon accounts.
+                    misuseMonitor.notifyEvent(RequestIPExtractor.getClientIpAddr(request),
```
sacummins: My only concern is how this could interact with NAT based proxies in schools... If we get our threasholds right it shouldn't matter but we should just be careful.

src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
```diff
@@ -196,13 +199,26 @@ public Response answerQuestion(@Context final HttpServletRequest request,
...
                 try {
+                    // Monitor misuse on a per-question per-anonymous user basis:
                     misuseMonitor.notifyEvent(((AnonymousUserDTO) currentUser).getSessionId() + "|" + questionId,
                             AnonQuestionAttemptMisuseHandler.class.toString());
                 } catch (SegueResourceMisuseException e) {
                     this.getLogManager().logEvent(currentUser, request, QUESTION_ATTEMPT_RATE_LIMITED, response.getEntity());
                     String message = "You have made too many attempts at this question part. Please log in or try again later.";
                     return SegueErrorResponse.getRateThrottledResponse(message);
                 }
+                try {
```
sacummins: Perhaps in this case we should care more if the user is persistently getting the question wrong? This may protect us from blocking genuine users. i.e. reset the counter for the IP address if the question is correct? It does allow bad users to hide more easily.

jsharkey13:
I'm happy that, given the usage we see, even the lower soft limit will not affect many schools. It will not affect registered users either, and so I'm completely happy to throttle those not logged in.

PR #117, opened 2017-09-29
Enable all users to request a password reset

sacummins:
Note for after release: The reset token must be added to the email template to complete this change. We probably want to reword that whole email...

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/auth/SegueLocalAuthenticator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/auth/SegueLocalAuthenticator.java
@@ -201,9 +201,14 @@ public class SegueLocalAuthenticator implements IPasswordAuthenticator {
         Validate.notNull(userToAttachToken);
 
         LocalUserCredential luc = passwordDataManager.getLocalUserCredential(userToAttachToken.getId());
-        if (null == luc || null == luc.getPassword() || null == luc.getSecureSalt()) {
-            log.debug(String.format("No credentials available for this account id (%s)", luc.getUserId()));
-            throw new NoCredentialsAvailableException("This user does not have any local credentials setup.");
+        if (null == luc) {
+            // create a new luc as this user didn't have one before - they won't ever be able to login with this
+            luc = new LocalUserCredential(userToAttachToken.getId(),
+                    "LOCKED@" + new String(this.preferredAlgorithm.computeHash(UUID.randomUUID().toString(),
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/auth/SegueLocalAuthenticator.java
```diff
@@ -201,9 +201,14 @@ public String createPasswordResetTokenForUser(final RegisteredUser userToAttachT
...
         LocalUserCredential luc = passwordDataManager.getLocalUserCredential(userToAttachToken.getId());
-        if (null == luc || null == luc.getPassword() || null == luc.getSecureSalt()) {
-            log.debug(String.format("No credentials available for this account id (%s)", luc.getUserId()));
-            throw new NoCredentialsAvailableException("This user does not have any local credentials setup.");
+        if (null == luc) {
+            // create a new luc as this user didn't have one before - they won't ever be able to login with this
+            luc = new LocalUserCredential(userToAttachToken.getId(),
+                    "LOCKED@" + new String(this.preferredAlgorithm.computeHash(UUID.randomUUID().toString(),
+                            UUID.randomUUID().toString(), SHORT_KEY_LENGTH)),
+                    new String(this.preferredAlgorithm.computeHash(UUID.randomUUID().toString(),
```
jsharkey13: I'm not 100% sure I understand this salt bit; is it that it's safe to set it to some random junk now because the account can never be logged in to (the `@` in the password ensures that, but that it will be replaced by an actually secure salt when they actually choose a password?
If they do reset their password it does:  https://github.com/ucam-cl-dtg/isaac-api/blob/7b401003eda1b183b5361ab65cc5a48fc14f39c2/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java#L548-L570 which calls https://github.com/ucam-cl-dtg/isaac-api/blob/7b401003eda1b183b5361ab65cc5a48fc14f39c2/src/main/java/uk/ac/cam/cl/dtg/segue/auth/SegueLocalAuthenticator.java#L90-L100 which calls https://github.com/ucam-cl-dtg/isaac-api/blob/7b401003eda1b183b5361ab65cc5a48fc14f39c2/src/main/java/uk/ac/cam/cl/dtg/segue/auth/SegueLocalAuthenticator.java#L263-L271
where a new salt is generated? So I think it's fine, but wouldn't mind a second opinion!

sacummins:
Yep the salt is generated anytime a new password is set so it should be
fine.

--
Please excuse the brevity of this message it was sent on my mobile phone.

PR #119, opened 2017-10-04
Feature/augment related content with question completion info

mlt47:
Expands on https://github.com/ucam-cl-dtg/isaac-api/pull/111 so it is probably easier to merge that into master first.

These changes use the user's question attempt information to set whether related content questions have previously been answered.
Makes no extra queries to elasticsearch/postgres but does add one or two more traversals of the question page in Java.
I will hopefully check that it doesn't affect timings much tomorrow, but probably worth checking with me before merging.


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
```diff
@@ -242,6 +237,50 @@ public SeguePageDTO augmentQuestionObjectWithAttemptInformation(final SeguePageD
...
 
     /**
+     * A mathod which audments related questions with attempt information.
+     * i.e. sets whether the related content summary has been completed.
+     * @param content the content to be augmented.
+     * @param usersQuestionAttempts the user's question attempts.
+     */
+    private static void augmentRelatedQuestionsWithAttemptInformation(
+            final ContentDTO content,
+            final Map<String, Map<String, List<QuestionValidationResponse>>> usersQuestionAttempts) {
```
mlt47: I believe this is enough information to tell whether a question has been completed or not, but what should happen if a question part is changed or added? How do we handle this in other parts of the site?
If we do not have enough information here to see whether the question is completed, I might need to alter generateDerivedSummaryValues(...) to add a list of the current question IDs.

mlt47: I have since edited this method to do the calculation correctly.

PR #122, opened 2017-10-13
Check Referer and Origin headers for Cross-Site Request Forgery protection

jsharkey13:


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -319,6 +323,23 @@ public RegisteredUser getUserFromSession(final HttpServletRequest request) {
             return null;
         }
 
+        if (properties.getProperty(Constants.SEGUE_APP_ENVIRONMENT).equals(EnvironmentType.PROD.name())) {
```
jsharkey13: I wonder whether checking this once on Manager creation and storing the boolean as a class attribute would be better than checking every request?

sacummins: Possibly, but I think the properties all get preloaded into a map in memory anyway so it is probably not much difference.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -319,6 +323,23 @@ public RegisteredUser getUserFromSession(final HttpServletRequest request) {
...
+            // If in production, check if the request originated from Isaac:
+            String referrer = request.getHeader("Referer");  // Note HTTP Header misspelling!
+            if (null == referrer) {
+                log.warn("Authenticated request had no 'Referer' information set!");
```
sacummins: Given that you are logging these to try and help find out in what circumstances they happen, would it be useful to log the request url seeing as you have the request object in scope?

jsharkey13: I think this is a good idea!

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
jsharkey13: In which case I will leave it as-is!

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
sacummins: Did you want the attempted to access part on this line too or just the ones below?

jsharkey13: It never happend :P
The magic of rebasing! Good spot again though, thanks for checking carefully!

PR #125, opened 2017-10-19
Prevent server-specific log events being logged by client

jsharkey13:
As we move to using Kafka to process log events, ensure those log events which are expected to be produced only on the server are not able to be logged by the frontend or a malicious client.


src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
```diff
@@ -98,6 +101,11 @@
 
     public static final String SUBJECT_INTEREST = "SUBJECT_INTEREST";
 
+    public static final List<String> ISAAC_LOG_EVENT_TYPES = ImmutableList.of(ADD_BOARD_TO_PROFILE, CREATE_GAMEBOARD,
```
sacummins: Could this be a set rather than a list?

src/main/java/uk/ac/cam/cl/dtg/segue/api/LogEventFacade.java
```diff
@@ -98,6 +99,13 @@ public Response postLog(@Context final HttpServletRequest httpRequest, final Map
                     + TYPE_FIELDNAME + " property.").toResponse();
         }
 
+        String eventType = (String) eventJSON.get(TYPE_FIELDNAME);
+
+        if (ISAAC_LOG_EVENT_TYPES.contains(eventType) || SEGUE_LOG_EVENT_TYPES.contains(eventType)) {
```
sacummins: Perhaps to reduce the leakage of isaac stuff into segue could we combine this collection in the IsaacGuiceConfigurationModule / Segue equivalent and then dependency inject them in? This would allow us to more easily make this into a config parameter rather than hard code what we consider to be protected classes of log events

src/main/java/uk/ac/cam/cl/dtg/segue/comm/EmailManager.java
```diff
@@ -255,7 +256,7 @@ public void sendCustomEmail(final RegisteredUserDTO sendingUser, final String co
         ImmutableMap<String, Object> eventDetails = new ImmutableMap.Builder<String, Object>().put("userIds", ids)
```
sacummins: I know this is not effected in this commit but are we logging everything we need in this event? I.e. do we record the admin user who did it?

PR #127, opened 2017-10-27
Issue/isaac app 874/events email problems

sacummins:
This is a fairly urgent commit as for some reason http://tzurl.org/zoneinfo-outlook/Europe/ now no longer contains london and so the live site is likely to break when someone tries to book on an event.


src/main/java/uk/ac/cam/cl/dtg/util/Mailer.java
```diff
@@ -141,6 +141,9 @@ public void sendMultiPartMail(final String[] recipient, final String from, @Null
 
 		if (attachments != null) {
 			for (EmailAttachment attachment : attachments) {
+				if (null == attachment)
+					continue;
```
jsharkey13: 🙀

PR #131, opened 2017-11-09
Feature/fasttrack progress bar endpoint

mlt47:


src/main/java/uk/ac/cam/cl/dtg/isaac/dto/GameboardDTO.java
```diff
@@ -30,7 +30,7 @@
 public class GameboardDTO {
     private String id;
     private String title;
-    private List<GameboardItem> questions;
+    private List<? extends GameboardItem> questions;
```
jsharkey13: Is this syntax necessary? I though subclasses could be included in lists of the parent class?

jsharkey13: https://stackoverflow.com/a/3009779

PR #137, opened 2017-12-12
Restrict FastTrack progress queries to whitelist of IDs on the…

mlt47:
…back-end and document FastTrack classes


src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
```diff
@@ -45,6 +47,8 @@
      * Game specific variables.
      */
     public static final int GAME_BOARD_TARGET_SIZE = 10;
+    public static final Set<String> FASTTRACK_GAMEBOARD_WHITELIST
+            = ImmutableSet.of("ft_core_2017"); // Keep in sync with APP's fastTrackProgressEnabledBoards constant
```
sacummins: It would be nice to avoid hard coding gameboard ids in the constants file... If you think it is necessary, can we make it a config value so we don't have to rebuild the api if we want to add more fast track boards?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
```diff
@@ -280,7 +291,10 @@ public final Response getFastTrackGameboard(@Context final Request request,
 
         try {
             GameboardDTO gameboard;
-
+            if (!Constants.FASTTRACK_GAMEBOARD_WHITELIST.contains(gameboardId)) {
```
sacummins: What was the motivation for the white list check? The endpoint doesn't look that different to the normal gameboard one.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
mlt47: Sure, it would be nice to be able to update the list without needing a release. I will do that.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
mlt47: FastTrack progress is calculated using more question results than the standard gameboard.
I (ab)use tags to represent the collection and structure of these related questions so to get all the questions I'm interested in, I query for questions including the gameboard ID as a tag.
I white-list the gameboard ID to avoid querying elasticsearch with a string which hasn't been validated. It is not really a problem at the moment as it is only staff who have the ability to set gamebaord IDs.

PR #138, opened 2017-12-14
Issue/898/user streak with progress page

dunderwood88:
This is an updated version of the user streak backend, initially started as a branch from the previous one. The logic is different from the previous in that users must answers a certain number of question parts before increasing their streak for the day. It also requires a different front end implementation.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
@@ -22,6 +22,7 @@ import uk.ac.cam.cl.dtg.segue.dto.users.RegisteredUserDTO;
 import java.io.IOException;
 import java.net.HttpCookie;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 import static uk.ac.cam.cl.dtg.segue.api.Constants.*;
 
@@ -41,7 +42,7 @@ public class UserAlertsWebSocket implements IAlertListener {
     private Session session;
     private static ObjectMapper objectMapper = new ObjectMapper();
 
-    public static HashMap<Long, List<UserAlertsWebSocket>> connectedSockets = new HashMap<>();
+    public static ConcurrentHashMap<Long, List<UserAlertsWebSocket>> connectedSockets = new ConcurrentHashMap<>();
 
 
     /**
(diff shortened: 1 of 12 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
```diff
@@ -25,123 +25,188 @@
...
-    private ObjectMapper objectMapper;
     private UserAccountManager userManager;
     private RegisteredUserDTO connectedUser;
     private final IUserAlerts userAlerts;
     private final ILogManager logManager;
+    private final UserStatisticsStreamsApplication userStatisticsStreamsApplication;
     private Session session;
+    private static ObjectMapper objectMapper = new ObjectMapper();
 
     public static HashMap<Long, List<UserAlertsWebSocket>> connectedSockets = new HashMap<>();
```
mlt47: ConcurrentHashMap

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
```diff
@@ -25,123 +25,188 @@
...
+     * @param session
+     **          - contains information on the currently connected session
+     * @param status
+     *          - the status of the connection
+     * @param reason
+     *          - states the cause for closing the connection
+     */
+    @OnWebSocketClose
+    public void onClose(Session session, int status, String reason) {
+        connectedSockets.get(connectedUser.getId()).remove(this);
```
mlt47: Might want to consider removing the connected user from connectedSockets map altogether, but, in the worst case, will only get to the low tens of thousands of entries given our current release cycle length and number of unique users.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/kafkaStreams/UserStatisticsStreamsApplication.java
```diff
@@ -0,0 +1,535 @@
...
+        // timestamp of latest event
+        Long latestEventTimestamp = latestEvent.path("timestamp").asLong();
+
+        // set up calendar objects for date comparisons
+        Calendar latest = Calendar.getInstance();
+        latest.setTimeInMillis(latestEventTimestamp);
+        latest = roundDownToDay(latest);
+
+
+        // 1) If the current activity threshold for the day has been reached
```
mlt47: These comments break up the logic of the code well and so, in future, would be good candidates for where to break the method down and split into smaller subroutines/methods.

PR #141, opened 2017-12-22
Feature/popular gameboards endpoint fix

mlt47:
Allow null to be a value for ownerUserId, change logic to not check for user when ID is null and apply some style checker corrections.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
@@ -349,8 +349,8 @@ public class GameboardsFacade extends AbstractIsaacFacade {
                 if (e.getValue() > 1) {
                     GameboardDTO liteGameboard = this.gameManager.getLiteGameboard(e.getKey());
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
```diff
@@ -349,8 +350,8 @@ public Response getBoardPopularityList(@Context final HttpServletRequest request
                 if (e.getValue() > 1) {
                     GameboardDTO liteGameboard = this.gameManager.getLiteGameboard(e.getKey());
 
-                    RegisteredUserDTO ownerUser = userManager.getUserDTOById(liteGameboard.getOwnerUserId());
-                    if (ownerUser != null) {
+                    if (liteGameboard.getOwnerUserId() != null) {
+                        RegisteredUserDTO ownerUser = userManager.getUserDTOById(liteGameboard.getOwnerUserId());
                         liteGameboard.setOwnerUserInformation(associationManager.enforceAuthorisationPrivacy(
                                 currentUser, userManager.convertToUserSummaryObject(ownerUser)));
                     }
```
mlt47: Check if the ID is not null before searching for the user

PR #144, opened 2018-01-05
ucam-cl-dtg/isaac-app#911 - Automatically add a student to a g…

sacummins:
…roup and grant access if they are promoted from the waiting list. Also allow event managers to add previously unbooked users to the waiting list when an event is already full.


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -148,21 +149,72 @@ public Long countNumberOfBookingsWithStatus(final String eventId, final BookingS
...
+        // check if already booked
+        if (this.isUserBooked(event.getId(), user.getId())) {
+            throw new DuplicateBookingException(String.format("Unable to book onto event (%s) as user (%s) is already"
+                    + " booked on to it.", event.getId(), user.getEmail()));
+        }
+
+        EventBookingDTO booking;
+        try {
+            // Obtain an exclusive database lock to lock the booking
+            this.bookingPersistenceManager.acquireDistributedLock(event.getId());
```
mlt47: It looks like the code tries to grab the lock for this event ID both here and then, after acquiring it, in the createBooking(...) method which should result in deadlock.

sacummins: Agreed! I think it should be safe in this method to not acquire a lock as if another process fills up the event before this method calls createBooking then it will add them to the waiting list which is fine and acquires a lock separately. I will just remove the acquireLock in this method.

PR #145, opened 2018-01-12
Set Kafka logged events topic name as a global variable

dunderwood88:
Removes the need to change the string across the API whenever the logged events topic name changes (even though this should happen rarely).

---

**Pull Request Check List**
- [ n/a] Added enough Logging to monitor expected behaviour change
- [n/a] Security - Access Control - Check authorisation on every new endpoint
- [n/a] Security - New dependency - configured sensibly not relying on defaults
- [n/a] Security - New dependency - Searched for any know vulnerabilities
- [n/a] Security - New dependency - Signed up team to mailing list
- [n/a] Security - New dependency - Added to dependency list
- [n/a] DB schema changes - postgres-rutherford-create-script updated
- [n/a] DB schema changes - upgrade script created matching create script
- [n/a] Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)
- [ ] Peer-Reviewed

Signed Off By: @

### DIFF
--- a/config-templates/linux-local-dev-segue-config.properties
+++ b/config-templates/linux-local-dev-segue-config.properties
@@ -45,6 +45,7 @@ POSTGRES_DB_PASSWORD=rutherf0rd
 KAFKA_HOSTNAME=localhost
 KAFKA_PORT=9092
 KAFKA_REPLICATION_FACTOR=1
+KAFKA_TOPIC_LOGGED_EVENTS=topic_logged_events_v1
 
 # Kafka Streams Applications
 KAFKA_STREAMS_STATE_DIR=/tmp/kafka-streams-state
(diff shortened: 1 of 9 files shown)


jsharkey13:
We discussed this, and fear that having it in the properties file could be tricky to keep the many versions of that file in sync. I had worried about multiple APIs, but Dan correctly points out that changing the topic version would require downtime to avoid loss of log events in Kafka.

So I think the conclusion was that just having the name in the `segue/api/Constants.java` file directly was the simplest way to avoid these issues, and we can cope with dev and staging using a new topic if we have to change the live one for whatever reason.

jsharkey13:
Superseded by https://github.com/ucam-cl-dtg/isaac-api/pull/150

PR #146, opened 2018-01-12
Adjust event bookings table to be able to record attendance

mlt47:
---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/resources/db_scripts/postgres-rutherford-create-script.sql
+++ b/src/main/resources/db_scripts/postgres-rutherford-create-script.sql
@@ -176,7 +176,8 @@ CREATE TABLE event_bookings (
     user_id integer NOT NULL,
     status text DEFAULT 'CONFIRMED'::text NOT NULL,
     updated timestamp without time zone,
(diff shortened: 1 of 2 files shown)


src/main/resources/db_scripts/postgres-rutherford-create-script.sql
```diff
@@ -176,7 +176,8 @@ CREATE TABLE event_bookings (
     user_id integer NOT NULL,
     status text DEFAULT 'CONFIRMED'::text NOT NULL,
     updated timestamp without time zone,
-    additional_booking_information jsonb
+    additional_booking_information jsonb,
+    attended boolean
```
sacummins: Do you think we need this or shall we just have another status enum entry that is attended instead of CONFIRMED?

mlt47: Ah yes, you're right. I had thought of doing that the other week when I first thought about adding this functionality, just didn't remember when it came to adding it yesterday.

PR #151, opened 2018-02-16
Feature/record event attendance

mlt47:
**Pull Request Check List**
- Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -876,16 +941,26 @@ public final Response getEventOverviews(@Context final HttpServletRequest reques
...
 
         fieldsToMatch.put(TYPE_FIELDNAME, Arrays.asList(EVENT_TYPE));
 
         Map<String, AbstractFilterInstruction> filterInstructions = null;
-        if (null == showActiveOnly || showActiveOnly) {
+        if (showActiveOnly != null && showActiveOnly && showInactiveOnly != null && showInactiveOnly) {
+            return new SegueErrorResponse(Status.BAD_REQUEST,
+                    "You cannot request both show active and inactive only.").toResponse();
+        }
+        if (showActiveOnly != null && showActiveOnly) {
```
jsharkey13: This used to be an OR, meaning that if nothing was provided there was a default case. Does it work if neither `showActiveOnly` or `showInactiveOnly` are provided?

jsharkey13:
Was it this code we decided would affect whether a user was counted as booked for the [`isUserBooked(...)`](https://github.com/ucam-cl-dtg/isaac-api/blob/24e6e4c8652c7a68cfe4f4d8ca3c8eb45a91b39f/src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java#L188-L195) function? I think it will, and I think we'll need to check if a booking is `CONFIRMED`, `ATTENDED` or `ABSENT` and still count that as booked.

PR #154, opened 2018-02-21
Handling thrown exceptions in kafka streams app

dunderwood88:
This is an uber quickfix which prevents an exception in websockets form crashing the streamsapp.

The streams app crashing in itself is an issue that we need to address so we might not end up using this...

---

**Pull Request Check List**
- [ ] Peer-Reviewed


dunderwood88:
Decided to change this from a "quick hack" to a "actually sensible fix". Instead of handling exceptions which we know about outside of the streams apps, it is probably more appropriate to ensure that the streams apps themselves correctly deal with uncaught exceptions which may occur in future without our prior knowledge.

PR #155, opened 2018-03-01
Ability for admin to set a streak value for a given user

dunderwood88:
This already exists the teacher badge branch, but have separated it out into its own one here to included in a separate release.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
```diff
@@ -1641,4 +1641,32 @@ public Response getDiagnostics(@Context final Request request, @Context final Ht
         }
     }
 
+
+    @GET
```
jsharkey13: Should ideally be a `POST`, with `streakValue` being the body.

PR #160, opened 2018-03-19
Catch SegueDatabaseException in QuestionFacade to avoid false logging of question attempts

dunderwood88:
---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
@@ -240,6 +240,10 @@ public class QuestionFacade extends AbstractSegueFacade {
             SegueErrorResponse error = new SegueErrorResponse(Status.BAD_REQUEST, "Bad request - " + e.getMessage(), e);
             log.error(error.getErrorMessage(), e);
             return error.toResponse();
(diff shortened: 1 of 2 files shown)


jsharkey13:
This looks sensible, but I have two comments:

1) We probably shouldn't send anything about the error back to the user; a `500 Internal Server Error` will display the big "Error on the Isaac server" page, and make it obvious something has gone wrong. The message ought to be more like "Error checking answer to question. Please try again later!".

2) The line [`AbstractSegueUserDTO currentUser = this.userManager.getCurrentUser(request);`](https://github.com/ucam-cl-dtg/isaac-api/blob/e1cdde66c57fbd9f9608ef4f0c69c8a52c079f70/src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java#L140) in the facade actually hides many database errors away from the rest of the method. If the database isn't there or seriously broken, this method returns an anonymous user instead of the correct registered user. There are server log messages, but otherwise it is 'silent' and swallows the exceptions. This means that the attempt is not stored to the database later on, only in memory, and so the user is unlikely to see the error page anyway if something goes badly wrong . . .

jsharkey13:
That addresses (1), but (2) is more serious and in no way limited to this part of the code.

PR #169, opened 2018-05-11
Feature/user badge management

dunderwood88:
Initial structure for generic user badge management. Currently includes solutions for a few specific teacher badges.
This branch contains functionality for:

- Communication with database storage to extract and update badge states;
- Inner-API management for Badge object manipulation;
- Interface definitions for specific badge unlocking policies;
- Facade-level triggering of badge updates and exposure of badge status;

---

**Pull Request Check List**
- [ ] Peer-Reviewed


dunderwood88:
I am a little uncertain about various aspects of the Guice injection here so it would be good to get a review of this.
Additionally we have explicitly included transactions in the badge management architecture, but the current implementation does not make use of passing around a single DB connection amongst layers. I'm wondering how to achieve this.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserBadgeManager.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Gets an up-to-date badge by either retrieving from the database or initialising first-time on the fly
+     *
+     * @param conn database connection
+     * @param user owner of badge record
+     * @param badgeName enum of badge to be updated
+     * @return user badge object
+     * @throws SegueDatabaseException
+     */
+    public UserBadge getOrCreateBadge(Connection conn, RegisteredUserDTO user, Badge badgeName)
```
sacummins: Could we avoid the manager level having to know about sql classes e.g. Connection? I would normally expect that to be hidden away in the DAO layer. It is not urgent but maybe something to refactor later?

sacummins:
> Additionally we have explicitly included transactions in the badge management architecture, but the current implementation does not make use of passing around a single DB connection amongst layers. I'm wondering how to achieve this. >

Transactions are fine, however the architecture expects that this feature is handled at the lower levels (mainly because not all databases provide transactions in the same way and we wanted to be able to change dbms's easily - ahem Mongo...) 

If you want to do transactions you could handle them in the PersistenceManager layer? Happy to chat about this separately.

dunderwood88:
Latest commit generalises database transactions for this particular area of the API. If the underlying database changes then a transactional template can be optionally used depending on whether we need it or not.

Unfortunately our model requires some form of transaction baked into at least one level higher than the persistence level. This is because of how we have content data (elasticsearch) separate from user activity data (postgres) which are both needed to compute badges. This requires a "get from db -> modify -> update db" approach, and of course we need this to be atomic.
Hence in this incarnation I have the transactions defined in one layer above the persistence layer.

The edge case which we are particularly worried about with the badges is when a user performs an action which triggers a badge update and the badge does not yet exist, which sets of a chain of events which computes the badge state from the beginning of time. An example is the teacher group badge, which updates every time a teacher makes a new group (the process also describes any badge):

1. Create Group -> Insert group to Group table;
**Start Transaction**
2. Get group badge from db;
3. a) Badge state is null -> requires computing (go to 4a);
b) Badge state is not null -> append group ID to badge state (go to 4b);
4. a) Get all user's groups from group table to compute badge;
b) Simply append group ID to badge state list if not already present;
5. Update badge state in db.
**End transaction**

The worry is that the user may trigger stage 1 while stage 3a/4a is already occurring from a previous trigger. The state of the badge on the second trigger must reflect the state after the first trigger and hence the whole "get -> modify -> update" chain must exist in a single transaction and block the second update attempt.
In the case of 3b/4b we have to check in the group ID is already present in the badge state list because it might have been included in a situation where 3a/4a blocked a second trigger. An alternative to this is to start the transaction at point 1, but this would require: 
- moving the transactions to higher layer of the API;
- enforcing "repeatable read" isolation level on the transaction to prevent additional inserts to the group table, which would require secondary triggers to retry if the were ever blocked.

It's possible that we would prefer this option if we believe checking for the presence of a group ID every time we try and add it to the state is less desirable.

Please let me know if any of this makes no sense.

dunderwood88:
Front end pull request:
https://github.com/ucam-cl-dtg/isaac-app/pull/969

Morpheu5:
- [x] Added database migration to create the `user_badges` table.
- [x] Manual testing.

PR #172, opened 2018-06-08
Move log event types to Enums for consistency and neatness

jsharkey13:
We've had issues with log constants having incorrect text values, and had to maintain a set of log names to prevent front-end clients logging with protected values. This fixes both of these things, using a LogType interface to maintain as much of the Isaac/Segue divide as possible.
The LogManagerEventPublisher will still pass on String log names, although currently I do not think there are any configured listeners.

---

**Pull Request Check List**


jsharkey13:
One question; the `EnumUtils.isValidEnum(...)` uses exception handling as it's control flow. As we expect these not to be valid enums almost all the time, this may be a poor way of doing things?

src/main/java/uk/ac/cam/cl/dtg/segue/dao/ILogManager.java
```diff
@@ -65,7 +82,7 @@
      *            object
      * @throws SegueDatabaseException 
      */
-    void logInternalEvent(AbstractSegueUserDTO user, String eventType, Object eventDetails);
+    void logInternalEvent(AbstractSegueUserDTO user, Constants.LogType eventType, Object eventDetails);
```
sacummins: Can this just be LogType?

PR #175, opened 2018-06-12
Feature/prometheus exporters

mlt47:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/isaac/configuration/IsaacApplicationRegister.java
```diff
@@ -64,8 +64,7 @@ public IsaacApplicationRegister() {
         SegueGuiceConfigurationModule segueGuiceConfigurationModule = new SegueGuiceConfigurationModule();
         IsaacGuiceConfigurationModule isaacGuiceConfigurationModule = new IsaacGuiceConfigurationModule();
         
-        injector = Guice.createInjector(new SchoolLookupConfigurationModule(),
-                isaacGuiceConfigurationModule, segueGuiceConfigurationModule);
+        injector = Guice.createInjector(isaacGuiceConfigurationModule, segueGuiceConfigurationModule);
```
jsharkey13: Why does this remove the school one?

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
```diff
@@ -43,12 +55,34 @@
...
 
-    public static ConcurrentHashMap<Long, ConcurrentLinkedQueue<UserAlertsWebSocket>> connectedSockets = new ConcurrentHashMap<>();
-    private static Long websocketsOpened = 0L;
-    private static Long websocketsClosed = 0L;
+    private static final int MAX_CONCURRENT_WEB_SOCKETS_PER_USER = 10;
+
+    // Named unsafeConnectedSockets because, although non-aggregate operations on the concurrent hash map are fine,
+    // operations on the user sets of websockets are unsafe unless used with the matching user lock.
+    private static Map<Long, Set<UserAlertsWebSocket>> unsafeConnectedSockets = Maps.newConcurrentMap();
+    private static final int MAX_NUMBER_OF_CONCURRENT_USER_TAB_OPERATIONS = 200; // ~ max number of concurrent users
```
jsharkey13: Huh? If this comment is accurate, this seems a strange place to put a limit on the number of concurrent users!
Is this "concurrently opening and closing WebSockets at any particular instant", or "concurrently have WebSockets open"?

mlt47: It is the "concurrently opening, closing and sending user messages down a WebSocket at any particular instant".
If we can find a more explanatory name, we should definitely use it.

src/main/java/uk/ac/cam/cl/dtg/isaac/configuration/IsaacApplicationRegister.java
mlt47: I moved it into the isaac guice configuration module to tidy things up. This comment did remind me to delete the SchoolLookupConfigurationModule though, thanks!

pom.xml
```diff
@@ -236,6 +236,37 @@
 			<version>4.0</version>
 		</dependency>
 
+		<dependency>
+			<groupId>io.prometheus</groupId>
+			<artifactId>simpleclient</artifactId>
+			<version>0.3.0</version>
```
sacummins: Is there a reason we are using v0.3.0 instead of v0.4.0?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/monitors/IsaacMetrics.java
```diff
@@ -0,0 +1,12 @@
+package uk.ac.cam.cl.dtg.isaac.api.monitors;
```
sacummins: Don't forget the apache license at the top of new files.

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
sacummins: This probably should be a config value in the settings file?

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
```diff
@@ -43,12 +55,34 @@
...
     private static final Logger log = LoggerFactory.getLogger(UserAlertsWebSocket.class);
-    private static final int MaxConcurrentWebSocketsPerUser = 10;
+
+    /**
+     * This static method obtains a user lock and sends an alert to each of that user's websockets.
+     * @param userId ID of the user to send the messages, we do not check its validity here.
+     * @param alert the alert to send to the user.
+     */
+    public static void notifyUserOfAlert(final long userId, final IUserAlert alert) {
+        Lock userLock = userLocks.get(userId);
```
sacummins: Are there any reasons by we would want this to be a distributed lock across multiple instances of the API? Just thinking we did the distributed locking stuff for the events manager in postgres.

src/main/java/uk/ac/cam/cl/dtg/segue/dos/PgUserStreakManager.java
```diff
@@ -89,19 +89,15 @@ public int getLongestStreak(final RegisteredUserDTO user) {
...
-            }
+        long userId = user.getId();
+        try {
+            IUserAlert alert = new PgUserAlert(null, userId,
+                    objectMapper.writeValueAsString(ImmutableMap.of("streakRecord", this.getCurrentStreakRecord(user))),
+                    "progress", new Timestamp(System.currentTimeMillis()), null, null, null);
+
+            UserAlertsWebSocket.notifyUserOfAlert(userId, alert);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
sacummins: Can we do something better than just printing the stack trace while you are fixing things? :)

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
mlt47: Yes, my first thought on this was the event management postgres lock.

The locks are used when interacting with the map of websocket connections.
As the map of connections is stored in the Java process's memory (i.e. not shared between multiple instances of the API) there is currently no benefit to having the lock distributed.

The striped locks are a little more lightweight than having to get postgres involved.

To get the websocket messaging working with multiple APIs we would need the map of websockets to be shared between the various APIs using redis/memcached or we might be able to load balance connections in a sticky way so that every one of a user's websocket connections are being handled by the same API.

Whichever way we decide to do it, it would need some rethinking and rejigging, and I'm pretty sure we'd reconsider the locking when adapting the map to work with multiple APIs. I will add a comment about using postgres for the distributed lock if we get to having multiple APIs processing websockets correctly at the same time.

src/main/java/uk/ac/cam/cl/dtg/segue/dos/PgUserStreakManager.java
mlt47: Will do.

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
mlt47: Moved MAX_CONCURRENT_WEB_SOCKETS_PER_USER  to be a configuration setting. The other, MAX_NUMBER_OF_CONCURRENT_USER_TAB_OPERATIONS , is used to decide the value of a static property of the class so does not play quite as nicely with our guice injected property reading processes.

The way around it would be to have a singleton instance of a UserAlertsManager class which holds the static stuff and instances of UserAlerts.

.gitignore
```diff
@@ -1,7 +1,10 @@
...
 .settings
 .classpath
-/target
+target
 .idea/
 *.iml
 .vagrant/
+.env
+grafana
+prometheus/data
```
jsharkey13: Should we ever end up with files in the `grafana` or `prometheus/data` directories? Surely the output is configured to be somewhere more reasonable?

jsharkey13:
I'm also worried about the `.env` file idea; there are secrets in that file that are mounted read-only into the API alone. With the environment variable file, every Docker container created from the API directory (the equality and chemistry checkers, and all the graphana ones too I guess?) will have access to these secrets that they do not need access to.

.gitignore
mlt47: This is only for local development, on the live site the docker compose file is generated (https://github.com/ucam-cl-dtg/isaac-monitor) and stores the data in var/lib/graphana|prometheus

mlt47:
I believe that the environment variables using .env are only visible while creating the docker containers and so the containers will only see what is passed in through in the docker compose file. I'll double check, though.

PR #176, opened 2018-06-29
Allow additional group managers to see all assignments in group CSV file

jsharkey13:
The method previously only listed assignments the current users had set to the group, not all the assignments the group has.
This is still restricted to the whitelisted group managers.

---

**Pull Request Check List**


jsharkey13:
Compare to permissions here: https://github.com/ucam-cl-dtg/isaac-api/blob/0ec46dd5ec60cc211b399f5d5d0c9a52d4369463/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java#L299-L305

PR #185, opened 2018-10-03
Expose a new endpoint which allows group managers to initiate password…

mlt47:
… reset requests for group members

---

**Pull Request Check List**
- Security - Injection - everything run by an interpreter (SQL, OS...) is either validated or escaped
- Security - Data Exposure - PII is not stored or sent unencrypted
- Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)
- [x] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -310,6 +310,73 @@ public Response getUserPreferences(@Context final HttpServletRequest httpServlet
...
+
+            UserSummaryDTO userOfInterestSummaryObject = userManager.convertToUserSummaryObject(userOfInterest);
+
+            // decide if the user is allowed to view this data.
+            if (!currentUser.getId().equals(userIdOfInterest)
+                    && !userAssociationManager.hasPermission(currentUser, userOfInterestSummaryObject)) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+
+            misuseMonitor.notifyEvent(currentUser.getEmail(), PasswordResetRequestMisuseHandler.class.toString());
```
jsharkey13: Is it sensible to make this request count towards the teacher's own password reset limit?

mlt47: Good point, it might not be a problem in practice but I think this is an improvement.
```suggestion
            misuseMonitor.notifyEvent(currentUser.getEmail() + "_group_member_reset", PasswordResetRequestMisuseHandler.class.toString());
```

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -310,6 +310,73 @@ public Response getUserPreferences(@Context final HttpServletRequest httpServlet
...
+            RegisteredUserDTO currentUser = userManager.getCurrentRegisteredUser(httpServletRequest);
+
+            RegisteredUserDTO userOfInterest = userManager.getUserDTOById(userIdOfInterest);
+            if (userOfInterest == null) {
+                throw new NoUserException("No user found with this ID.");
+            }
+
+            UserSummaryDTO userOfInterestSummaryObject = userManager.convertToUserSummaryObject(userOfInterest);
+
+            // decide if the user is allowed to view this data.
```
sacummins: Can we use this endpoint for admins too or do we already have one?

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -310,6 +310,73 @@ public Response getUserPreferences(@Context final HttpServletRequest httpServlet
...
+     * having to know the group members account email.
+     *
+     * @param request - request information used for caching
+     * @param httpServletRequest - the request, to work ou the current user
+     * @param userIdOfInterest - userId of interest - usually a the teacher's student
+     * @return a successful response regardless of whether the email exists or an error code if there is a technical
+     *         fault
+     */
+    @POST
+    @Path("users/{user_id}/resetpassword")
```
sacummins: Can this endpoint replace the usual reset password endpoint? I suspect the only difference is the log event and who is triggering the change. It might reduce the amount of duplicated code?

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -310,6 +310,73 @@ public Response getUserPreferences(@Context final HttpServletRequest httpServlet
...
+                    && !userAssociationManager.hasPermission(currentUser, userOfInterestSummaryObject)) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+
+            misuseMonitor.notifyEvent(currentUser.getEmail() + "_group_member_reset", PasswordResetRequestMisuseHandler.class.toString());
+            userManager.resetPasswordRequest(userOfInterest);
+
+            this.getLogManager()
+                    .logEvent(currentUser, httpServletRequest, SegueLogType.PASSWORD_RESET_REQUEST_RECEIVED,
+                            ImmutableMap.of(
```
sacummins: If we are using these logs for security audit too, should we be recording the email address of the requester or is the userId sufficient?

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
mlt47: Possibly, but the reset password functionality is also available for people who are not signed in, and so would not have a user ID.

sacummins: Ah good point! Happy that it differs on that basis.

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
mlt47: This endpoint will work for admins because `userAssociationManager.hasPermission(...)` returns `True` for admins. 

However, there is no front-end way for admins to use it yet, partially because it is less useful for admins to create the token. The token is only valid for 24 hours and if the user doesn't read their emails until after the 24 hours it would likely cause more confusion than it's worth so it is usually easier to just ask them to request the reset email themselves when they are ready.

mlt47: This endpoint will work for admins because `userAssociationManager.hasPermission(...)` returns `True` for admins. 

However, there is no front-end way for admins to use it yet, partially because it is less useful for admins to create the token. The token is only valid for 24 hours and if the user doesn't read their emails until after the 24 hours it would likely cause more confusion than it's worth so it is usually easier to just ask them to request the reset email themselves when they are ready.

PR #188, opened 2018-10-12
Stop pulling down the json blob which has question attempt data in wh…

sacummins:
…en we aren't using it. This seems to save a significant amount of time when loading the assignment progress page for a particular assignment.

---

**Pull Request Check List**
- [x] Peer-Reviewed


sacummins:
The thing I don't like about it was hijacking the same dto and not fully populating it. I only could bring myself to commit it because I thought it would go in quicker. If we have time I would create a superclass with fewer fields and use that in stead.

Morpheu5:
Probably unrelated, and I suspect the answer is going to be MongoDB, but why do we not use some kind of ORM or Active Record thing? They typically provide a query builder that creates a single query as much as possible. As long as we commit to rDB, libraries normally provide drivers for many different databases.

sacummins:
So we used to try a few different ORM libraries but they built stupid (and slow) queries (when we used them on Mongo - these probably should feature on the scale of horrible if they aren't already on there). 

I think we used MongoJack before...

Anyway when we stripped mongo out and then put postgres in we never put an equivalent ORM back. Happy for us to evaluate some and put them in if we think it will make life better.

Morpheu5:
I'm partial to Active Record because I know and love it from RoR. In RoR, you have the option of including or excluding fields when pulling data from the DB. It's easy to do with a dynamic language, I don't know how it is going to work in Java. For now, I'll stick to trying to bring down the execution time of the most horrible queries we have at the moment. Happy to look into this afterwards.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
```diff
@@ -374,40 +375,6 @@ public void recordQuestionAttempt(final AbstractSegueUserDTO user,
...
-            }
-            
-            return mapToReturn;
-        } else {
-            AnonymousUserDTO anonymousUser = (AnonymousUserDTO) user;
-            // since no user is logged in assume that we want to use any anonymous attempts
-            return this.questionAttemptPersistenceManager.getAnonymousQuestionAttempts(anonymousUser.getSessionId());
-        }
-    }
-
```
mlt47: Was not being used anywhere.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -1095,8 +1122,8 @@ private GameboardItem augmentGameItemWithAttemptInformation(final GameboardItem
         gameItem.setQuestionPartStates(questionPartStates);
         Integer questionPartsTotal = questionPartsCorrect + questionPartsIncorrect + questionPartsNotAttempted;
         gameItem.setQuestionPartsTotal(questionPartsTotal);
-        Float percentCorrect = 100f * new Float(questionPartsCorrect) / questionPartsTotal;
-        Float percentIncorrect = 100f * new Float(questionPartsIncorrect) / questionPartsTotal;
+        Float percentCorrect = 100f * questionPartsCorrect / questionPartsTotal;
+        Float percentIncorrect = 100f * questionPartsIncorrect / questionPartsTotal;
```
jsharkey13: Does this definitely avoid the integer division issues?

mlt47: It does, although I agree, I was sceptical too. 
> *15.17. Multiplicative Operators*
> The operators *, /, and % are called the multiplicative operators.
> The multiplicative operators have the same precedence and are syntactically left-associative (they group left-to-right).
> ...
> If the promoted type is float or double, then floating-point arithmetic is performed.
> https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.17

It was IntelliJ that suggested the change... happy to rearrange it and add a cast so that it is more explicit if you like though:

`float percentCorrect = ((float)questionPartsCorrect / questionPartsTotal) * 100`

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -582,13 +677,13 @@ public Response getGroupAssignmentsProgressDownloadCSV(@Context final HttpServle
             ArrayList<String> headerRow = Lists.newArrayList();
             Collections.addAll(headerRow, "Last Name,First Name,% Correct Overall".split(","));
             List<String> gameboardTitles = Lists.newArrayList();
-            for (AssignmentDTO assignment : assignments) {
```
jsharkey13: This change proved to be problematic!

PR #189, opened 2018-10-12
Feature/cookie invalidation

jsharkey13:
Add a per-user session token that has to match as part of the authentication process, allowing us to invalidate a single users cookies, without having to log everyone out.
Cookies now store the expiry time, not the date signed; this will allow variable length sessions in future.
This also refactors the way WebSockets authenticate, to try and move the authorisation code into the right place. Except that the classes are all different and so they cannot easily share a method. The duplicated methods are really not nice . . .

---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -345,28 +330,66 @@ public RegisteredUser getUserFromSession(final HttpServletRequest request) {
...
+            log.error("Internal Database error. Failed to resolve current user.", e);
+            return null;
+        } catch (NumberFormatException e) {
+            log.info("Invalid user id detected in session. " + currentSessionInformation.get(SESSION_USER_ID));
+            return null;            
+        }
+    }
+
+    /**
+     * @see #getUserFromSession(HttpServletRequest) - the two types of "request" have identical methods but are not
```
sacummins: Not sure I understand the need to duplicate methods in this way? Could it be refactored?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -460,6 +461,38 @@ private static EmailManager getMessageCommunicationQueue(final IUserDataManager
...
+     *            - list of known providers.
+     * @param emailQueue
+     *            - so that we can send e-mails.
+     * @param mapperFacade
+     *            - for DO and DTO mapping.
+     * @return Content version controller with associated dependencies.
+     */
+    @Inject
+    @Provides
+    @Singleton
```
sacummins: I am not sure what about this change means this now has to be a singleton?

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -772,14 +795,22 @@ private void createSession(final HttpServletRequest request, final HttpServletRe
...
         String userId = user.getId().toString();
+        String userSessionToken = user.getSessionToken().toString();
         String hmacKey = properties.getProperty(HMAC_SALT);
 
         try {
-            String currentDate = sessionDateFormat.format(new Date());
-            String sessionHMAC = this.calculateSessionHMAC(hmacKey, userId, currentDate);
+            Calendar calendar = Calendar.getInstance();
+            calendar.add(Calendar.SECOND, sessionExpiryTimeInSeconds);
+            String sessionExpiryDate = sessionDateFormat.format(calendar.getTime());
```
sacummins: A much better idea!

sacummins:
Did I miss the endpoint / mechanism for changing the session token?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
jsharkey13: It'll be created once per WebSocket request otherwise, ~50k times a week, since it's injected into each WebSocket object and one is made per request. I think? @sacummins

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
jsharkey13: I'd love for it to be, but they're different classes with different class structures and return types for "getCookies", so I can't find a way to make them use the same method.

jsharkey13:
There is as yet no way of changing the session token in-platform, figured that could come later!

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
Morpheu5: I'm glad to see that single inheritance is causing more problems than it is solving 👍

PR #193, opened 2018-10-26
Feature/667 group membership fixes

sacummins:
This PR converts groups and group membership delete functionality into a tombstoned version for relational consistency.

It also provides endpoints for users to manage their membership from active, to inactive.

There is also a change to enable any additional manager to delete assignments set to a group they manage.

This PR should be merged along with: https://github.com/ucam-cl-dtg/isaac-app/pull/978


src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
```diff
@@ -131,6 +131,90 @@ public Response getGroupsForCurrentUser(@Context final HttpServletRequest reques
...
+     * @return List of groups for the current user.
+     */
+    @GET
+    @Path("/membership")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getGroupMembership(@Context final HttpServletRequest request,
+                                            @Context final Request cacheRequest, @QueryParam("archived_groups_only") final boolean archivedGroupsOnly) {
+        try {
+            RegisteredUserDTO user = userManager.getCurrentRegisteredUser(request);
+            List<UserGroupDTO> groups = groupManager.getGroupMembershipList(user);
```
Morpheu5: There is some serious potential for repeated identical queries hidden in here. Not that we can do anything about this in this PR, but we may want to start becoming more aware of these issues.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/GroupManager.java
```diff
@@ -252,11 +286,34 @@ public void addUserToGroup(final UserGroupDTO group, final RegisteredUserDTO use
...
+     *            - the new membership status
+     * @throws SegueDatabaseException
+     *             - If an error occurred while interacting with the database.
+     */
+    public void setMembershipStatus(final UserGroupDTO group, final RegisteredUserDTO user, GroupMembershipStatus newStatus)
+            throws SegueDatabaseException {
+        Validate.notNull(group);
+        Validate.notNull(user);
+        // we don't want people to delete user membership via this route as observers are not notified.
+        Validate.isTrue(!GroupMembershipStatus.DELETED.equals(newStatus), "Deletion of a group membership should not use this route.");
```
mlt47: One small suggestion: instead of Validate.isTrue(...) doing a run-time check, we could have two Enums, one inheriting the other, with the subclass also including delete. Then we could use types to enforce this at compile-time. Probably not worth the work now but an idea for the future.

jsharkey13:
Ah, that's a good point. I may have been overzealous. The other two migrations caused NPEs, but perhaps last_updated ought to be NULL if unknown. Will have to check what that does!

PR #194, opened 2018-11-07
Update and fix the API Swagger configuration

jsharkey13:
Updating Swagger required updates to several other dependencies (it was incompatible with our outdated other dependencies). I have updated them too, and all seems ok.

I have applied IntelliJ suggested refactoring to the Facade classes I touched, since I was carefully reviewing them all anyway. I have also updated JavaDoc and comments to be more accurate.
This review may be easier commit by commit, since they are all nicely standalone but may look confusing together (especially optimising the Java imports! It's a separate commit before the rest for a reason).

There are one or two permissions adjustments where some methods in a Facade allowed all staff and others only admins/event managers etc.

I also added some deprecation warnings to endpoints that did not seem to work, or were unused and/or unmaintained. We should review these after Swagger is deployed and remove them if possible.


jsharkey13:
You should be able to naviagte to http://localhost:8080/isaac-api/api-docs/#/ to test this!

src/main/java/uk/ac/cam/cl/dtg/segue/api/SegueDefaultFacade.java
```diff
@@ -89,6 +89,7 @@ public SegueDefaultFacade(final PropertiesLoader properties, final ContentMapper
     @Path("/")
     @Produces(MediaType.TEXT_HTML)
     @Cache
+    @Deprecated
```
sacummins: I think originally this was here so that when someone went to /api they gout the swagger page which out having to type anything else. Does this now work as part of the swagger upgrade?

PR #195, opened 2018-11-22
Directly use Postgres Docker image rather than a pre-built Isaac one

jsharkey13:
This avoids the need for building a pretty pointless image, and adds a volume so that local databases are persisted. 

*WARNING*
This updates the local postgres image and container definition. If you run `docker-compose -f compose-api-deps.yml up -d` _you will lose your local database_!

I strongly recommend doing the following:
1) Dump your local database: `docker exec -i postgres pg_dump --username=rutherford --clean > postgres.sql`
2) Pull these changes.
3) Create a new Docker volume for the data: `docker volume create --name=pg-local`
4) Update your local postgres: `docker-compose -f compose-local-deps.yml up -d postgres`
5) Restore your local database: `docker exec -i postgres psql -U rutherford -f - --quiet < postgres.sql`
6) Delete the temporary file: `rm postgres.sql`

### DIFF
--- a/compose-local-deps.yml
+++ b/compose-local-deps.yml
@@ -29,7 +29,14 @@ services:
   postgres:
     network_mode: bridge
     container_name: postgres
-    image: isaac-pg
+    image: postgres:9.6
+    volumes:
+      - pg-local:/var/lib/postgresql/data
+      - ./src/main/resources/db_scripts/postgres-rutherford-create-script.sql:/docker-entrypoint-initdb.d/00-isaac-create.sql:ro
+      - ./src/main/resources/db_scripts/postgres-rutherford-functions.sql:/docker-entrypoint-initdb.d/01-isaac-functions.sql:ro
+    environment:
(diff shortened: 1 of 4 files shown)


compose-local-deps.yml
```diff
@@ -29,7 +29,14 @@ services:
   postgres:
     network_mode: bridge
     container_name: postgres
-    image: isaac-pg
+    image: postgres:9.6
+    volumes:
+      - pg-local:/var/lib/postgresql/data
+      - ./src/main/resources/db_scripts/postgres-rutherford-create-script.sql:/docker-entrypoint-initdb.d/00-isaac-create.sql:ro
+      - ./src/main/resources/db_scripts/postgres-rutherford-functions.sql:/docker-entrypoint-initdb.d/01-isaac-functions.sql:ro
```
Morpheu5: This is precisely the reason why using a custom built image is preferable in this case.

PR #196, opened 2019-01-09
Refactor validators

jsharkey13:
In writing up how the validators work, it became clear that there is a fair deal of duplicated code shared between them; but that they are also full of oddities from their initial designs. This brings them into a more consistent design and updates or removes incorrect documentation and comments.

There's still two duplicated methods that ought to be in a shared file, but I'm not sure where. Since we use Java 8, we could commit the abomination of adding methods to the `IValidator` interface itself? Or use a utility class as for [`RequestIPExtractor`](https://github.com/ucam-cl-dtg/isaac-api/blob/master/src/main/java/uk/ac/cam/cl/dtg/util/RequestIPExtractor.java) or suchlike?

I also quite like the `exactStringMatch(...)` method `IsaacNumericValidator` has, but I don't think that can be generalised because we want to compare different properties depending on what type of question it is (and it's only a separate method in numeric validation because the flow is so complicated already).

---

**Pull Request Check List**

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
@@ -15,16 +15,9 @@
  */
 package uk.ac.cam.cl.dtg.isaac.quiz;
 
-import java.math.BigDecimal;
-import java.util.Collections;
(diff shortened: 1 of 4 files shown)


jsharkey13:
We went with default methods since they don't rely on internal state.

bh412:
Looks sensible (refactoring methods out)

PR #200, opened 2019-01-23
Feature/ou free text question

mlt47:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13:
I'd like to ensure that we only use one of the wildcard syntaxes, and automatically escape the other syntax under the hood. So I'd use just `*` and `.`, and escape `&` and `#` to allow them as ordinary characters. That's much more future proof and consistent.

PR #214, opened 2019-03-26
Add graph sketcher validator and specification generator

robinmessage:
I'm happy for this to just be a starting point and take whatever edits.

There are two commits: one to add the validator and one to add the feature specification generator. I've tried to add that in a generic way but happy to amend.

---

**Pull Request Check List**
- [ ] N/A - Security - New dependency - configured sensibly not relying on defaults
- [ ] N/A - Security - New dependency - Searched for any know vulnerabilities
- [ ] N/A - Security - New dependency - Signed up team to mailing list
- [ ] N/A - Security - New dependency - Added to dependency list
- [ ] N/A - DB schema changes - postgres-rutherford-create-script updated
- [ ] N/A - DB schema changes - upgrade script created matching create script
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/api/AbstractSegueFacade.java
```diff
@@ -230,4 +230,20 @@ public static boolean isUserStaff(final UserAccountManager userManager, final Ht
...
+     *            - Instance of User Manager
+     * @param request
+     *            - with session information
+     * @return true if user is logged in as an admin or content editor, false otherwise.
+     * @throws NoUserLoggedInException
+     *             - if we are unable to tell because they are not logged in.
+     */
+    public static boolean isUserAnAdminOrContentEditor(final UserAccountManager userManager,
+                                                      final HttpServletRequest request) throws NoUserLoggedInException {
+        return userManager.checkUserRole(request, Arrays.asList(Role.ADMIN, Role.CONTENT_EDITOR));
```
jsharkey13: I think there's an `isUserStaff(...)` at the facade level, too?

robinmessage: Ah I didn't realise that was the right access level; changed it.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacGraphSketcherValidator.java
```diff
@@ -1,48 +1,161 @@
...
+        if (null == graphSketcherQuestion.getChoices() || graphSketcherQuestion.getChoices().isEmpty()) {
+            log.error("Question does not have any answers. " + question.getId() + " src: "
+                + question.getCanonicalSourceFile());
+
+            feedback = new Content("This question does not have any correct answers");
+        }
+
+        // STEP 1: Did they provide an answer at all?
+
+        GraphAnswer graphAnswer = null;
```
sacummins: I'm just wondering if there is a nicer way of encapsulating the GraphAnswer library do as an isaac Choice / Answer class so that this mapping can be done as the object comes into RESTEasy. 

This is fine for now though; it is just a thought.

src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
```diff
@@ -353,4 +353,59 @@ public Response answerQuestion(@Context final HttpServletRequest request,
...
+     * @param jsonAnswer
+     *            - answer body which will be parsed as a Choice and then converted to a ChoiceDTO.
+     * @return Response containing a QuestionValidationResponse object or containing a SegueErrorResponse .
+     */
+    @POST
+    @Path("generateSpecification")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Turn an answer into a question specification.")
+    public Response generateSpecification(@Context final HttpServletRequest request, final String jsonAnswer) {
```
sacummins: I was wondering about the difference between the specification generated here and what the student submits / what the student's answer actually turns into later. Would there be value in aligning these or converting what the student submits earlier in the API flow?

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacGraphSketcherValidator.java
robinmessage: Possibly. I just took what was already here and we were keen the library could develop separately so that's why the structures are split across the two places a bit.

src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java
robinmessage: If I've understood you correctly, then that might make sense except that: what the student submits can be _checked_ against the specification but can't be turned into something similar to the specification for reasons. So I think the current shape is correct.

(Reasons are: a line can be counted as going through multiple sectors at any point, so the structure of the user's line is different to the structure of the specification.)

robinmessage:
@jsharkey13 I'm not going to merge this myself, but it looks like @sacummins is happy with it 😄

jsharkey13:
I think we're blocked on @mlt47 doing the other Maven releases, and then it being tested once that's possible.

PR #215, opened 2019-04-08
Removes @JsonIgnore from individual questionDTOs

bh412:
@JsonIgnore added into parent DTOs for getAnswer() and getChoices() methods so they aren't accidentally passed through to new question types. Can be manually given with @JsonIgnore(false)

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacMultiChoiceQuestionDTO.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacMultiChoiceQuestionDTO.java
@@ -31,7 +31,9 @@ public class IsaacMultiChoiceQuestionDTO extends IsaacQuestionBaseDTO {
 
     /**
      * Gets the choices.
-     * 
+     * Unignores getting the choices as they are required
+     * to be presented in multi-choice questions
+     *
      * @return the choices
      */
     @Override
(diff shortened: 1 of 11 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacMultiChoiceQuestionDTO.java
```diff
@@ -27,9 +29,15 @@
...
+     * Gets the choices.
+     * 
+     * @return the choices
+     */
     @Override
-    public ContentBaseDTO getAnswer() {
-        return super.getAnswer();
+    @JsonIgnore(false)
+    public List<ChoiceDTO> getChoices() {
+        return super.choices;
```
jsharkey13: This is even worse than before! You need to use the getter, and not directly access the internal state.
```suggestion
        return super.getChoices();
```

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacQuickQuestionDTO.java
```diff
@@ -15,13 +15,21 @@
...
 
+
 /**
  * Quick Question DTO.
  * 
  */
 @JsonContentType("isaacQuestion")
 public class IsaacQuickQuestionDTO extends IsaacQuestionBaseDTO {
 
+    @Override
```
jsharkey13: I'd add some comment here explaining why you're unignoring it!

src/main/java/uk/ac/cam/cl/dtg/segue/dto/content/QuestionDTO.java
```diff
@@ -39,10 +42,11 @@ public QuestionDTO() {
     }
 
     /**
-     * Gets the answer.
-     * 
+     * Gets the answer, defaults to not return to the user.
+     * Manually unignored e.g. quick questions
```
jsharkey13: This is a bit too terse to be clear. You mean it needs to be manually unignored for questions where the answer should be public.

PR #222, opened 2019-06-07
Upversioned some dependencies in the pom file - resteasy and jackson

sacummins:
This should be fully regression tested but does seem to work on initial local testing.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,9 +11,11 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<segue.version>2.8.7-SNAPSHOT</segue.version>
-		<resteasy.version>3.0.11.Final</resteasy.version>
+		<resteasy.version>4.0.0.Final</resteasy.version>
+		<jackson.version>2.9.9</jackson.version>
 		<powermock.version>1.6.3</powermock.version>
 		<swagger-ui-version>3.19.5</swagger-ui-version>
+		<jetty-maven-version>9.4.18.v20190429</jetty-maven-version>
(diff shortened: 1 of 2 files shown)


jsharkey13:
Jetty 9.4 has radically changed the way sessions work; we use these as anonymous identifiers and rely on the session duration to align with our constant for cache usage (~40 minutes).

The new session IDs are now prefixed by "node0" followed by the same style of random identifier as we used to use. Maybe we can cope with this, however silly it looks.
A larger problem is that, upon reading about the new style sessions, they last only for 10 minutes rather than the 30 they used to. More info [in the documentation](https://www.eclipse.org/jetty/documentation/9.4.x/session-configuration-housekeeper.html). I cannot work out how or where to configure either of those options, and I have tried several places.

Yet another problem I discovered, since I thought the issue was perhaps with the `jetty-maven-plugin` we were using (since that has to pass in the Jetty config somehow), is that `jetty-maven-plugin` is not recommended for production use: https://www.eclipse.org/jetty/documentation/9.4.x/jetty-maven-plugin.html

We could fall back to Jetty `9.3.27.v20190418` which seems to still have old-style sessions, and keep the other changes (which seem to add some extra log output at startup, but doesn't seem to break anything).

@sacummins - any opinions?

sacummins:
@jsharkey13 I think you are right. Roll back the jetty maven update and make me a ticket to sort out the problem properly. We probably shouldn't be using it if it isn't recommended for prod. Also even if we do decide to use it, I want to sort out how we deal with anonymous users (sessions) anyway so that we can scale the API more easily.

PR #224, opened 2019-06-10
Feature/topic page augmented content

jsharkey13:


src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
```diff
@@ -574,30 +577,30 @@ public final Response getTopicSummaryPage(@Context final Request request,
         if (null != summaryPageId) {
             fieldsToMatch.put(ID_FIELDNAME + "." + UNPROCESSED_SEARCH_FIELD_SUFFIX, Collections.singletonList(summaryPageId));
         }
-        Response result = this.findSingleResult(fieldsToMatch);
+
+        AbstractSegueUserDTO user = userManager.getCurrentUser(httpServletRequest);
```
jsharkey13: The only thing about doing this first is that if there's no page found, you wait ages whilst it loads all your question attempts out of the database _before_ it then fails to find anything in elasticsearch.

PR #225, opened 2019-06-17
Alter rounding behaviour of sig fig checking code after FP issues

jsharkey13:
A question relies on the value of 0.2425 to 3sf, which ought to be 0.243 but when doing the rounding it gets parsed into something like 0.2424999999997 which then rounds down and not up.

To fix this, use BigDecimals as we do when working out the number of sig figs to use; this can exactly represent all of the cases we're likely to encounter on Isaac. We can also be explicit about the rounding scheme "round half away from zero" that we want to use.

This will need additional testing, but I've added the case to the unit tests too.


jsharkey13:
(Ensure you [view the diff with "Hide whitespace changes"](https://github.com/isaacphysics/isaac-api/pull/225/files?w=1) set!)

jsharkey13:
Moving entirely to arbittary precision libraries seems quite sensible, but not in this PR. So I think this is good to go!

PR #229, opened 2019-07-10
Sanitises given and last names presented in emails

bh412:


src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
```diff
@@ -218,8 +219,9 @@ public final Response getEmailInBrowserById(@Context final HttpServletRequest re
             previewProperties.putAll(emailManager.flattenTokenMap(userPropertiesMap, Maps.newHashMap(), ""));
```
mlt47: The user propertiesMap contain most user fields including givenName and familyName which then get added to the previewProperties without being escaped. So if a user was to use `givenname` it would be escaped, if they were to use `givenName` it would not be escaped.

mlt47: This is also true in the email manager, so when we fix this it would be good to also remove the duplicated logic.

pom.xml
```diff
@@ -251,6 +251,11 @@
 			<artifactId>commons-collections4</artifactId>
 			<version>4.3</version>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-text</artifactId>
```
sacummins: This is fine but you may consider trying out https://jsoup.org/cookbook/cleaning-html/whitelist-sanitizer 

You can set it to reject all html. Not sure if this will also cover javascript escaping?

src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
sacummins: It would be nice if we could get rid of the legacy givenname and use the camelcase properties. I remember we added it in as both mainly because we made a mistake once and used givenname. Would require a find and replace over the content repo though.

src/main/java/uk/ac/cam/cl/dtg/segue/comm/EmailManager.java
```diff
@@ -154,7 +155,7 @@ public void sendTemplatedEmailToUser(final RegisteredUserDTO userDTO, final Emai
 
         // default properties
         //TODO: We should find and replace this in templates as the case is wrong.
```
sacummins: Can we arrange to do this prior to this merge and therefore remove all reference to givenname

src/main/java/uk/ac/cam/cl/dtg/segue/comm/EmailManager.java
```diff
@@ -227,8 +229,9 @@ public void sendCustomEmail(final RegisteredUserDTO sendingUser, final String co
             p.putAll(this.flattenTokenMap(userPropertiesMap, Maps.newHashMap(), ""));
 
             // TODO - Remove once content templates have all been replaced with correct tokens
-            p.put("givenname", user.getGivenName() == null ? "" : user.getGivenName());
-            p.put("familyname", user.getFamilyName() == null ? "" : user.getFamilyName());
+            // Sanitize user name inputs
+            p.put("givenname", user.getGivenName() == null ? "" : StringEscapeUtils.escapeHtml4(user.getGivenName()));
+            p.put("familyname", user.getFamilyName() == null ? "" : StringEscapeUtils.escapeHtml4(user.getFamilyName()));
```
mlt47: We can remove these (also remove the comment to avoid future confusion).

PR #233, opened 2019-08-02
Enable groups assignments to be returned in a single API request

sacummins:
Fix to enable fully resolved gameboards for a given group to be returned when requesting a groups assignments (in a single api request).

---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -262,9 +267,13 @@ public Response getAssigned(@Context final HttpServletRequest request,
                 Collection<AssignmentDTO> allAssignmentsSetToGroup
                         = this.assignmentManager.getAssignmentsByGroup(group.getId());
 
+                // we currently need to use the currently logged in users information to as a parameter to get all the pass mark information.
```
jsharkey13: I don't like this, since the vast majority of this data is never used by the front-end at all. On the other hand, better to do this once than the N times we currently do it for the N boards!

PR #239, opened 2019-09-03
Feature/refactor item validators

mlt47:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


mlt47:
One thing to consider: if `IsaacParsonsValidator` is a more specific `IsaacItemQuestionValidator` and therefore does strictly more validation than the other (now and probably in the future), then inheritance between the two classes might be the better Object Oriented solution to sharing code. `validateQuestionResponse(...)` can then be overridden in the Parsons validator and call the super method to also do the checks that the item question validator runs.
This removes the need for some conditional branches on types, therefore stopping `IsaacItemQuestionValidator` needing to know about its subclasses. With inheritance, adding a new subclass in the future, for example  `ClozeQuestionValidator`, would not require editing the parent class (and will stop it getting too bloated) - open-close principle. (@jsharkey13 what we were talking about yesterday)

jsharkey13:
I don't think we gain enough from this to go forward any longer. Closing, but not deleting the branch in case we review later on!

PR #244, opened 2019-10-10
Glossary terms

Morpheu5:
* 7b4fd2af Fix up javadoc
* a08ba891 Return exactly one glossary term when searching by one term id.
* a9db4538 Add two endpoints for retrieving glossary terms
* c3dae46e Merge branch 'master' into feature/glossary-terms
* 2c698dfb Add a new Glossary Term content piece.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


override-api-web.xml
```diff
@@ -31,7 +31,7 @@
         </init-param>
         <init-param>
             <param-name>allowedOrigins</param-name>
-            <param-value>(https://*.isaac(physics|computerscience).org|http://localhost:800[0123])</param-value>
+            <param-value>(https://*.isaacphysics.org|http://localhost:800[0123.idea]|https://isaac(physics|chemistry|biology|computerscience).eu.ngrok.io)</param-value>
```
bh412: Revert this (James/ Meurig input) as it will break Isaac CS

.factorypath
```diff
@@ -0,0 +1,182 @@
+<factorypath>
```
mlt47: Does this and the classpath changes need to be committed? I imagine some of the classpath changes is fine but maybe we should gitignore this file

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * @param limit      - Maximum amount of terms to retrieve. Used for pagination.
+     * @param startIndex - Index from which to start retrieving when results exceed limit.
+     *
+     * @return Paginated list of glossary terms.
+     */
+    @GET
+    @Path("terms")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Get all the glossary terms that are indexed.")
```
mlt47: If we have more terms than the DEFAULT_RESULTS_LIMIT this will not return all of the glossary terms. Could lead to some hard to find bugs. Do we need the start_index and limit for this query?

override-api-web.xml
Morpheu5: Ah, poop.

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            if (null != startIndex) {
+                startIndexOfResults = Integer.parseInt(startIndex);
+            } else {
+                startIndexOfResults = 0;
+            }
+
+            c = this.contentManager.findByFieldNames(this.contentIndex, fieldsToMatch, startIndexOfResults, resultsLimit);
+            return Response.ok(c).build();
```
mlt47: Probably worth adding an etag cache for this endpoint if it is going to be hit many times as it will not change between content version changes.

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
```diff
@@ -0,0 +1,167 @@
...
+            c = this.contentManager.getContentDOById(this.contentIndex, term_id);
+            if (null == c) {
+                SegueErrorResponse error = new SegueErrorResponse(Status.NOT_FOUND, "No glossary term found with id: " + term_id);
+                log.debug(error.getErrorMessage());
+                return error.toResponse();
+            }
+        } catch (NumberFormatException e) {
+            return new SegueErrorResponse(Status.BAD_REQUEST,
+                    "Unable to convert one of the integer parameters provided into numbers. "
+                            + "Params provided were: limit" + DEFAULT_RESULTS_LIMIT + " and startIndex 0", e).toResponse();
```
mlt47: This might just be a checked exception from getContentDOById(...) but you're not converting any parameters to numbers so this should never happen.

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        Content c = null;
+        try {
+            c = this.contentManager.getContentDOById(this.contentIndex, term_id);
+            if (null == c) {
+                SegueErrorResponse error = new SegueErrorResponse(Status.NOT_FOUND, "No glossary term found with id: " + term_id);
+                log.debug(error.getErrorMessage());
+                return error.toResponse();
+            }
```
mlt47: Again an etag cache would be useful if we think this endpoint will be hit quite often

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
Morpheu5: The object returned is something like this:

```
{
  "results": [
    // ...
  "totalResults": 125
}
```

this is for any limit that is passed with the query, so if you get `DEFAULT_RESULTS_LIMIT` (which is 10, I think) you should also check `totalResults` to see if there are more available, and then it's your responsibility to query for more, using the `start_index` parameter.

Currently, I hard-coded 10000 in the `isaac-cs-app`, which I believe is well beyond what we could even get as far as glossary terms go, but I agree that this can lead to weird bugs.

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
Morpheu5: Nowhere do we tag errors, though…

PR #246, opened 2019-10-18
Add a content error for images larger than 300KB

jsharkey13:
We've had images over a megabyte included in the content before, and they both slow the page down and use an unnecessary amount of bandwidth for users. Hopefully this will discourage PNGs inside SVGs too!

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
@@ -676,6 +676,7 @@ public class ContentIndexer {
                 if (f.getSrc() != null && !f.getSrc().startsWith("http")) {
                     ByteArrayOutputStream fileData = null;
                     try {
+                        // This will return null if the file is not found:
                         fileData = database.getFileByCommitSHA(sha, f.getSrc());
                     } catch (IOException | UnsupportedOperationException e) {
                         // Leave fileData = null;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
```diff
@@ -671,10 +673,21 @@ private boolean checkForContentErrors(final String sha, final Map<String, Conten
...
+                if (f.getSrc() != null && !f.getSrc().startsWith("http")) {
+                    ByteArrayOutputStream fileData = null;
+                    try {
+                        fileData = database.getFileByCommitSHA(sha, f.getSrc());
+                    } catch (IOException | UnsupportedOperationException e) {
+                        // Leave fileData = null;
+                    }
+                    if (null == fileData) {
+                        this.registerContentProblem(c, "Unable to find Image: " + f.getSrc()
+                                + " in Git. Could the reference be incorrect? SourceFile is " + c.getCanonicalSourceFile(), indexProblemCache);
```
sacummins: Looks good - just wondering why you don't want this in the the catch block? And the size check in the try block? It might look cleaner? Just a matter of style though so happy with this PR either way.

jsharkey13: It has to be like this to avoid duplication, since `database.getFileByCommitSHA(....)` can also return `null` which needs to be registered as the same type of content problem.

PR #248, opened 2019-11-22
Add linked gameboards to topic summary pages

jsharkey13:
This allows topic summary pages to have a list of linked gameboards specified, which gets augmented at request time. This augmentation isn't nice and requires loading both the DO and DTO separately from ElasticSearch. This is because the mapping at the Facade layer cannot cope with mapping lists of IDs into lists of objects with those IDs.

It is fully backwards compatible, so long the old ETL is used and nobody adds `linkedGameboards` to an existing topic summary until after the old API is gone.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
@@ -27,6 +27,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.managers.GameManager;
 import uk.ac.cam.cl.dtg.isaac.api.managers.URIManager;
+import uk.ac.cam.cl.dtg.isaac.dos.IsaacTopicSummaryPage;
+import uk.ac.cam.cl.dtg.isaac.dto.GameboardDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacQuestionPageDTO;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
```diff
@@ -490,46 +492,59 @@ public final Response getTopicSummaryPage(@Context final Request request,
...
-                ImmutableMap<String, String> logEntry = new ImmutableMap.Builder<String, String>()
-                        .put(PAGE_ID_LOG_FIELDNAME, summaryPageId)
-                        .put(CONTENT_VERSION_FIELDNAME, this.contentManager.getCurrentContentSHA()).build();
-                getLogManager().logEvent(user, httpServletRequest, IsaacLogType.VIEW_TOPIC_SUMMARY_PAGE, logEntry);
+            // Augment related questions with attempt information:
+            userQuestionAttempts = questionManager.getQuestionAttemptsByUser(user);
+            this.augmentContentWithRelatedContent(this.contentIndex, topicSummaryDTO, userQuestionAttempts);
+
+            // Augment linked gameboards using the list in the DO:
+            // FIXME: this requires loading both the DO and DTO separately, since augmenting things is hard right now.
```
sacummins: I think the nicer fix is likely to involve more work and adding special content mapping rules/deserialisers under ContentMapper.java generateNewPreconfiguredContentMapper.

This will do for now though

PR #250, opened 2019-12-06
Add support for weekly streaks on the user progress endpoint

jsharkey13:
We don't want to use these yet, but being able to access them on the site will be useful testing.


jsharkey13:
The "database resource leak" issue is the same as every other database connection; they will all need addressing, but JDBC will close connections idle for too long; so this can wait until we've agreed a plan for all of them.

PR #251, opened 2019-12-11
Allow teachers to reserve spaces for batches of students onto events

Morpheu5:
- [x] Add an event booking status: RESERVED
- [x] Reserved bookings will work exactly as waiting list, they will take up an available place.
- [x] Will need an extra field to record reservedBy and reservationCloseDate.
  - [x] reserved by
  - [x] reservationCloseDate -- I'm going to use `additional_booking_information` for this as it isn't data that we use for searching/indexing.
- [x] Reservation close date will be 14 days from the time of reservation or the date of the event, whichever is sooner.
  - [x] A cron job will sweep the database once a day, presumably.
- [x] The event will need to record student event group limit
  - [x] Needs editor changes
  - [x] There is a "hardcoded" default of 10

- [x] When a reservation is made, the api will create a new event booking record with the status of RESERVED for the student.
- [x] CANCELLED reservations do not count toward the reservation limit.
  - [x] check that this is true (it may already be)
- [x] If a reservation is cancelled by the teacher an email is sent to the student and the reservation is cancelled.
- [x] We should make sure that the reservation is valid by the teacher having an association with the student and that the number of reservations doesn't exceed the max number of reservations provided by the event.
- [x] That will send an email to the student saying that the teacher has reserved a space for them and that they will have n days to take up the reservation or cancel it.

- [x] If a batch reservation is attempted and there are not enough spaces remaining for the entire batch, the entire batch fails to make a reservation and the teacher will get an explanation and an opportunity to retry with fewer students.
- [x] Students will receive the reservation email, follow the link to the booking form where they can add their personal information to confirm the booking or cancel. Both actions should notify the teacher (reservedBy field).
  - [x] Students will receive a reservation email (NEEDS correct email template).
  - [x] Students will follow the link to the booking form and complete the booking.
  - [x] Students will follow the link to the booking form and cancel the reservation.
  - [x] Confirming and cancelling should email the teacher (identified by the reservedBy field)

---

**Pull Request Check List**
- [x] NEEDS Unit Tests & Regression Tests
  - [x] Broken tests were fixed
- [x] DB schema changes - make sure they make sense
- [x] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -486,6 +497,50 @@ public final Response getEventBookingByEventId(@Context final HttpServletRequest
...
+                    .stream().map(RegisteredUserDTO::getId)
+                    .collect(Collectors.toList());
+
+            // Filter eventBookings based on whether the booked user is a member of the given group
+            List<EventBookingDTO> eventBookings = bookingManager.getBookingByEventId(eventId)
+                    .stream().filter(booking -> groupMemberIds.contains(booking.getUserBooked().getId()))
+                    .collect(Collectors.toList());
+
+            // Event leaders are only allowed to see the bookings of connected users
+            if (Role.EVENT_LEADER.equals(currentUser.getRole())) {
```
sacummins: Should this check not apply to teachers too? I.e. teachers should only see information about the users they have been given permission to view?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -486,6 +497,50 @@ public final Response getEventBookingByEventId(@Context final HttpServletRequest
...
+                eventBookings = userAssociationManager.filterUnassociatedRecords(
+                        currentUser, eventBookings, booking -> booking.getUserBooked().getId());
+            }
+
+            // af599 TODO: This seems to return UserSummaryWithEmailAddressDTOs even when the request is done as a TEACHER.
+
+            return Response.ok(eventBookings).build();
+        } catch (Exception e) {
```
sacummins: ... spotted your todo below... Yeah probably best not to catch all exceptions.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -552,6 +609,108 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+            currentUser = userManager.getCurrentRegisteredUser(request);
+            if (!Arrays.asList(Role.TEACHER, Role.EVENT_LEADER, Role.EVENT_MANAGER, Role.ADMIN).contains(currentUser.getRole())) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+            event = this.getEventDTOById(request, eventId);
+
+            // af599 TODO: Make sure this makes sense, adapt it, or remove it
+            /* if (!bookingManager.isUserAbleToManageEvent(currentUser, event)) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            } */
```
sacummins: Yeah would be good to have a permissions check here.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -552,6 +609,108 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+            String errorMsg = "Database error occurred while trying to reserve space for a user onto an event.";
+            log.error(errorMsg, e);
+            return new SegueErrorResponse(Status.INTERNAL_SERVER_ERROR, errorMsg).toResponse();
+        } catch (ContentManagerException e) {
+            return new SegueErrorResponse(Status.INTERNAL_SERVER_ERROR,
+                    "Content Database error occurred while trying to retrieve all event booking information.")
+                    .toResponse();
+        }
+
+        try {
```
sacummins: Could this try block be merged with the above one?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -392,6 +399,101 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+                    reservation = this.bookingPersistenceManager.createBooking(event.getId(), user.getId(), reservingUser.getId(),
+                            BookingStatus.RESERVED, null);
+                }
+                reservations.add(reservation);
+
+                try {
+                    // af599 TODO: Use the correct email template here.
+                    // af599 TODO: Content team action.
+                    emailManager.sendTemplatedEmailToUser(user,
+                            emailManager.getEmailTemplateDTO("email-event-booking-confirmed"),
```
sacummins: Don't forget to update the template before merging

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -392,6 +399,101 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+                                    .put("authorizationLink", String.format("https://%s/account?authToken=%s",
+                                            propertiesLoader.getProperty(HOST_NAME), event.getIsaacGroupToken()))
+                                    .put("event.emailEventDetails", event.getEmailEventDetails() == null ? "" : event.getEmailEventDetails())
+                                    .put("event", event)
+                                    .build(),
+                            EmailType.SYSTEM,
+                            Collections.singletonList(generateEventICSFile(event, reservation)));
+
+                } catch (SegueDatabaseException | ContentManagerException e) {
+                    log.error(String.format("Unable to send welcome email (%s) to user (%s)",
```
sacummins: might want to update from welcome email to event email - I suspect there are a number of places in the file with this copy and paste error (many of them my fault)

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -392,6 +399,101 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+
+                // auto add them to the group and grant the owner permission
+                if (event.getIsaacGroupToken() != null && !event.getIsaacGroupToken().isEmpty()) {
+                    try {
+                        this.userAssociationManager.createAssociationWithToken(event.getIsaacGroupToken(), user);
+                    } catch (InvalidUserAssociationTokenException e) {
+                        log.error(String.format("Unable to auto add user (%s) using token (%s) as the token is invalid.",
+                                user.getEmail(), event.getIsaacGroupToken()));
+                    }
+                }
```
sacummins: Not sure about the auto granting of user association permissions for reservations when this could bypass the philosophy of students always having to positively grant access to users. Is it possible that this endpoint could be misused?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -392,6 +399,101 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+            }
+            return reservations;
+
+        } catch (EventIsFullException | SegueDatabaseException e) {
+            for (EventBookingDTO reservation : reservations) {
+                try {
+                    bookingPersistenceManager.deleteBooking(event.getId(), reservation.getUserBooked().getId());
+                } catch (SegueDatabaseException f) {
+                    // Tough luck?
```
sacummins: Yeah - although with the events system we could use real database transactions if we wanted to push more logic down to the persistence manager.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
```diff
@@ -111,8 +111,8 @@ public EventBookingDTO getBookingByEventIdAndUserId(final String eventId, final
      * @throws SegueDatabaseException
      *             - if an error occurs.
      */
-    public EventBookingDTO updateBookingStatus(final String eventId, final Long userId, final BookingStatus bookingStatus, final Map additionalEventInformation) throws SegueDatabaseException {
-        dao.updateStatus(eventId, userId, bookingStatus, additionalEventInformation);
+    public EventBookingDTO updateBookingStatus(final String eventId, final Long userId, final Long reservedById, final BookingStatus bookingStatus, final Map additionalEventInformation) throws SegueDatabaseException {
```
sacummins: What is the reservedById parameter for? I noticed null being passed in a few times in some method calls. If is nullable maybe we could annotated it as such / add explanation in the javadoc? I assume it is to let us record who made the booking/reservation? In which case maybe it could be called that and always populated - even if it is just the same as the userId?

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
```diff
@@ -188,7 +188,7 @@ public boolean isUserBooked(final String eventId, final Long userId) throws Segu
         try {
             final EventBooking bookingByEventAndUser = dao.findBookingByEventAndUser(eventId, userId);
             List<BookingStatus> bookedStatuses = Arrays.asList(
-                    BookingStatus.CONFIRMED, BookingStatus.ATTENDED, BookingStatus.ABSENT);
+                    BookingStatus.CONFIRMED, BookingStatus.ATTENDED, BookingStatus.ABSENT, BookingStatus.RESERVED);
```
sacummins: We should be careful about the meaning of this method. Before it meant that a user had a confirmed booking. Are we defining reserved as definitely confirmed now?

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
```diff
@@ -255,7 +255,13 @@ private EventBookingDTO convertToDTO(final EventBooking eb, final IsaacEventPage
         try {
             UserSummaryWithEmailAddressDTO user = userManager.convertToDetailedUserSummaryObject(userManager.getUserDTOById(eb
                     .getUserId()), UserSummaryWithEmailAddressDTO.class);
-
+            Long reservedById = eb.getReservedBy();
+            // af599 TODO: Make this null, not 0L...
+            // af599 honestly can't remember the circumstances of this. This *should* come back as null, not 0L...
+            if (!reservedById.equals(0L)) {
```
sacummins: :S 0L is weird. Is postgres set to default to that or null? Is it a foreign key to userId?

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/eventbookings/PgEventBookings.java
```diff
@@ -82,22 +78,27 @@ public EventBooking add(final String eventId, final Long userId, final BookingSt
...
+                    "INSERT INTO event_bookings (id, user_id, reserved_by, event_id, status, created, updated, additional_booking_information) VALUES (DEFAULT, ?, ?, ?, ?, ?, ?, ?::text::jsonb)",
                     Statement.RETURN_GENERATED_KEYS);
             pst.setLong(1, userId);
-            pst.setString(2, eventId);
-            pst.setString(3, status.name());
-            pst.setTimestamp(4, new java.sql.Timestamp(creationDate.getTime()));
+            try {
+                pst.setLong(2, reserveById);
+            } catch (NullPointerException e) {
+                pst.setNull(2, Types.INTEGER);
```
sacummins: If we made it a foreign key and always populated it with the user who made the booking / reservation we wouldn't need nasty catches for null pointers.

src/test/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManagerTest.java
```diff
@@ -93,7 +93,7 @@ public void requestBooking_checkTeacherAllowedOnStudentEventDespiteCapacityFull_
         dummyEventBookingPersistenceManager.acquireDistributedLock(testEvent.getId());
         expectLastCall().atLeastOnce();
 
-        expect(dummyEventBookingPersistenceManager.createBooking(testEvent.getId(), someUser.getId(), BookingStatus
+        expect(dummyEventBookingPersistenceManager.createBooking(testEvent.getId(), someUser.getId(), null, BookingStatus
```
sacummins: Would be good to have a couple tests to make sure reserved bookings behave properly when events are full / nearly full.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: I'll file this under "it was already there", as in fact it is already there in many other places. It did struck me as odd, but I didn't have time to question existing code :)

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
Morpheu5: Makes sense unless there is business reason for having this information associated with a "reservation" event.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
Morpheu5: Good point – I saw it as a way of retrieving users that are associated with an event and blissfully ignored any implied meaning. We could have a separate method for reservations and do both checks only where necessary.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
mlt47: I think this should only be done when the student confirms their place on the event, not when the teacher reserves a space for their student on the event.
On the booking form there is bit of text explaining that they give permission to the event managers to see their work and relevant details, they will need to go through the booking form to promote from reserved to confirmed.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
Morpheu5: I forgot to set the default to null. Now I did and re-tested it, and null values still come back as `0L`.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Fixed in 4cb805b.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
Morpheu5: There may be a business reason after all, in case we want to enforce the maximum limit of reservations per event, per teacher, on the number of participants that have been reserved by a certain teacher, rather than the ones who currently have a reservation that will be either confirmed or cancelled later. @mlt47 @jsharkey13

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -734,29 +924,53 @@ public boolean hasBookingWithAnyOfStatuses(final String eventId, final Long user
...
+
+        Long reservedById = null;
         try {
             // Obtain an exclusive database lock to lock the booking
             this.bookingPersistenceManager.acquireDistributedLock(event.getId());
-            this.bookingPersistenceManager.updateBookingStatus(event.getId(), user.getId(),
+            UserSummaryDTO reservedBy = this.bookingPersistenceManager
+                    .getBookingByEventIdAndUserId(event.getId(), user.getId()).getReservedBy();
+            reservedById = reservedBy == null ? null : reservedBy.getId();
+            BookingStatus previousBookingStatus = this.getBookingStatus(event.getId(), user.getId());
```
jsharkey13: This loads the same event booking at least twice, only to write back a new version of it. That seems unnecessary?

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
```diff
@@ -255,7 +264,11 @@ private EventBookingDTO convertToDTO(final EventBooking eb, final IsaacEventPage
         try {
             UserSummaryWithEmailAddressDTO user = userManager.convertToDetailedUserSummaryObject(userManager.getUserDTOById(eb
                     .getUserId()), UserSummaryWithEmailAddressDTO.class);
-
+            Long reservedById = eb.getReservedBy();
+            if (reservedById != null && !reservedById.equals(0L)) {
+                UserSummaryWithEmailAddressDTO reservingUser = userManager.convertToDetailedUserSummaryObject(userManager.getUserDTOById(reservedById), UserSummaryWithEmailAddressDTO.class);
```
jsharkey13: I think this is dangerous, and I'd rather the DTO did not contain this information unless you add the requisite permissions framework to check wether the current user is allowed to see the reserving user's PII!

jsharkey13: Just the user ID is enough in the DTO!

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/eventbookings/PgEventBookings.java
```diff
@@ -111,28 +112,38 @@ public EventBooking add(final String eventId, final Long userId, final BookingSt
...
                 pst = conn.prepareStatement("UPDATE event_bookings " +
-                    "SET status = ?, updated = ? " +
+                    "SET status = ?, updated = ?, reserved_by = ? " +
                     "WHERE event_id = ? AND user_id = ?;");
                 pst.setString(1, status.name());
                 pst.setTimestamp(2, new java.sql.Timestamp(new Date().getTime()));
-                pst.setString(3, eventId);
-                pst.setLong(4, userId);
+                try {
+                    pst.setLong(3, reservedById);
```
jsharkey13: Surely this could be an `if`, rather than a try-catch? Java is not Python, where exceptions are preferred control flow :P

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
```diff
@@ -634,7 +634,9 @@ public Response getUserIdToSchoolMap(@Context final HttpServletRequest httpServl
                                          @QueryParam("user_ids") final String userIdsQueryParam) {
         try {
             RegisteredUserDTO currentUser = userManager.getCurrentRegisteredUser(httpServletRequest);
-            if (!isUserStaff(userManager, currentUser) && !Role.EVENT_LEADER.equals(currentUser.getRole())) {
+            // af599 TODO: Check that including TEACHERs is OK -- old code: !Role.EVENT_LEADER.equals(currentUser.getRole())
+            // af599 TODO: This may not be necessary anymore as we are not doing a school lookup when grabbing event reservations.
+            if (!isUserStaff(userManager, currentUser) && !Arrays.asList(Role.EVENT_LEADER, Role.TEACHER).contains(currentUser.getRole())) {
```
jsharkey13: I don't think this is a good idea, and if it is, then we'd need to change our Privacy Policy since this is not listed as something we share with teachers presently.

src/main/resources/db_scripts/event_bookings_add_reserved_by.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE ONLY event_bookings ADD COLUMN reserved_by integer
+        DEFAULT NULL
+        CONSTRAINT event_bookings_users_id_fk
+        REFERENCES users
```
jsharkey13: Does this not need to be `users.id`, or is Postgres magic?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -513,6 +527,53 @@ public final Response getEventBookingByEventId(@Context final HttpServletRequest
...
+                                                       @PathParam("event_id") final String eventId,
+                                                       @PathParam("group_id") final String groupId) {
+        try {
+            RegisteredUserDTO currentUser = userManager.getCurrentRegisteredUser(request);
+            if (!Arrays.asList(Role.TEACHER, Role.EVENT_LEADER, Role.EVENT_MANAGER, Role.ADMIN).contains(currentUser.getRole())) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+
+            UserGroupDTO group = groupManager.getGroupById(Long.parseLong(groupId));
+            // af599 TODO: Might make sense to check whether the currentUser is allowed to interact with the group
```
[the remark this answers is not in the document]
Morpheu5: Fixed in a9f83e70.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/eventbookings/EventBookings.java
```diff
@@ -55,7 +57,7 @@
      * @return the newly updated event booking.
      * @throws SegueDatabaseException - if the database goes wrong.
      */
-    void updateStatus(final String eventId, final Long userId, final BookingStatus status, Map<String, String> additionalEventInformation) throws SegueDatabaseException;
+    void updateStatus(final String eventId, final Long userId, final Long reservedById, final BookingStatus status, Map<String, String> additionalEventInformation) throws SegueDatabaseException;
```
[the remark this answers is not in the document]
Morpheu5: #6aede53c

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/eventbookings/PgEventBookings.java
Morpheu5: #6aede53c

src/main/resources/db_scripts/event_bookings_add_reserved_by.sql
Morpheu5: Considering I copypasted this from DataGrip, I would suspect magic, but that's a good point…

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -693,6 +762,106 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+                RegisteredUserDTO userDTOById = userManager.getUserDTOById(bookableId);
+                usersToBook.add(userDTOById);
+            }
+            List<EventBookingDTO> bookings = bookingManager.requestReservations(event, usersToBook, currentUser);
+            this.getLogManager().logEvent(currentUser, request,
+                    SegueLogType.EVENT_RESERVATIONS_CREATED,
+                    ImmutableMap.of(
+                            EVENT_ID_FKEY_FIELDNAME, event.getId(),
+                            USER_ID_FKEY_FIELDNAME, currentUser.getId(),
+                            USER_ID_LIST_FKEY_FIELDNAME, usersToBook,
```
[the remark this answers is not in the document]
Morpheu5: Well spotted. Fixed in 8940def5.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -513,6 +527,53 @@ public final Response getEventBookingByEventId(@Context final HttpServletRequest
...
+                    .stream().map(RegisteredUserDTO::getId)
+                    .collect(Collectors.toList());
+
+            // Filter eventBookings based on whether the booked user is a member of the given group
+            List<EventBookingDTO> eventBookings = bookingManager.getBookingByEventId(eventId)
+                    .stream().filter(booking -> groupMemberIds.contains(booking.getUserBooked().getId()))
+                    .collect(Collectors.toList());
+
+            // Event leaders are only allowed to see the bookings of connected users
+            if (Arrays.asList(Role.TEACHER, Role.EVENT_LEADER).contains(currentUser.getRole())) {
```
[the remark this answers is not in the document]
Morpheu5: Removed in d6a4f24e. Seems to be redundant indeed. Something similar may appear somewhere else from which I copypasted.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Fixed in b7e7fe72.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -724,12 +893,21 @@ public final Response createBookingForMe(@Context final HttpServletRequest reque
                         .toResponse();
             }
 
+            // If there is a reservation, delete the reservation before creating a booking.
+            RegisteredUserDTO reservedBy = null;
+            if (bookingManager.isUserReserved(eventId, user.getId())) {
```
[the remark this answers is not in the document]
Morpheu5: Should be fixed by e1d2b69a

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/EventBookingPersistenceManager.java
Morpheu5: Fixed in a2c97014

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -181,6 +205,29 @@ public boolean isUserAbleToManageEvent(RegisteredUserDTO user, IsaacEventPageDTO
...
+     * @throws SegueDatabaseException
+     */
+    public boolean isReservationMadeByRequestingUser(RegisteredUserDTO user,
+                                                     RegisteredUserDTO userOwningReservation,
+                                                     IsaacEventPageDTO event)
+            throws SegueDatabaseException {
+
+        EventBookingDTO booking = this.getBookingByEventIdAndUserId(event.getId(), userOwningReservation.getId());
+        UserSummaryDTO reservingUser = booking.getReservedBy();
+        if (null != reservingUser) {
```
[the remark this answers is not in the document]
Morpheu5: Because this should be only used for reservations, and because reservations should always have that attribute set to not null, this null check should not be necessary, but try telling that to Java…

But yeah, I'd rather flip the `.equals()`.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -196,6 +243,10 @@ public boolean isUserAbleToManageEvent(RegisteredUserDTO user, IsaacEventPageDTO
     public EventBookingDTO createBookingOrAddToWaitingList(final IsaacEventPageDTO event, final RegisteredUserDTO user,
                                          final Map<String, String> additionalEventInformation)
             throws SegueDatabaseException, DuplicateBookingException {
+        // If a reservation exists, delete it before creating a booking
+        if (this.isUserReserved(event.getId(), user.getId())) {
```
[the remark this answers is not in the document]
Morpheu5: I refactored this heavily in e1d2b69a – I think.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -251,6 +302,10 @@ public EventBookingDTO createBooking(final IsaacEventPageDTO event, final Regist
                                          final Map<String, String> additionalEventInformation,
                                          final BookingStatus status)
             throws SegueDatabaseException, DuplicateBookingException, EventIsFullException {
+        // If a reservation exists, delete it before creating a booking
+        if (this.isUserReserved(event.getId(), user.getId())) {
```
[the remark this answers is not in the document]
Morpheu5: See above (e1d2b69a).

src/main/java/uk/ac/cam/cl/dtg/segue/api/UsersFacade.java
Morpheu5: Fixed in 2fd148fc.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -693,6 +762,106 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+            );
+        }
+
+        List<Long> bookableIds;
+        try {
+            currentUser = userManager.getCurrentRegisteredUser(request);
+            if (!Arrays.asList(Role.TEACHER, Role.EVENT_LEADER, Role.EVENT_MANAGER, Role.ADMIN).contains(currentUser.getRole())) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+
```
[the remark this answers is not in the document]
Morpheu5: Fixed in 1022353a.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -1370,14 +1560,16 @@ private IsaacEventPageDTO getEventDTOById(final HttpServletRequest request, fina
...
+                                                                 final ContentDTO possibleEvent)
+            throws SegueDatabaseException {
         if (possibleEvent instanceof IsaacEventPageDTO) {
             IsaacEventPageDTO page = (IsaacEventPageDTO) possibleEvent;
 
             try {
                 RegisteredUserDTO user = userManager.getCurrentRegisteredUser(request);
 
-                Boolean userBooked = this.bookingManager.isUserBooked(page.getId(), user.getId());
+                Boolean userBooked = this.bookingManager.isUserBooked(page.getId(), user.getId()) || this.bookingManager.isUserReserved(page.getId(), user.getId());
```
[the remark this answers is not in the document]
Morpheu5: Fixed in 5ffa0d22.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -306,7 +321,7 @@ public final Response getEvents(@Context final HttpServletRequest request, @Quer
 
 			filteredResults.add(eventDTOById);
 		}
-        return new ResultsWrapper<>(filteredResults, new Long(filteredResults.size()));
+        return new ResultsWrapper<>(filteredResults, (long) filteredResults.size());
```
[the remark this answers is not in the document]
jsharkey13: Ok, LGTM and IntelliJ disagree with me!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -693,6 +743,188 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+            List<RegisteredUserDTO> validUsers = new ArrayList<>();
+            for (Long userId : userIds) {
+                RegisteredUserDTO userOwningBooking = userManager.getUserDTOById(userId);
+                if (userIsAbleToManageEvent || (bookingManager.isReservationMadeByRequestingUser(userLoggedIn, userOwningBooking, event) && userAssociationManager.hasPermission(userLoggedIn, userOwningBooking))) {
+                    if (bookingManager.hasBookingWithAnyOfStatuses(eventId, userId, new HashSet<>(Arrays.asList(BookingStatus.CONFIRMED, BookingStatus.WAITING_LIST, BookingStatus.RESERVED)))) {
+                        validUsers.add(userOwningBooking);
+                    } else {
+                        // Maybe silently carry on instead?
+                        return new SegueErrorResponse(Status.fromStatusCode(422), // Unprocessable Entity
+                                "Some of the reservations cannot be cancelled. Please reload the page and try again.")
```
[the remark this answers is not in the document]
jsharkey13: Ah, I've decided this is fine as is.

src/main/resources/db_scripts/event_bookings_add_reserved_by.sql
jsharkey13: You learn something new every day: https://stackoverflow.com/a/28560619

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -693,6 +743,188 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+
+            List<RegisteredUserDTO> validUsers = new ArrayList<>();
+            for (Long userId : userIds) {
+                RegisteredUserDTO userOwningBooking = userManager.getUserDTOById(userId);
+                if (userIsAbleToManageEvent || (bookingManager.isReservationMadeByRequestingUser(userLoggedIn, userOwningBooking, event) && userAssociationManager.hasPermission(userLoggedIn, userOwningBooking))) {
+                    if (bookingManager.hasBookingWithAnyOfStatuses(eventId, userId, new HashSet<>(Arrays.asList(BookingStatus.CONFIRMED, BookingStatus.WAITING_LIST, BookingStatus.RESERVED)))) {
+                        validUsers.add(userOwningBooking);
+                    } else {
+                        // Maybe silently carry on instead?
+                        return new SegueErrorResponse(Status.fromStatusCode(422), // Unprocessable Entity
```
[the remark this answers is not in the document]
Morpheu5: But that is wrong. The error here is 422, Unprocessable Entity. The fact that Java doesn't know about it does not mean that we shouldn't use the right response. And indeed all that matters is that it is a 4xx.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -513,6 +521,46 @@ public final Response getEventBookingByEventId(@Context final HttpServletRequest
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @GZIP
+    @ApiOperation(value = "List event bookings for a specific event and group.")
+    public final Response getEventBookingForGivenGroup(@Context final HttpServletRequest request,
+                                                       @PathParam("event_id") final String eventId,
+                                                       @PathParam("group_id") final String groupId) {
+        try {
+            RegisteredUserDTO currentUser = userManager.getCurrentRegisteredUser(request);
+            UserGroupDTO group = groupManager.getGroupById(Long.parseLong(groupId));
+
```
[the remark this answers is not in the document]
Morpheu5: I don't understand why, but there seems to be a permission check right after that line. Did you add it?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -693,6 +743,188 @@ public final Response createBookingForGivenUser(@Context final HttpServletReques
...
+            }
+
+            bookableUsers = new HashMap<>();
+            List<Long> unbookableIds = new ArrayList<>();
+            for (Long userId : userIds) {
+                // Do not add a reservation if a user is already booked or reserved.
+                RegisteredUserDTO u = userManager.getUserDTOById(userId);
+                if (!userAssociationManager.hasPermission(currentUser, u)) {
+                    return new SegueErrorResponse(Status.FORBIDDEN,
+                            "You are not authorized to interact with some of the user IDs specified.")
```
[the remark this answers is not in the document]
Morpheu5: Fixed in 78692d3b.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/eventbookings/EventBookings.java
```diff
@@ -30,6 +30,24 @@
...
+     *            - the user id of who made the reservation (can be null)
+     * @param status
+     *            - the initial status of the booking.
+     * @param additionalInformation - additional information required for the event.
+     * @return the newly created booking
+     * @throws SegueDatabaseException
+     *             - if an error occurs.
+     */
+    EventBooking add(final String eventId, final Long userId, final Long reservedById, final BookingStatus status, final Map<String,String> additionalInformation) throws SegueDatabaseException;
+
```
sacummins: Could the add interface method below just be extended with some nullable reservedById parameter? It is perfectly possible for an event manager to make a booking on behalf of someone else and that could be captured using the same parameter. I realise it means having to add the missing parameter to every invocation but there shouldn't be many.

Just a suggestion though. Not crucial - I see you did the nice thing in the actual implementation.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -873,6 +1084,55 @@ private void ensureCapacity(final IsaacEventPageDTO event, final RegisteredUserD
...
+     * Helper method to ensure a batch can be booked onto an event without violating space restrictions on the event.
+     * <p>
+     * If it does an exception will be thrown if a new booking wouldn't no exception will be thrown.
+     *
+     * @param event the event the user wants to book on to
+     * @param users  the users who are trying to be booked onto the event.
+     * @throws SegueDatabaseException - if an error occurs
+     * @throws EventIsFullException   - if the event is full according to the event rules established.
+     */
+    private void ensureCapacity(final IsaacEventPageDTO event, final List<RegisteredUserDTO> users) throws
```
sacummins: Are we assuming reservations whose interval has elapsed don't count towards capacity?

src/test/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManagerTest.java
```diff
@@ -11,6 +11,7 @@
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacEventPageDTO;
```
sacummins: It would be good to have a few more tests dealing with what happens when users try and reserve more than they can. I've added one for you on capacity checking and fixed the bug that it caught.

The tests could do with a refactor to remove some duplicate code but they are fairly well defined and straight forward to extend.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/eventbookings/EventBookings.java
Morpheu5: The problem is with the slightly twisted semantics. It captures the workflow without actually capturing the workflow. Right now, the presence of that field in the record indicates that there was a reservation before there was a booking. To apply the same meaning to an evento manager booking someone else, we'd have to introduce a `bookedById` rathern than a `reservedById` field.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: That should be the case but it seems like it isn't (EventBoonkingManager.java, L 818). I think somehow I assumed we'd run a cronjob to remove the expired reservations. If we don't do that, we need to change the query in PgEventBookings.java, L 316, to exclude expired reservations.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -394,6 +444,104 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+    public List<EventBookingDTO> requestReservations(final IsaacEventPageDTO event, final List<RegisteredUserDTO> users,
+                                                     final RegisteredUserDTO reservingUser)
+            throws EventDeadlineException, EmailMustBeVerifiedException, DuplicateBookingException,
+            SegueDatabaseException, EventIsFullException, EventGroupReservationLimitException {
+
+        // af599 TODO: Is it wise to do this before acquiring a database lock?
+        for (RegisteredUserDTO user : users) {
+            this.ensureValidBooking(event, user, true, BookingStatus.RESERVED);
+        }
+
```
mlt47, line 467: We should answer this. From the looks of things isUserBooked(...) and isUserReserved(...) check the database event booking records for that user. One possible problematic execution trace would be if the student completes a booking between the time that the reserving teacher does this check and then reserves that student. That would create two bookings which count towards capacity for the same student.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -394,6 +444,104 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+                    log.error(String.format("Unable to send event email (%s) to user (%s)",
+                            event.getId(), user.getEmail()), e);
+                }
+            }
+            return reservations;
+
+        } catch (EventIsFullException | SegueDatabaseException | EventGroupReservationLimitException | NullPointerException e) {
+            for (EventBookingDTO reservation : reservations) {
+                try {
+                    bookingPersistenceManager.deleteBooking(event.getId(), reservation.getUserBooked().getId());
```
mlt47, line 531: This would lose records for previously cancelled users. I think we can live with that but a database transaction would not lose that information.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -394,6 +444,104 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+
+                SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
+                Map<String, String> additionalEventInformation = new HashMap<>();
+                additionalEventInformation.put("reservationCloseDate", dateFormat.format(reservationCloseDate));
+
+                // attempt to book them on the event
+                if (this.hasBookingWithStatus(event.getId(), user.getId(), BookingStatus.CANCELLED)) {
+                    // if the user has previously cancelled we should let them book again.
+                    reservation = this.bookingPersistenceManager.updateBookingStatus(event.getId(), user.getId(),
+                            BookingStatus.RESERVED, additionalEventInformation);
```
mlt47, line 500: Updating a booking does not record the reservedBy user and so messes up the reservation limit counts a bit, and it could be confusing to have a reserved record with no reservedBy value.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -844,6 +1043,19 @@ public void resendEventEmail(final IsaacEventPageDTO event, final RegisteredUser
                             .put("event", event)
                             .build(),
                     EmailType.SYSTEM);
+        } else if (booking.getBookingStatus().equals(BookingStatus.RESERVED)) {
+            emailManager.sendTemplatedEmailToUser(user,
+                    emailManager.getEmailTemplateDTO("email-event-reservation-requested"),
+                    new ImmutableMap.Builder<String, Object>()
+                            .put("reservingUser", "")  // FIXME: no UserManager to find details of booking.getReservedById()
```
mlt47, line 1050: Unresolved FIXME

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
mlt47: We should implement one of these solutions

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -899,12 +1160,17 @@ private void ensureValidBooking(final IsaacEventPageDTO event, final RegisteredU
...
+            throw new DuplicateBookingException(String.format("Unable to reserve onto event (%s) as user (%s) is"
+                    + " already reserved on to it.", event.getId(), user.getEmail()));
+        }
         // check if already booked
-        if (this.isUserBooked(event.getId(), user.getId())) {
+        if (requestedBookingStatus.equals(BookingStatus.CONFIRMED) && this.isUserBooked(event.getId(), user.getId())) {
             throw new DuplicateBookingException(String.format("Unable to book onto event (%s) as user (%s) is already"
                     + " booked on to it.", event.getId(), user.getEmail()));
         }
+        // should we check if already on waiting list too?
```
mlt47, line 1173: Unresolved question

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacEventPage.java
```diff
@@ -431,4 +441,30 @@ public String getEmailWaitingListBookingText() {
...
+	 * Set the maximum number of reservations per event that a teacher can request.
+	 * @param groupReservationLimit
+	 */
+	public void setGroupReservationLimit(Integer groupReservationLimit) {
+		this.groupReservationLimit = groupReservationLimit;
+	}
+
+	public Boolean getAllowGroupReservations() {
+		return allowGroupReservations;
+	}
```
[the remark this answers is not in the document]
Morpheu5: Fixed in ca8744e7.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Option 2 is probably the easiest but I can't recall what I was looking at when I referred to that query and I'm now lost as that line is now an empty line…

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Fixed in e251ca97.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: If we have facilities for defining transactions via Java, then I'm happy to consider this option. However, considering this is done after acquiring a lock, the likelihood of incurring in problems here is equivalent to that of incurring in a `SegueDatabaseException`, at which point I'm not sure that a transaction would be of much help anyway.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: I think this goes back [to here](https://github.com/isaacphysics/isaac-api/pull/251#discussion_r400056774). I may have been rushed with that attribute but it did seem too much of an overkill to handle this differently.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Should be fixed in 5840b0a7.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Does a7f0a5ff answer this question? I think this is what I was thinking of when I asked it, but it's been a long time…

Morpheu5: Or maybe this is better, actually c6f25866.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
mlt47: Would it not be just a matter of overloading updateBookingStatus and having the signature with less parameters call the one with more parameters?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
mlt47: It doesn't answer the question as it doesn't justify why those changes were made.
I guess the question is:
if a user is on the waiting list, should it be possible to create a reservation for that user.
It should not be possible to create a new reserved booking because they both count towards the capacity of the event.
You probably also don't want to be updating from Waiting List to Reserved because reserved implies that the student needs to go and fill in their details where as if they are on the waiting list they will have already provided that.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
mlt47: We do have access to transactions from Java, we use them for user Badge management, but I don't think it's worth doing now.

mlt47: Any of EventIsFullException | SegueDatabaseException | EventGroupReservationLimitException | NullPointerException would cause the deletion of previously cancelled users not just SegueDatabaseExceptions.
I feel we should be catching something higher up Exception's inheritance hierarchy than NullPointerException. Currently we wouldn't roll back if any of the other unchecked runtime exceptions were to be thrown in the try block.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Indeed. Should be fixed in f80935d4.

src/test/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManagerTest.java
```diff
@@ -737,18 +814,206 @@ public void isUserAbleToManageEvent_checkUsersWithDifferentRoles_success() throw
...
+
+        dummyEventBookingPersistenceManager.releaseDistributedLock(testCase.event.getId());
+        expectLastCall().atLeastOnce();
+
+        expect(dummyTransactionManager.getTransaction()).andReturn(mock(ITransaction.class)).once();
+
+        expect(dummyUserAccountManager.getUserDTOById(testCase.student1.getId(), anyBoolean())).andReturn(testCase.student1).atLeastOnce();
+        expect(dummyUserAccountManager.getUserDTOById(testCase.student1.getId())).andReturn(testCase.student1).atLeastOnce();
+        expect(dummyUserAccountManager.getUserDTOById(testCase.student2.getId(), anyBoolean())).andReturn(testCase.student2).atLeastOnce();
+        expect(dummyUserAccountManager.getUserDTOById(testCase.student2.getId())).andReturn(testCase.student2).atLeastOnce();
```
Morpheu5, line 903: @mlt47 This is the test I was working on.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: Ok, I gave this method a full overhaul. This now uses a database transaction so we can rollback safely, and also avoids sending emails before completing all the reservations. Tests are broken and I can't figure out how to fix them though.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: This should be resolved in #264 by adding a Quartz job.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
Morpheu5: I'm going assume it's NOT OK to go from waiting list to reservation because that would allow a teacher (and/or group manager?) to make a student "jump the queue".

It will not be possible to create both a reservation and a waiting list booking precisely because of double counting.

I'm going to review the flow of that method with these assumptions.

mlt47: Reservation = take up one space but student needs to add more information
Waiting List = take up one space and student has provided all necessary information

So we wouldn't want to change a booking from waiting list to a reservation.
This would actually only be a possible if there are say 8 bookings for a 5 space event, leaving 3 on the waiting list, then the 5 confirmed cancelled. The teacher could then try to reserve one of the 3 on the waiting list into the 2 available spaces. As it is possible we should cover the case.

Morpheu5: OK, now. db5daf08 should clarify this a bit more. Hopefully that's the last of it.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -899,12 +1186,24 @@ private void ensureValidBooking(final IsaacEventPageDTO event, final RegisteredU
...
                     + " booked on to it.", event.getId(), user.getEmail()));
         }
+        // af599 TODO: Should we consider moving this up to ensureCapacity?
+        if (requestedBookingStatus.equals(BookingStatus.RESERVED) && this.isUserInWaitingList(event.getId(), user.getId())) {
+            Long availablePlacesConfirmed = this.getPlacesAvailable(event, true);
+            if (availablePlacesConfirmed == null || availablePlacesConfirmed < 1L) {
+                throw new EventIsFullException(String.format("Unable to reserve user (%s) onto event (%s) as "
+                        + "there are no available spaces.", user.getId(), event.getId()));
+            }
+        }
```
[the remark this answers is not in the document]
mlt47: I also think if we keep this it should be in ensure capacity

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventBookingManager.java
```diff
@@ -394,6 +448,114 @@ public EventBookingDTO requestBooking(final IsaacEventPageDTO event, final Regis
...
+            try {
+                UserSummaryDTO userSummary = reservation.getUserBooked();
+                Long userId = userSummary.getId();
+                RegisteredUserDTO user = userAccountManager.getUserDTOById(userId);
+                emailManager.sendTemplatedEmailToUser(user,
+                        emailManager.getEmailTemplateDTO("email-event-reservation-requested"),
+                        new ImmutableMap.Builder<String, Object>()
+                                .put("reservingUser", getTeacherNameFromUser(reservingUser))
+                                .put("contactUsURL", generateEventContactUsURL(event))
+                                .put("eventURL", String.format("https://%s/eventbooking/%s",
```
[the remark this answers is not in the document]
mlt47: I see we have a new route for this, presumably for Analytics reasons

PR #256, opened 2020-01-17
Removed log.path property from pom file 

sacummins:
Tidied up logging config so that it defined in one place and follows the rules of using environment variables, then sys properties then default variables.

### DIFF
--- a/src/main/resources/log4j2.yaml
+++ b/src/main/resources/log4j2.yaml
@@ -2,7 +2,7 @@ Configuration:
   properties:
     property:
         - name: LOG_PATH_VAR
-          value: ${env:SEGUE_LOGGING_PATH:-${sys:logging.path:-/isaac-logs}}
+          value: ${env:SEGUE_LOGGING_PATH:-${sys:log.path:-/isaac-logs}}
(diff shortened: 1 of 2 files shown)


src/main/resources/log4j2.yaml
```diff
@@ -1,7 +1,12 @@
 Configuration:
+  properties:
+    property:
+        - name: LOG_PATH_VAR
+          value: ${env:SEGUE_LOGGING_PATH:-${sys:logging.path:-/isaac-logs}}
```
jsharkey13: If I understand this correctly, it moves from using `log.path` as the system property to using `logging.path`. Was this intentional? I don't think we set `logging.path`, but we do still set `log.path` in [compose](https://github.com/isaacphysics/isaac-cs-app/blob/master/compose#L86), [compose-live](https://github.com/isaacphysics/isaac-cs-app/blob/master/compose-live#L55) and [compose-etl].(https://github.com/isaacphysics/isaac-cs-app/blob/master/compose-etl#L38)

jsharkey13: Spoken, and agreed no reason for change. Will stick to `log.path`.

PR #259, opened 2020-02-13
Fix date of birth timezone issue on servers running outside of UTC

robinmessage:
https://trello.com/c/H0hx6dYp/617-dob-off-by-one-gmt-issue

**Pull Request Check List**
- [No] Unit Tests & Regression Tests Added (Optional) (no tests for any PG code at the moment)
- [✅] Removed Unnecessary Logs/System.Outs/Comments/TODOs
- [N/A ] Added enough Logging to monitor expected behaviour change
- [N/A] Security - Injection - everything run by an interpreter (SQL, OS...) is either validated or escaped
- [N/A] Security - Data Exposure - PII is not stored or sent unencrypted
- [N/A] Security - Access Control - Check authorisation on every new endpoint
- [N/A] Security - New dependency - configured sensibly not relying on defaults
- [N/A] Security - New dependency - Searched for any know vulnerabilities
- [N/A] Security - New dependency - Signed up team to mailing list
- [N/A] Security - New dependency - Added to dependency list
- [N/A] DB schema changes - postgres-rutherford-create-script updated
- [N/A] DB schema changes - upgrade script created matching create script
- [N/A] Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgUsers.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgUsers.java
@@ -762,7 +762,16 @@ public class PgUsers implements IUserDataManager {
         u.setGivenName(results.getString("given_name"));
         u.setEmail(results.getString("email"));
         u.setRole(results.getString("role") != null ? Role.valueOf(results.getString("role")) : null);
-        u.setDateOfBirth(results.getDate("date_of_birth"));
+        java.sql.Date dateOfBirth = results.getDate("date_of_birth");
+        // So, DOB is a date in the database.
+        // Because Java is Java, and this code goes back to 1996, the date is interpreted as a datetime at midnight
+        // in the local timezone. This is obviously insane, but there you go.
+        // It works on the servers because they are sensibly set to a UTC timezone. However, it is broken on local
+        // machines where the timezone is London time.
+        // The front-end expects a timestamp in UTC, so this line uses four deprecated functions to make one. The joy!
+        Date utcDate = new Date(
+            Date.UTC(dateOfBirth.getYear(), dateOfBirth.getMonth(), dateOfBirth.getDate(), 0, 0, 0));
+        u.setDateOfBirth(utcDate);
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgUsers.java
```diff
@@ -762,7 +762,16 @@ private RegisteredUser buildRegisteredUser(final ResultSet results) throws SQLEx
...
+        java.sql.Date dateOfBirth = results.getDate("date_of_birth");
+        // So, DOB is a date in the database.
+        // Because Java is Java, and this code goes back to 1996, the date is interpreted as a datetime at midnight
+        // in the local timezone. This is obviously insane, but there you go.
+        // It works on the servers because they are sensibly set to a UTC timezone. However, it is broken on local
+        // machines where the timezone is London time.
+        // The front-end expects a timestamp in UTC, so this line uses four deprecated functions to make one. The joy!
+        Date utcDate = new Date(
+            Date.UTC(dateOfBirth.getYear(), dateOfBirth.getMonth(), dateOfBirth.getDate(), 0, 0, 0));
+        u.setDateOfBirth(utcDate);
```
sacummins: Got to love the old programming problems - thinking back to the millennium bug. 

We have the joda library in our dependency list already so I think that might have nicer less deprecated methods. I can see this fixes the bug so the use of joda is only a suggestion.

mlt47: Java has since also added LocalDate which is worth having a look at. I think its designers were heavily influenced by the joda library.
https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html

robinmessage: Just to note, I didn't see any particularly neat way to do this, and you have to end up with a Date object in the right timezone so I thought just keeping it simple was safer.

PR #260, opened 2020-02-18
Strip whitespace by default when checking string match questions

jsharkey13:
StringMatchQuestions now have a `preserveWhitespace` option, should checking of leading and trailing whitespace be important.


jsharkey13:
@bh412 - I have changed this to trim only trailing whitespace after the content team agreed with you. Could you test it again when you get chance?

PR #261, opened 2020-03-13
Issue/pom upgrades

sacummins:
Upgraded a few packages in the api pom file including jackson databind

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -249,7 +249,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>${jackson.version}</version>
+            <version>2.9.10.3</version>
         </dependency>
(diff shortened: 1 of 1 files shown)


jsharkey13:
This actually undoes my fix to use the latest version of `jackson-databind`; they release patch updates for `jackson-databind` that are not reflected with mirrored releases for other `jackson-...` libraries and so `jackson-databind` should not use the `${jackson-version}` variable since it gets updated so frequently.

sacummins:
There was a later version apparently including the patch (updated in March - I believe the version you were using was older)...

jsharkey13:
Ah, you're right! I missed that there was no `.9.` in the version number, since otherwise they were the same as before my change! `2.9.10.2` --> `2.9.10.3` vs `2.10.2`. I need to read more carefully. Thanks for chasing this! Will fix now!

PR #265, opened 2020-05-15
Add role check for notifications

bh412:


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/NotificationPicker.java
```diff
@@ -107,23 +107,24 @@ public NotificationPicker(final IContentManager contentManager, @Named(CONTENT_I
...
-                // either the use hasn't responded to the notification before...
-                resultsToReturn.add(c);
-            } else if (record.getStatus().equals(NotificationStatus.POSTPONED)) {
-                // or they have but they postponed it...
-                Calendar postPoneExpiry = Calendar.getInstance();
-                postPoneExpiry.setTime(record.getCreated());
-                postPoneExpiry.add(Calendar.SECOND, Constants.NUMBER_SECONDS_IN_ONE_DAY);
 
-                if (new Date().after(postPoneExpiry.getTime())) {
+            if (c.getTags().iterator().next().equals(user.getRole().toString().toLowerCase())) {
```
jsharkey13, line 111: This is definitely not the right or best way to do this. What happens if it is tagged both `student` and `teacher`? Or has another tag before the role tag? Or is not tagged and should thus be shown to everyone?

I think 
```java
            if (c.getTags() != null && !c.getTags().contains(user.getRole().name().toLowerCase())) {
                // Skip irrelevant notifications
                continue;
            }
```
after the skip expired ones will do these things?

mlt47:
> Ian and I are confident this does the right thing, I've got a truth table and I have tested it with an online Java compiler quickly too.

I was looking at stale code 👍

PR #267, opened 2020-06-02
User IDs in staff CSV exports

jsharkey13:
One task I have to do often is produce lists of user IDs for competitions and other things where those who have completed a gameboard need to be emailed. This removes the need for database lookups, since all the required data is in the CSV file.

When you test this, please make sure there are at least two users in the group, one of which has revoked access to test the NOT_SHARING branch.

This also removes unnecessary database lookups by re-using the user object and not extracting it from the request twice in the same function! This needed `isUserAdminOrEventManager` to be consistent with `isUserAdmin` and `isUserStaff` and have two versions, one taking a request and one taking a user DTO; I don't know why it didn't already have this.

PR #268, opened 2020-06-02
Avoid duplicate user checks

jsharkey13:
Save reviewing this until #267 is merged, since this logically follows and is branched from that; the diff contains irrelevant changes until that PR is merged. This is also not urgent, since it's been wrong for ages.

Since 53c68dbc83f88db3129bf6efb81ed3f7b4b85199, extracting a user from a cookie has required a database lookup to ensure the session is still valid. Many places extract the user from the session cookie multiple times, requiring the lookup to be done multiple times too.
We have two facade-layer styles of function for checking roles; one taking the request, for cases where we don't need to know precisely _who_ we've allowed; and one taking the user DTO, more useful when we later need to know _who_ too.
Anywhere we subsequently want the user object, we should always get the user DTO and _then_ check the role using the user object; this is always less costly since we load the full user from the database anyway whichever style you use. Doing it this way round avoids doing it twice.

I also removed the wrapper functions like `isUserAnAdminOrEventManager(RegisteredUserDTO)` from the `AdminFacade` that didn't take a `userManager`, since they made auditing role checks (by finding all usages of the Abstract Facade method) more tricky - and only saved 13 characters of code each time.

I'm confident this changes no behavior, but it needs careful review.


sacummins:
> I guess this is just a style suggestion more than anything, if the user manager is necessary for all of these role checks now it would make sense for them to be methods on the user manager. You'd then have `userManager.isUserAnAdmin(currentlyLoggedInUser)` or even `isAdmin(...)` as the object and parameter would make clear it would be relating to the user, then (as you'll know from python's explicit `self`s!) you'll have `this` to reference the user manager instance.

I'm not too worried about which of your suggestions get implemented now but just a note that I think really some of this authorisation checking stuff belongs in its own class and should be pulled from the user account manager (UAM). Already the UAM is quite big. It is on my list of things to refactor at some point.

mlt47:
> Already the UAM is quite big. It is on my list of things to refactor at some point.

James mentioned this at our daily stand-up. That makes sense, I'm not fussed which way we go for now either then.

PR #269, opened 2020-06-08
Avoid unnecessary database lookups when loading current assignments

jsharkey13:
The facade looks up the owner of each assignment but did not cache the values; if the same teacher sets most of your assignments (the common case) this is a lot of duplicated work.
The AssignmentManager also loaded owner and manager information for every group, only to immediately discard this information and extract just the group IDs. The GroupManager now supports skipping this augmentation, but leaves default methods with the old signatures that do augment.

May be easier to view changes ignoring whitespace: https://github.com/isaacphysics/isaac-api/pull/269/files?w=1


bh412:
Works as before loading assignments, code seems to do as it says

PR #271, opened 2020-06-08
2FA API work (TOTP)

sacummins:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgTOTPDataManager.java
```diff
@@ -0,0 +1,237 @@
...
+
+        return lc;
+    }
+
+    private TOTPSharedSecret createCredentials(final TOTPSharedSecret credsToSave) throws SegueDatabaseException {
+        PreparedStatement pst;
+        try (Connection conn = database.getDatabaseConnection()) {
+            pst = conn
+                    .prepareStatement(
+                            "INSERT INTO user_totp(user_id, shared_secret, created) "
```
jsharkey13, line 88: I think a `last_updated` might be useful here, too!

src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgTOTPDataManager.java
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Helper that picks the correct pst method based on the value provided.
+     *
+     * @param pst - prepared statement - already initialised
+     * @param index - index of the value to be replaced in the pst
+     * @param value - value
+     * @throws SQLException
+     */
+    private void setValueHelper(final PreparedStatement pst, final int index, final Object value) throws SQLException {
```
jsharkey13, line 211: Don't we have at least two of these helper functions in other files already? Can we not add a static method to a class in the `utils` package?

jsharkey13:
More general comments:
 - changing the 2FA device seems buggy (I couldn't make it work from the frontend), and does not ask for password confirmation or other proof like most sites do. Is this intentional?
 - the temporary half-authorised cookie is valid for 7 days, which seems wildly unnecessary for a temporary login flow cookie. Since I changed cookies to store the validity date in them, it's fairly straightforward to limit these to say 5 minutes instead. That seems sensible?
 - our cookies are actually in violation of the [RFC for cookies](https://tools.ietf.org/html/rfc6265#section-4.1.1) since they're not supposed to contain double quotes, backslashes, whitespace or commas. They work, so that seems irrelevant, but I don't like the addition of a whole array inside them. There's also an odd mix of naming conventions for the keys; the core ones are camelCase but two of them are BLOCK_CAPITALS and the new key name is both block caps and half redundant (we don't need `AUTH_COOKIE_` inside a `SEGUE_AUTH_COOKIE`?). Would a boolean flag as simple as `partialLogin` or `partial2FALogin` not be enough, rather than a whole array?

jsharkey13:
Another thing I have thought of is that [`UserAuthenticationManager::isValidUsersSession`](https://github.com/isaacphysics/isaac-api/pull/271/files#diff-2ba8f6ab18c33abf845ec91b38571607R894) is a public method, but does not do the actual validation of whether partial logins are valid. I guess the easiest fix is to make it `private`, but we could push the validation from [`UserAuthenticationManager::getUserFromSessionInformationMap`](https://github.com/isaacphysics/isaac-api/pull/271/files#diff-2ba8f6ab18c33abf845ec91b38571607R388-R396) down into `isValidUsersSession`?

sacummins:
> Another thing I have thought of is that [`UserAuthenticationManager::isValidUsersSession`](https://github.com/isaacphysics/isaac-api/pull/271/files#diff-2ba8f6ab18c33abf845ec91b38571607R894) is a public method, but does not do the actual validation of whether partial logins are valid. I guess the easiest fix is to make it `private`, but we could push the validation from [`UserAuthenticationManager::getUserFromSessionInformationMap`](https://github.com/isaacphysics/isaac-api/pull/271/files#diff-2ba8f6ab18c33abf845ec91b38571607R388-R396) down into `isValidUsersSession`?

I think if it is not currently used externally it should be made private and a comment added. Sessions shouldn't really be a concern for anything other than the authentication manager...

PR #278, opened 2020-08-14
Allow staff to view the teacher connections of another user

Scratchcat1:
---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -73,15 +73,37 @@ import uk.ac.cam.cl.dtg.util.locations.PostCodeRadius;
 
 import javax.annotation.Nullable;
 import javax.servlet.http.HttpServletRequest;
-import javax.ws.rs.*;
-import javax.ws.rs.core.*;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.ForbiddenException;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.EntityTag;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Request;
+import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
-import java.io.*;
-import java.util.*;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import static uk.ac.cam.cl.dtg.isaac.api.Constants.IsaacUserPreferences;
+import static uk.ac.cam.cl.dtg.isaac.api.Constants.*;
 import static uk.ac.cam.cl.dtg.segue.api.Constants.*;
 
 /**
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -26,7 +26,6 @@ import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import com.opencsv.CSVWriter;
 import io.swagger.annotations.Api;
-import io.swagger.annotations.ApiOperation;
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
@@ -54,7 +53,6 @@ import uk.ac.cam.cl.dtg.segue.dao.content.IContentManager;
 import uk.ac.cam.cl.dtg.segue.dao.schools.SchoolListReader;
 import uk.ac.cam.cl.dtg.segue.dao.schools.UnableToIndexSchoolsException;
 import uk.ac.cam.cl.dtg.segue.dos.AbstractUserPreferenceManager;
-import uk.ac.cam.cl.dtg.segue.dos.UserAssociation;
 import uk.ac.cam.cl.dtg.segue.dos.UserPreference;
 import uk.ac.cam.cl.dtg.segue.dos.content.Content;
 import uk.ac.cam.cl.dtg.segue.dos.users.EmailVerificationStatus;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
```diff
@@ -72,37 +73,15 @@
...
-import java.io.IOException;
-import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
+import java.io.*;
+import java.util.*;
```
Morpheu5, line 80: Please revert to individual imports :) sorry…

src/main/java/uk/ac/cam/cl/dtg/segue/api/AuthorisationFacade.java
```diff
@@ -104,20 +104,31 @@ public AuthorisationFacade(final PropertiesLoader properties, final UserAccountM
...
      *            - so we can identify the current user.
+     * @param userId
+     *            - the user id we want the associations of.
+     * @throws NoUserException
+     *            - when the target user cannot be found.
      * @return List of user associations.
      */
     @GET
-    @Path("/")
+    @Path("/{userId}")
```
Morpheu5, line 118: Hang on, are we OK with losing an endpoint like this? @jsharkey13 @mlt47 @sacummins

src/main/java/uk/ac/cam/cl/dtg/segue/api/AuthorisationFacade.java
```diff
@@ -289,20 +300,31 @@ public Response releaseAllAssociations(@Context final HttpServletRequest request
...
      *            - so we can identify the current user.
+     * @param userId
+     *            - the id of the user we want the access rights of.
+     * @throws NoUserException
+     *            - when the target user cannot be found.
      * @return List of user associations.
      */
     @GET
-    @Path("/other_users")
+    @Path("/other_users/{userId}")
```
Morpheu5, line 314: Same issue with losing an endpoint here.

src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
```diff
@@ -142,16 +142,27 @@ public Response getGroupsForCurrentUser(@Context final HttpServletRequest reques
...
      * @param request            - so we can identify the current user.
      * @param cacheRequest       - so that we can control caching of this endpoint
+     * @param userId             - the user we want the groups of.
      * @param archivedGroupsOnly - include archived groups in response - default is false - i.e. show only unarchived
+     * @throws NoUserException   - when the user cannot be found.
      * @return List of groups for the current user.
      */
     @GET
-    @Path("/membership")
+    @Path("/membership/{userId}")
```
Morpheu5, line 151: Same here…

src/main/java/uk/ac/cam/cl/dtg/segue/api/AuthorisationFacade.java
Scratchcat1: They could be combined with `"/{userId: .*}"` and then have the id fallback to that of the current user.

Morpheu5: Yes, if that's an option, it also allows us to keep the code compact. Go for it.

Scratchcat1: It appears @Path only supports paths which still have at least one character excluding the trailing slash e.g. `authorisations/foo/{userId: .*
}` matches with `authorisations/foo` and `/foo/123` but `{userId: .*}` doesn't match `authorisations/` since the part in @Path only contains a slash.
Instead should I separate the endpoints?

jsharkey13: I think these should be separate endpoints. But it is likely that you can make the shortcut one call the Java function for the specified user one, passing in the Request object and setting the current user's ID as the target.

Your existing permission check should cover that use case just fine.

src/main/java/uk/ac/cam/cl/dtg/segue/api/AuthorisationFacade.java
```diff
@@ -289,20 +323,54 @@ public Response releaseAllAssociations(@Context final HttpServletRequest request
...
+     * @throws NoUserException
+     *            - when the target user cannot be found.
+     * @return List of user associations.
+     */
+    @GET
+    @Path("/other_users/{userId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @GZIP
+    @ApiOperation(value = "List all users the specified user has been granted access by.")
+    public Response getCurrentAccessRightsSpecificUser(@Context final HttpServletRequest request, @PathParam("userId") Long userId) throws NoUserException{
```
Morpheu5, line 364: Stylistic pet-peeve here, feel free to ignore me, I'm just being a grumpy old man, but I'd rather rename this to something like `getCurrentAccessRightsForSpecificUser` or even `getCurrentAccessRightsForUser`, but maybe Apple led me astray with their Smalltalk-inspired messages they used in their Objective-C Cocoa API.

Morpheu5 (approved):
Stylistic pet-peeves aside, this looks OK to me now.

src/main/java/uk/ac/cam/cl/dtg/segue/api/AuthorisationFacade.java
```diff
@@ -104,20 +104,54 @@ public AuthorisationFacade(final PropertiesLoader properties, final UserAccountM
...
+    @GET
+    @Path("/{userId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @GZIP
+    @ApiOperation(value = "List all users granted access to the specified user's data.")
+    public Response getUsersWithAccessSpecificUser(@Context final HttpServletRequest request, @PathParam("userId") Long userId) throws NoUserException {
+        try {
+            RegisteredUserDTO requestingUser = userManager.getCurrentRegisteredUser(request);
+
+            if (!isUserStaff(userManager, request) && !userId.equals(requestingUser.getId())) {
```
jsharkey13, line 149: It's a bit late now, but this was missed by whoever reviewed this PR. On the most common path (viewing users who have access to the current user's data), we load the user object from the database 3 times. Once to get the ID in the first method, once to get `requestingUser` and then again to do the `isUserStaff` check. This is a bit much!

We could avoid loading the user more than once by passing `null` rather than a user ID, but that seems a bit hideous. I think we live with loading the user twice, and use `isUserStaff(userManager, requestingUser)` instead!

jsharkey13: Oh, it's actually four times, because we then do `getUserDTOById(userId)` below, too!

jsharkey13: See https://github.com/isaacphysics/isaac-api/commit/b0082b47213e2a77f147c2cb3d856bde136c7648 for fix.

PR #280, opened 2020-08-19
Add an endpoint to allow updating of email statuses without a user's cookie

jsharkey13:
This endpoint expects an `Authorization: Bearer some-secret-token` header to be present in the request to provide the authentication.
It gracefully fails if this token has not been configured in the server properties.
It may be vulnerable to timing attacks to guess the token, so it prints a log message for all successful and unsuccessful requests.
The header usage is declared in such a way that Swagger will ask for it, allowing easier use by hand if necessary.

To test it, you'll need to copy the config line from the template into your local config file, although do test the case when you have not, too. Feel free to chose something different as your local token. To test with Swagger, use: http://localhost:8080/isaac-api/api-docs/#/admin/setUsersEmailVerificationStatusFailed


Scratchcat1 (approved):

Morpheu5 (approved):

sacummins:
Hmmmmn.. Not sure I like that it is stored in the config file. Can this be easily revoked without logging into servers? Perhaps a dedicated API token service is a better idea?

jsharkey13:
No, it can't easily be revoked without logging in to the servers, but on the other hand;
 - the only way abuse of this can be detected is by logging in to the servers,
 - this is much less likely to be broken/accidentally insecure than a complicated token service for (hopefully) a single use-case,
 - this method cannot be used to confirm the existence of accounts even if the token is known,
 - the worst abuse can do is annoy users by making a banner appear on the targeted account(s) and stop us from emailing them,
 - the token will be longer than the HMAC in cookies, meaning it would be quicker and easier to brute force those than this, for far greater harm.

PR #283, opened 2020-09-09
Feature/my progress++ endpoints

Scratchcat1:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


sacummins:
It does seem that some of these new API endpoints would make more sense to live elsewhere. The GameboardFacade was nicely cohesive around operations dealing with gameboards before and some of these methods are really closer to user progress calculations. (Unless I have misunderstood?)

Perhaps the utility functions you need could be refactored to be in a shared parent class to prevent duplication?

I can't tell which segue imports you are worried about (mentioned in your commit comment) but given we have discussed that we are likely to dissolve the boundary between segue and isaac it could be a good opportunity to start that work if it makes this  nicer and isn't too hard... That being said, if this is an urgent feature we could decide to accept some tech debt and just add a ticket to fix it later.

src/main/java/uk/ac/cam/cl/dtg/segue/quiz/PgQuestionAttempts.java
```diff
@@ -227,14 +227,61 @@ public void registerQuestionAttempt(final Long userId, final String questionPage
...
+        }
+    }
+
+    @Override
+    public List<String> getUnsolvedQuestions(final Long userId)
+            throws SegueDatabaseException {
+        PreparedStatement pst;
+        try (Connection conn = database.getDatabaseConnection()) {
+            // Perform nested query to extract the page ids with at least one question which has been attempted but never solved
+            pst = conn.prepareStatement("SELECT SUBSTRING(question_id, 1, STRPOS(question_id, '|') - 1) AS question_page_id " +
```
jsharkey13, line 268: Unfortunately, this won't actually find all questions that are not solved. It is perfectly possible to have got parts A, B and C correct, and just not have attempted Part D. That question will show as incomplete on the My Progress page, but this will not find it, since there is no attempt in the database for Part D. The database does not know how many question parts a question has, and so you have to use `GameManager::getAllMarkableQuestionPartsDFSOrder` in order to know what parts there are to a question. I can't see an easy way of fixing this without rewriting it entirely, and since we need all the attempts to do the augmenting we don't need this additional database query :/

jsharkey13 (changes requested):
Not for Connor to fix, and entirely my fault for not spotting this earlier nor thinking and investigating the method used more carefully. I had intended this information be geenrated by the `StatisticsManager::getUserQuestionInformation` method, since that already loads all questions and thus knows which ones are incorrect and which ones were most recent.
This PR attempts to use the database to judge most recent, which will (making some iffy assumptions about question ID formats) work; but it cannot be used to work out which questions are incomplete (rather than say incorrect). Since there is no way to distinguish these states; they still count as "Not Done" from the user's perspective, this does not actually do what it needs to :(

We could move more logic down to the database layer, as this tries to do; but if we were to go down that path, then a more useful first step would be to add a `page_id` column to the question_attempts table, which would improve both this and many other queries.

PR #286, opened 2020-09-18
Add a filter to respond to any invalid sessions with 412 Precondition failed

Scratchcat1:
- Requires an extra database lookup for each request from a logged in user. User must be requested in order to fetch the current session token.
- If the segue auth cookie is present but a NoUserLoggedInException is thrown when requesting the user the request is aborted and a 412 is sent as a response.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13:
This is a nice proof of concept, and it will be good to have to refer to. But for now we are not willing to pay the performance penalty of an extra database lookup on every request, and we'll see if the session expiry information in the header is enough.
(Thanks for the hard work though!)

PR #287, opened 2020-09-21
Allow event managers to send up to 100 emails per day

Scratchcat1:
- Add SendMailMisuseHandler
- Only allow event managers to send event emails
- Add new misuse handler method `willHaveMisused` to check if adding a number of events would cause a misuse exception.
  This is used to detect if the user would exceed the hard limit when sending the emails without adding them to the misuse monitor.
  This prevents a large batch of emails being counted by the misuse monitor without actually being sent.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
@@ -32,6 +32,7 @@ import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
 import uk.ac.cam.cl.dtg.segue.api.monitors.EmailVerificationMisuseHandler;
 import uk.ac.cam.cl.dtg.segue.api.monitors.EmailVerificationRequestMisuseHandler;
 import uk.ac.cam.cl.dtg.segue.api.monitors.IMisuseMonitor;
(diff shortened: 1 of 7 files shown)


bh412 (approved):
Works as expected, fixed an off by one error in the willHaveMisused monitor. Easy to alter in future if we expand to staff or allow different email types sending

src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
```diff
@@ -488,10 +489,25 @@ public final Response sendEmailsToUserIds(@Context final HttpServletRequest requ
 
         try {
             RegisteredUserDTO sender = this.userManager.getCurrentRegisteredUser(request);
-            if (!isUserAnAdmin(userManager, sender)) {
+            if (!isUserAnAdminOrEventManager(userManager, sender)) {
                 return SegueErrorResponse.getIncorrectRoleResponse();
             }
 
+            if (isUserAnEventManager(userManager, sender)) {
+                if (!emailType.equals(EmailType.EVENTS)) {
```
jsharkey13, line 497: Where did this restriction come from? I think it's ok for now, but I foresee it causing issues.

jsharkey13 (approved):

PR #291, opened 2020-10-01
Feature/report log events to Prometheus

mlt47:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


Morpheu5 (approved):

mlt47:
This does not do what it should do. EVENT_TYPES included in constants only include the event types produced by the server (for data integrity reasons). The CLIENT_SIDE_ERROR event type, by definition, is not generated by the API but reported from the client. We have discussed solutions and they will be implemented in a separate pull request.

PR #293, opened 2020-10-06
Add session expiry information as current user request header

jsharkey13:
Since the client can't read the cookie, it does not know how long until the session expires. Add the information in a header, and ensure it can be accessed by browsers using the CORS exposedHeaders parameter.
The date is provided in HTTP-Date format.


bh412 (approved):
works with/without remember me, matches cookie and same on reload with same login. Endpoint functions as before otherwise

PR #297, opened 2020-10-15
Skip over deleted users for events booking CSV.

Morpheu5:
Deleted users cannot be retrieved by the UserAccountManager if they have been deleted, yet we were doing just that and re-throwing the exception while building the CSV. This commit catches the exception and advances the loop to skip over the student that was deleted.

We could possibly want to add a line that says DELETED or something like that.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
@@ -615,7 +615,13 @@ public class EventsFacade extends AbstractIsaacFacade {
             for (EventBookingDTO booking : eventBookings) {
                 ArrayList<String> resultRow = Lists.newArrayList();
                 UserSummaryDTO resultUser = booking.getUserBooked();
-                RegisteredUserDTO resultRegisteredUser = this.userAccountManager.getUserDTOById(resultUser.getId());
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -615,7 +615,13 @@ public Response getEventBookingCSV(@Context final HttpServletRequest request,
             for (EventBookingDTO booking : eventBookings) {
                 ArrayList<String> resultRow = Lists.newArrayList();
                 UserSummaryDTO resultUser = booking.getUserBooked();
-                RegisteredUserDTO resultRegisteredUser = this.userAccountManager.getUserDTOById(resultUser.getId());
+                RegisteredUserDTO resultRegisteredUser;
+                try {
+                    resultRegisteredUser = this.userAccountManager.getUserDTOById(resultUser.getId());
+                } catch (NoUserException e) {
+                    log.error(String.format("User with ID \"%d\" could not be retrieved. Continuing...", resultUser.getId()));
```
jsharkey13, line 622: This isn't helpful, since we can commonly expect this case and it is not an error. Either replace it with a `log.debug(...)` or with a comment to say "// Deleted user, skipping`?

PR #301, opened 2020-11-13
Monitor question validation times by validator type

mlt47:
**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (approved):
I guess we don't mind not being able to distunguish validator errors cases from normal cases, since they all print server log messages and the the vast minority. This will be much better than nothing!

PR #302, opened 2020-11-19
Override question attempt misuse threshold 

jsharkey13:
This property is only read at startup, and so can't be changed on the fly. It wouldn't be too difficult to change to reading it on the fly, but it felt more efficient to parse the string only once, not every single time a question is attempted on the platform.
It fails gracefully if the server property is not provided, to the normal default value, so it can safely be ignored on the live platform and just adjusted on staging. It only applies to logged in users, since anonymous users have several rate limits.


Morpheu5 (approved):
So far as the override mechanism is "set a higher threshold", I think we are fine here. It could be possibly desirable to disable the threshold entirely, but I suspect setting it to 9999 is good enough.

mlt47 (approved):
This looks sensible

PR #307, opened 2021-01-22
Require admin users authenticating with a password to have 2FA enabled

jsharkey13:
This will block login via email and password for admin accounts that do not have 2FA configured on their accounts. It does not automatically demote the account, just prevents login with a warning message.


sacummins (approved):
Looks ok to me. Do we have many admin accounts without MFA?

jsharkey13:
@sacummins We're trying for all staff accounts, not just admins, to use 2FA and we're mostly there. But we're going to hard-enforce the admin ones; everyone has been forewarned and we set a deadline of January. Matthew was chasing up anyone who hadn't configured it last month, so I think we're all good to deploy this!

jsharkey13:
My next concern is what to do about the test accounts on staging and how to do the automated tests on test!

PR #309, opened 2021-01-26
Create quiz facade and add /available and preview endpoints

robinmessage:
This also extracts out ContentSummarizerService and ContentService.

---

**Pull Request Check List**
- [ ] It seems like the deprecated endpoints on SegueContentFacade (`search` and `getContentById`) should be removed since there is no sensible way for them to not return quizzes that are  not `visibleToStudents` (and they already expose full question page details.)
- [ :+1: ] Removed Unnecessary Logs/System.Outs/Comments/TODOs
- [ None ] Added enough Logging to monitor expected behaviour change
- [ :+1: ] Security - Injection - everything run by an interpreter (SQL, OS...) is either validated or escaped
- [ :+1: ] Security - Data Exposure - PII is not stored or sent unencrypted
- [ :+1: ] Security - Access Control - Check authorisation on every new endpoint
- [ N/A ] Security - New dependency - configured sensibly not relying on defaults
- [ N/A ] Security - New dependency - Searched for any know vulnerabilities
- [ N/A ] Security - New dependency - Signed up team to mailing list
- [ N/A ] Security - New dependency - Added to dependency list
- [ N/A ] DB schema changes - postgres-rutherford-create-script updated
- [ N/A ] DB schema changes - upgrade script created matching create script
- [ N/A ] Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.managers.QuizManager;
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacQuizDTO;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
+import uk.ac.cam.cl.dtg.segue.auth.exceptions.NoUserLoggedInException;
 import uk.ac.cam.cl.dtg.segue.dao.ILogManager;
 import uk.ac.cam.cl.dtg.segue.dao.SegueDatabaseException;
 import uk.ac.cam.cl.dtg.segue.dao.content.ContentManagerException;
@@ -102,13 +103,9 @@ public class QuizFacade extends AbstractIsaacFacade {
     @ApiOperation(value = "Get quizzes visible to this user.")
     public final Response getAvailableQuizzes(@Context final HttpServletRequest request) {
         try {
-            AbstractSegueUserDTO user = this.userManager.getCurrentUser(request);
+            RegisteredUserDTO user = this.userManager.getCurrentRegisteredUser(request);
 
-            if (!isUserRegistered(user)) {
-                throw new ForbiddenException("Only registered users can see quizzes.");
-            }
-
-            boolean isStudent = !isUserTeacherOrAbove(user);
+            boolean isStudent = !isUserTeacherOrAbove(userManager, user);
 
             EntityTag etag = new EntityTag(this.contentManager.getCurrentContentSHA().hashCode() + "");
 
@@ -117,14 +114,12 @@ public class QuizFacade extends AbstractIsaacFacade {
             return Response.ok(summary).tag(etag)
                 .cacheControl(getCacheControl(NUMBER_SECONDS_IN_ONE_HOUR, isStudent))
                 .build();
-        } catch (SegueDatabaseException e) {
-            String message = "SegueDatabaseException whilst getting available quizzes";
-            log.error(message, e);
-            return new SegueErrorResponse(Response.Status.INTERNAL_SERVER_ERROR, message).toResponse();
         } catch (ContentManagerException e) {
             String message = "ContentManagerException whilst getting available quizzes";
             log.error(message, e);
             return new SegueErrorResponse(Response.Status.INTERNAL_SERVER_ERROR, message).toResponse();
+        } catch (NoUserLoggedInException e) {
+            return SegueErrorResponse.getNotLoggedInResponse();
         }
     }
 
@@ -147,18 +142,16 @@ public class QuizFacade extends AbstractIsaacFacade {
     public final Response previewQuiz(@Context final Request request, @Context final HttpServletRequest httpServletRequest,
                                       @PathParam("quizId") final String quizId) {
         try {
-            AbstractSegueUserDTO user = this.userManager.getCurrentUser(httpServletRequest);
+            RegisteredUserDTO user = this.userManager.getCurrentRegisteredUser(httpServletRequest);
 
-            if (!(isUserTeacherOrAbove(user))) {
-                throw new ForbiddenException("Students cannot preview quizzes.");
+            if (!(isUserTeacherOrAbove(userManager, user))) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
             }
 
             if (null == quizId || quizId.isEmpty()) {
                 return new SegueErrorResponse(Response.Status.BAD_REQUEST, "You must provide a valid quiz id.").toResponse();
             }
 
-            // Calculate the ETag on current live version of the content
-            // NOTE: Assumes that the latest version of the content is being used.
             EntityTag etag = new EntityTag(this.contentManager.getCurrentContentSHA().hashCode() + quizId.hashCode() + "");
 
             Response cachedResponse = generateCachedResponse(request, etag);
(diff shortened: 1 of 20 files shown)


robinmessage:
Also, this PR puts in place `IsaacQuiz` and `IsaacQuizPage`. I don't know how the content editor handles things, but I'm wondering if I have the right structure here. In particular:

- Should the only top level objects in git be IsaacQuiz objects, which then contain the IsaacQuizPage objects, or should there be an indirection from IsaacQuiz to IsaacQuizPage?
- Either way, is IsaacQuizPage really a page? I thought it was a page like a question page, but now I see it has a load of gubbins around related content and tags which are irrelevant.
- - If anything, IsaacQuiz is more like a page, so perhaps it should inherit from `SeguePage`, and contain page-lite objects for each page of the quiz?
- Does any of this complicate the content editor?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -0,0 +1,192 @@
...
+            log.error(message, e);
+            return new SegueErrorResponse(Response.Status.INTERNAL_SERVER_ERROR, message).toResponse();
+        }
+    }
+
+    private static boolean isUserRegistered(AbstractSegueUserDTO user) {
+        return user instanceof RegisteredUserDTO;
+    }
+
+    private static boolean isUserTeacherOrAbove(AbstractSegueUserDTO user) {
```
jsharkey13, line 189: I added [`AbstractSegueFacade::isUserTeacherOrAbove`](https://github.com/isaacphysics/isaac-api/blob/v2.15.2/src/main/java/uk/ac/cam/cl/dtg/segue/api/AbstractSegueFacade.java#L326-L329) after seeing this and noticing the lack of checks in the Groups and Assignments facades. 
I know the AbstractFacade method is a little clunckier, but it can at least be used consistently that way. It just needs the usages of this method changing to `isUserTeacherOrAbove(userManager, user)`, noting that it needs to be a registered user though.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -0,0 +1,192 @@
...
+    @GET
+    @Path("available")
+    @Produces(MediaType.APPLICATION_JSON)
+    @GZIP
+    @ApiOperation(value = "Get quizzes visible to this user.")
+    public final Response getAvailableQuizzes(@Context final HttpServletRequest request) {
+        try {
+            AbstractSegueUserDTO user = this.userManager.getCurrentUser(request);
+
+            if (!isUserRegistered(user)) {
```
jsharkey13, line 107: I think the usual pattern for this check is `userManager.getCurrentRegisteredUser(request)` and then to catch the exception that throws for anonymous users:
```java
} catch (NoUserLoggedInException e) {
    return SegueErrorResponse.getNotLoggedInResponse();
}
```
in places where we don't want anonymous users at all. That is probably why there is no `isUserRegistered` method already. Could you swap to that style?

jsharkey13: See e.g. the [Assignment Facade](https://github.com/isaacphysics/isaac-api/blob/v2.15.2/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java#L254).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
jsharkey13: If you swap to `getCurrentRegisteredUser` as mentioned above, this should work just fine!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -0,0 +1,192 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @GZIP
+    @ApiOperation(value = "Preview an individual quiz.")
+    public final Response previewQuiz(@Context final Request request, @Context final HttpServletRequest httpServletRequest,
+                                      @PathParam("quizId") final String quizId) {
+        try {
+            AbstractSegueUserDTO user = this.userManager.getCurrentUser(httpServletRequest);
+
+            if (!(isUserTeacherOrAbove(user))) {
+                throw new ForbiddenException("Students cannot preview quizzes.");
```
jsharkey13, line 153: We tend to use `return SegueErrorResponse.getIncorrectRoleResponse();` rather than whatever this  `ForbiddenException` does.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacQuiz.java
```diff
@@ -0,0 +1,84 @@
...
+	public boolean getVisibleToStudents() {
+		return visibleToStudents;
+	}
+
+	public void setVisibleToStudents(boolean visibleToStudents) {
+		this.visibleToStudents = visibleToStudents;
+	}
+
+	@JsonIgnore
+	public List<IsaacQuizSection> getSections() {
```
jsharkey13, line 81: Maybe @mlt47 has opinions on this. As far as I remember, most of our DOs don't have logic in them, and this sort of stuff resides in the Manager layer, c.f. `GameManager::getAllMarkableQuestionPartsDFSOrder`?

jsharkey13 (commented):
There are one or two small changes that could be nice, and then one question for Meurig about whether the logic should live in the DO or the Manager (and whether any of us care enough).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            if (null == quizId || quizId.isEmpty()) {
+                return new SegueErrorResponse(Response.Status.BAD_REQUEST, "You must provide a valid quiz id.").toResponse();
+            }
+
+            // Calculate the ETag on current live version of the content
+            // NOTE: Assumes that the latest version of the content is being used.
```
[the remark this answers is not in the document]
jsharkey13: (I should probably remove the original comment, too!)

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacQuiz.java
robinmessage: I don't even use this code yet, so maybe I should delete it?

I guess I saw this more as the contract of the DO (that it only contains IsaacQuizSection objects) and isn't actually intended to do any logic (the code should be a no-op).

Is there a way to distinguish production from staging/dev? In prod, it could do what it does above (filter out non-IsaacQuizSection); in other environments it could be an assertion that there are no non-IsaacQuizSection and throw a runtime exception if there are.

Also, it should probably cache its response so it appears more referentially transparent.

Another option is to override and narrow `getChildren` and override `setChildren` with a guard, but `setChildren` is final so I can't without changing `Content`.

The final alternative is to make this not inherit from SeguePage, which maybe makes sense as it isn't exactly a page and is never presented as one?

mlt47 (commented):
This all looks good and sensible to me.
I am also quite confident it won't break any existing code, but did we mention trying to hide this behind a feature flag (perhaps SEGUE_APP_ENVIRONMENT in the config file so that it is DEV only)? I feel like it wouldn't gain us much (except for some extra lines of code!) at this point.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizManager.java
```diff
@@ -0,0 +1,142 @@
...
+        } else {
+            resultsLimit = DEFAULT_RESULTS_LIMIT;
+        }
+
+        if (null != startIndex) {
+            startIndexOfResults = startIndex;
+        } else {
+            startIndexOfResults = 0;
+        }
+
```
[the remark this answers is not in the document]
robinmessage: Good spot Meurig; that's totally what I intended when I created `ContentService` and I'm not sure how I left that out.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacQuiz.java
robinmessage: Thanks. But yeah, for consistency, I'll remove it and do it in the manager.

robinmessage:
@mlt47 re: feature flag: when I make a functionality change to the existing code I think that might be worth putting behind a flag. Is `DEV` the best flag? Or would `not PROD` (is it called PROD) be better? But for all this new stuff it doesn't seem worth it to me.

mlt47:
> @mlt47 re: feature flag: when I make a functionality change to the existing code I think that might be worth putting behind a flag. Is `DEV` the best flag? Or would `not PROD` (is it called PROD) be better? But for all this new stuff it doesn't seem worth it to me.

Yes @jsharkey13  agrees, there's no need to hide these endpoints behind a feature flag.

jsharkey13 (approved):

PR #312, opened 2021-02-03
Quiz: start quiz attempt

robinmessage:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (approved):
I think the two relevant commits look good, I just wondered whether `quiz_attempts`, which stores the metadata of an overall attempt at a quiz, might get compared to `question_attempts` which is not similar. But I guess it will be more clear what is what when something like `quiz_question_attempts` exists, so I don't think it matters.

PR #313, opened 2021-02-03
Quiz: cancel quiz assignment

robinmessage:
---

**Pull Request Check List**

- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -22,6 +22,7 @@ import io.swagger.annotations.ApiOperation;
 import org.jboss.resteasy.annotations.GZIP;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.ac.cam.cl.dtg.isaac.api.managers.AssignmentCancelledException;
 import uk.ac.cam.cl.dtg.isaac.api.managers.AttemptCompletedException;
 import uk.ac.cam.cl.dtg.isaac.api.managers.DueBeforeNowException;
 import uk.ac.cam.cl.dtg.isaac.api.managers.DuplicateAssignmentException;
@@ -47,6 +48,7 @@ import uk.ac.cam.cl.dtg.segue.dto.users.RegisteredUserDTO;
 import uk.ac.cam.cl.dtg.util.PropertiesLoader;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -278,7 +280,7 @@ public class QuizFacade extends AbstractIsaacFacade {
             }
 
             // Get the quiz assignment
-            QuizAssignmentDTO quizAssignment = this.quizAssignmentManager.getQuizAssignment(quizAssignmentId);
+            QuizAssignmentDTO quizAssignment = this.quizAssignmentManager.getById(quizAssignmentId);
 
             // Check the user is an active member of the relevant group
             UserGroupDTO group = groupManager.getGroupById(quizAssignment.getGroupId());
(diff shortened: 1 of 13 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgQuizAssignmentPersistenceManager.java
```diff
@@ -167,14 +167,17 @@ public Long saveAssignment(final QuizAssignmentDTO assignment) throws SegueDatab
...
-    public QuizAssignmentDTO getAssignmentById(Long quizAssignmentId) throws SegueDatabaseException {
+    public QuizAssignmentDTO getAssignmentById(Long quizAssignmentId) throws SegueDatabaseException, AssignmentCancelledException {
         try (Connection conn = database.getDatabaseConnection()) {
             PreparedStatement pst;
             pst = conn.prepareStatement("SELECT * FROM quiz_assignments WHERE id = ?");
             pst.setLong(1, quizAssignmentId);
 
             ResultSet results = pst.executeQuery();
             if (results.next()) {
+                if (results.getBoolean("deleted")) {
```
jsharkey13, line 178: Could this not just be an `AND NOT deleted` in the query, as we do for users and groups when they are deleted? Admittedly this is less useful information for the client, but perhaps saves the need for an exception?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -305,6 +307,8 @@ public final Response startQuizAttempt(@Context final Request request, @Context
             return new SegueErrorResponse(Status.INTERNAL_SERVER_ERROR, message).toResponse();
         } catch (AttemptCompletedException e) {
             return new SegueErrorResponse(Status.FORBIDDEN, "You have already completed your attempt at this quiz.").toResponse();
+        } catch (AssignmentCancelledException e) {
+            return new SegueErrorResponse(Status.GONE, "This quiz assignment has been cancelled.").toResponse();
```
jsharkey13, line 311: See my other comment, but we've never needed a 410 Gone before because we've done the filtering of deleted things in the database so they have no side effects in the code; it means we can't tell the difference between deleted and non-existent but I think that's fine.

jsharkey13 (commented):
This commit looks ok too. Although I have noticed that some of the paths are `camelCase`, whereas all of our existing endpoints follow the `snake_case` convention. I must have missed that in previous PRs.

robinmessage:
I'll fix paths in another commit as they all need a once over (snake case, slashes, maybe tweaking them to be more RESTful).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
robinmessage: So, I went for the `GONE` because we send out links as soon as teacher sets a task, so I expected to use the status in the front-end to give a nicer message when that occurs instead of a generic site error or SegueDatabaseException message.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgQuizAssignmentPersistenceManager.java
mlt47: When James was explaining this to me I was going to suggest pushing the check and exception down as close to the DB as possible but it looks like this is already done. I personally am happy for the deleted check to be done at the persistence level and throwing a sensible checked exception if it improves the information to the user.

I guess one failure that I could see happening that we would like to avoid would be someone copying this method to create their own SELECT and not copying the deleted check. It is more practical than pretty... but we could get around that by adding a comment to the end of line 173 noting that deleted assignments are handled in code?

mlt47 (approved):

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgQuizAssignmentPersistenceManager.java
jsharkey13: Since this aims to cover exactly the same error case as happens with ordinary assignments, should we think about refactoring that @mlt47, or just start here first?

mlt47: Start here first, it is not Robin's responsibility anyway :)

PR #316, opened 2021-02-04
Quiz: answer question

robinmessage:
Getting there, you can now actually answer questions...

I'm not 100% happy with `ResponseWrapper`; as `SegueResponse` itself says, maybe it should be an exception or a Throwable. But it very much felt like I wanted exceptional control flow and didn't want to make significant changes to the logic of existing code.

---

**Pull Request Check List**
- [ ? ] Added enough Logging to monitor expected behaviour change
- [ ] Peer-Reviewed


jsharkey13 (approved):
The detail of this all looks good. I wonder if @mlt47 has opinions on the ResponseWrapper exception class?

mlt47:
> The detail of this all looks good. I wonder if @mlt47 has opinions on the ResponseWrapper exception class?

I can see the usefulness of the ResponseWrapper for doing a similar job to what the suggested refactor would do. I'm happy with doing that in the Facades. I would slightly prefer not using it in the managers so that the managers don't need to know about responses and requests but, as there are already many manager methods that are returning responses, that is OK with me for now.

mlt47:
> > The detail of this all looks good. I wonder if @mlt47 has opinions on the ResponseWrapper exception class?
> 
> I can see the usefulness of the ResponseWrapper for doing a similar job to what the suggested refactor would do. I'm happy with doing that in the Facades. I would slightly prefer not using it in the managers so that the managers don't need to know about responses and requests but, as there are already many manager methods that are returning responses, that is OK with me for now.

Let's rename it to ErrorResponseWrapper or something to differentiate it from ResultsWrapper

robinmessage:
@mlt47 Renaming to `ErrorResponseWrapper` sounds good to me, I'll do that.

PR #317, opened 2021-02-05
Quiz: augment quiz questions

robinmessage:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizQuestionManager.java
```diff
@@ -81,4 +85,56 @@ public void recordQuestionAttempt(QuizAttemptDTO quizAttempt, QuestionValidation
...
+        Map<String, List<QuestionValidationResponse>> answers = quizQuestionAttemptManager.getAnswers(quizAttempt.getId());
+
+        for (QuestionDTO question : questionsToAugment) {
+            List<QuestionValidationResponse> questionAttempts = answers.get(question.getId());
+
+            // The latest answer is the only answer we consider.
+            if (questionAttempts.size() > 0) {
+                QuestionValidationResponseDTO lastAttempt = questionManager.convertQuestionValidationResponseToDTO(questionAttempts.get(questionAttempts.size() - 1));
+                if (!includeCorrect) {
+                    lastAttempt.setCorrect(null);
```
jsharkey13, line 134: Will this not leave the `explanation` field from the validation response still present, which might explain the correct answer or inform them of the correct/incorrectness. What about `QuantityValidationResponseDTO` s which have `correctValue` and `correctUnits`? (There are other subclasses, but that is the only one that is exposed to the user other than the base class by `QuestionValidationResponseOrikaConverter`).

I wonder whether the safest thing to do is to manually create a new `QuestionValidationResponseDTO`, with only the `questionId`, `answer` and `dateAttempted` fields copied over from the actual DO, using a new private method?

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgQuizQuestionAttemptPersistenceManager.java
```diff
@@ -75,4 +81,35 @@ public void registerQuestionAttempt(Long quizAttemptId, QuestionValidationRespon
             throw new SegueDatabaseException("Unable to process json exception", e);
         }
     }
+
+    @Override
+    public Map<String, List<QuestionValidationResponse>> getAnswers(Long quizAttemptId) throws SegueDatabaseException {
```
jsharkey13, line 86: Could this be clearer, `getAllAnswersForQuizAttempt` or something?

jsharkey13 (changes requested):
I think this will leak the correctness of answers, despite trying not to. Our validation response classes are a bit messy, and just blanking out `correct` isn't going to be enough I don't think!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizQuestionManager.java
robinmessage: Yes, I think I added `lastAttempt.setExplanation(null);` but maybe that was on another PR?

In any case, I hadn't thought about subclasses. I think manually copying out of the DO into a DTO with those fields would be safer, yes. I'll do that.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgQuizQuestionAttemptPersistenceManager.java
robinmessage: Sure, can do.

PR #318, opened 2021-02-05
Quiz: complete quiz attempt

robinmessage:
Tests to follow in separate PR

---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -436,6 +436,72 @@ public final Response getQuizAttempt(@Context final HttpServletRequest httpServl
...
+     * @return Confirmation or an error.
+     */
+    @POST
+    @Path("/attempt/{quizAttemptId}/complete")
+    @Produces(MediaType.APPLICATION_JSON)
+    @GZIP
+    @ApiOperation(value = "Mark a QuizAttempt as complete (or incomplete).")
+    public final Response completeQuizAttempt(@Context final HttpServletRequest httpServletRequest,
+                                              @PathParam("quizAttemptId") final Long quizAttemptId,
+                                              @FormParam("markIncomplete") final Boolean markIncomplete) {
```
jsharkey13, line 461: I notice I mised a few already, but `@FormParam` isn't very JSON RESTful, I don't think? I've pushed the bounds before now by allowing a [top-level list as JSON](https://github.com/isaacphysics/isaac-api/blob/v2.15.2/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java#L532-L535) or [a single string value as JSON](https://github.com/isaacphysics/isaac-api/blob/v2.15.2/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java#L1599-L1601), since we have never had `application/x-www-form-urlencoded` in requests to the API before I don't think. You might also find it tricky to make Axios, our frontend request library to make non-JSON requests although I have never tried.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -436,6 +436,72 @@ public final Response getQuizAttempt(@Context final HttpServletRequest httpServl
...
+            } else if (quizAttempt.getUserId().equals(user.getId()) && requestedMarkComplete == true) {
+                markComplete = true;
+            } else {
+                // Heuristic for most helpful error message.
+                return new SegueErrorResponse(Status.FORBIDDEN, isUserTeacherOrAbove(userManager, user)
+                    ? "You can only mark assignments incomplete for groups you own or manage."
+                    : "You cannot complete someone else's quiz.").toResponse();
+            }
+
+            if ((quizAttempt.getCompletedDate() != null) == markComplete) {
```
jsharkey13, line 484: 👀 

(This is both very clever and also hard to read without thinking, maybe have a `currentCompletionStatus` variable or a comment?)

jsharkey13 (commented):
Two small things. 

Then I wonder whether, since you're using `markComplete` to both mark as complete and mark as incomplete if necessary, whether something like `updateAttemptCompletionStatus` might make more sense for the method names? (And/or call the variable `markComplete` something more like `newCompletionStatus`?) 
But we already have methods with names like this that have this do-and-undo nature, so if you think it's fine it's fine!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
robinmessage: I'm not sure if I'm 100% following.

I was assuming I would call this with ordinary POST params, not JSON.

But I think you're saying there is generally a preference for a packaged JSON object? If so, then I can modify APIs to do that instead. (Although this one I'd probably just split into two endpoints.)

jsharkey13: As far as I am aware, every POST endpoint we have consumes JSON, yes. I think the original way was to make a DTO for each endpoint, but those two places above are examples of being loose about what sort of JSON value is allowed at top-level, and there are [other examples with lists too](https://github.com/isaacphysics/isaac-api/blob/v2.15.2/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java#L372-L377).
But I have asked @mlt47 for an opinion when he's out of a meeting!

mlt47: I don't think it matters too much, but yeah sticking to JSON or a query param is what we've generally done

robinmessage: I've split into two endpoints, which I think actually makes them clearer anyway.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
robinmessage: This has gone away with the split into two endpoints.

PR #320, opened 2021-02-08
Quiz: log quiz section view

robinmessage:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (approved):
I had wondered when I saw that logging a quiz section view was a separate thing in the backend, why the logging of a section view (much like a hint view or accordion view) didn't just use the generic `/log` endpoint from the frontend; I still think it would have worked fine that way, but I can see this gives us much better constraints on data integrity without much additional complexity in the API 🎉

robinmessage:
> I had wondered when I saw that logging a quiz section view was a separate thing in the backend, why the logging of a section view (much like a hint view or accordion view) didn't just use the generic `/log` endpoint from the frontend; I still think it would have worked fine that way, but I can see this gives us much better constraints on data integrity without much additional complexity in the API 🎉

Hah, I didn't realise there was a generic `/log` endpoint I could use! But yeah, it's not too complex to wire up something that pulls useful info from the ids.

PR #322, opened 2021-02-09
Re-design the My Assignment quick view

Morpheu5:
Adds `notes` and `groupName` to assignments.

- `notes` is also added to the database.
- `groupName` is only added to `AssignmentDTO` and only returned when the requesting user has permission to do so.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -185,7 +185,9 @@ public class AssignmentFacade extends AbstractIsaacFacade {
 
                 // Augment with group name if allowed
                 UserGroupDTO group = groupManager.getGroupById(assignment.getGroupId());
-                assignment.setGroupName(group.getGroupName());
+                if (group.getLastUpdated() != null) {
+                    assignment.setGroupName(group.getGroupName());
+                }
             }
 
             this.assignmentService.augmentAssignerSummaries(assignments);
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -45,6 +45,7 @@ import uk.ac.cam.cl.dtg.segue.dao.SegueDatabaseException;
 import uk.ac.cam.cl.dtg.segue.dao.content.ContentManagerException;
 import uk.ac.cam.cl.dtg.segue.dos.LightweightQuestionValidationResponse;
 import uk.ac.cam.cl.dtg.segue.dos.QuestionValidationResponse;
+import uk.ac.cam.cl.dtg.segue.dos.UserGroup;
 import uk.ac.cam.cl.dtg.segue.dto.SegueErrorResponse;
 import uk.ac.cam.cl.dtg.segue.dto.UserGroupDTO;
 import uk.ac.cam.cl.dtg.segue.dto.content.QuestionDTO;
@@ -180,6 +181,13 @@ public class AssignmentFacade extends AbstractIsaacFacade {
             // we want to populate gameboard details for the assignment DTO.
             for (AssignmentDTO assignment : assignments) {
                 assignment.setGameboard(gameboardsMap.get(assignment.getGameboardId()));
+
+                // Augment with group name if allowed
+                // TODO: Do we want staff and/or admins here?
+                UserGroupDTO group = groupManager.getGroupById(assignment.getGroupId());
+                if (GroupManager.isOwnerOrAdditionalManager(group, currentlyLoggedInUser.getId())) {
+                    assignment.setGroupName(group.getGroupName());
+                }
             }
 
             this.assignmentService.augmentAssignerSummaries(assignments);
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -960,12 +960,14 @@ public class AssignmentFacade extends AbstractIsaacFacade {
             RegisteredUserDTO currentlyLoggedInUser = userManager.getCurrentRegisteredUser(request);
             UserGroupDTO assigneeGroup = groupManager.getGroupById(assignmentDTOFromClient.getGroupId());
 
-            if (!isUserTeacherOrAbove(userManager, currentlyLoggedInUser)) {
-                return new SegueErrorResponse(Status.FORBIDDEN, "You need a teacher account to create groups and set assignments!").toResponse();
-            }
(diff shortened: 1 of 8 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -956,6 +964,10 @@ public Response assignGameBoard(@Context final HttpServletRequest request,
                 return new SegueErrorResponse(Status.FORBIDDEN, "You need a teacher account to create groups and set assignments!").toResponse();
             }
 
+            if (!isUserStaff(userManager, currentlyLoggedInUser) && (null == assignmentDTOFromClient.getNotes() || !assignmentDTOFromClient.getNotes().isEmpty())) {
```
bh412, line 967: This blocks out anyone who isn't staff from creating an assignment! Conditionals need altering

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -180,6 +181,13 @@ public Response getAssignments(@Context final HttpServletRequest request,
             // we want to populate gameboard details for the assignment DTO.
             for (AssignmentDTO assignment : assignments) {
                 assignment.setGameboard(gameboardsMap.get(assignment.getGameboardId()));
+
+                // Augment with group name if allowed
+                // TODO: Do we want staff and/or admins here?
+                UserGroupDTO group = groupManager.getGroupById(assignment.getGroupId());
+                if (GroupManager.isOwnerOrAdditionalManager(group, currentlyLoggedInUser.getId())) {
+                    assignment.setGroupName(group.getGroupName());
```
bh412, line 189: This didn't appear to be firing correctly, the groupname wasn't returned for an owned group

bh412 (changes requested):
Needs conditionals altering otherwise breaks assignment setting for all except staff. The groupname and notes need double checking they actually return

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
Morpheu5: By design.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
Morpheu5: I can't reproduce this -- how did the problem present itself?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
bh412: This shouldn't be note dependant though, in this form it's stopping teachers from setting assignments to their groups of students, as they're not allowed to add notes, even though they don't/can't provide notes

bh412 (changes requested):
I understand only staff can add notes, but this bit of logic locks down setting assignments at all to just staff - It prevents teachers setting note-less assignments to their students

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -973,6 +985,12 @@ public Response assignGameBoard(@Context final HttpServletRequest request,
                         .toResponse();
             }
 
+            if (null != assignmentDTOFromClient.getNotes() && assignmentDTOFromClient.getNotes().length() > 240) {
```
bh412, line 988: I think we should move this out to a segue constant, like `MAX_NOTE_CHAR_LENGTH = 240` and then use that `.toString()` in the error response, so we don't end up with this 240/ 500 discrepancy.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
bh412: I was misinterpreting what it should do/ it's purpose, it seems to work as intended

bh412 (commented):
Functions as intended - if a little convoluted - TODO needs reviewing

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
jsharkey13: I'm not sure I'm clear on why we need this. But as well as owners and additional managers, students can see the group name if the group last_updated is not null:
https://github.com/isaacphysics/isaac-api/blob/v2.16.1/src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java#L214-L217

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -967,6 +980,17 @@ public Response assignGameBoard(@Context final HttpServletRequest request,
                         "You can only set assignments to groups you own or manage.").toResponse();
             }
 
+            if (userIsStaff) {
+                if (notesIsTooLong) {
+                    return new SegueErrorResponse(Status.BAD_REQUEST, "Your assignment notes exceed the maximum allowed length of " +
+                            MAX_NOTE_CHAR_LENGTH.toString() + " characters.").toResponse();
+                }
+            } else { // user is not staff
+                if (userIsTeacher && !notesIsNullOrEmpty) {
```
jsharkey13, line 989: Why do we need this `userIsTeacher`? There's a check above for at-least-teacher, so this is redundant. It can just be one if, with an `} else if (!notesIsNullOrEmpty) {`, no?

src/main/java/uk/ac/cam/cl/dtg/segue/api/AbstractSegueFacade.java
```diff
@@ -327,4 +327,9 @@ public static boolean isUserTeacherOrAbove(final UserAccountManager userManager,
             throws NoUserLoggedInException {
         return !userManager.checkUserRole(userDTO, Collections.singletonList(Role.STUDENT));
     }
+
+    public static boolean isUserTeacher(final UserAccountManager userManager, final RegisteredUserDTO userDTO)
```
jsharkey13, line 331: Am I right in thinking we do not need this? Generally we have things admin can do, things staff can do, things teachers or above can do, and things anyone can do. I don't think we want anything that explicitly only applies to teachers except by exclusion. And I think the place that calls this doesn't need it either?

jsharkey13 (commented):
Generally this looks ok; one or two questions and one thing I think we don't need and don't want?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
Morpheu5: a0f231c0 should fix this.

jsharkey13 (approved):

PR #324, opened 2021-02-11
Feature/audit logs

mlt47:
This API change requires the following front-end change: https://github.com/isaacphysics/isaac-react-app/pull/356

Entire log path is recorded as a shared volume of the compose script so no need to alter that.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
@@ -320,6 +320,14 @@ public class UserAuthenticationManager {
             }
         }
 
+        // Fallback to use JSESSIONID
+        if (null == userIdentifier) {
+            try {
+                userIdentifier = this.getJSessionIdFromRequest(request);
+            } catch (InvalidSessionException ex) {
+                errors.add("no jsessionid found");
+            }
+        }
 
         // No identifier found
         if (null == userIdentifier) {
@@ -1099,6 +1107,25 @@ public class UserAuthenticationManager {
                 HashMap.class);
 
         return sessionInformation;
+    }
+
+    private String getJSessionIdFromRequest(final HttpServletRequest request) throws InvalidSessionException {
+        Cookie jSessionCookie = null;
+        if (request.getCookies() == null) {
+            throw new InvalidSessionException("There are no cookies set.");
+        }
+
+        for (Cookie c : request.getCookies()) {
+            if (c.getName().equals(JSESSION_COOOKIE)) {
+                jSessionCookie = c;
+            }
+        }
+
+        if (null == jSessionCookie) {
+            throw new InvalidSessionException("There are no cookies set.");
+        }
+
+        return jSessionCookie.getValue();
     }
 
     /**
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
@@ -17,6 +17,7 @@ package uk.ac.cam.cl.dtg.segue.api.managers;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.api.client.util.Lists;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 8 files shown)


jsharkey13:
I know this will now be a pain, having looked at the code, but I had hoped we could record both the `JSESSIONID` and the Segue user ID (or null), since both separately are useful for working out what is going on.

mlt47:
> I know this will now be a pain, having looked at the code, but I had hoped we could record both the `JSESSIONID` and the Segue user ID (or null), since both separately are useful for working out what is going on.

I had though that that could maybe be useful, that's OK shouldn't be too difficult 👍

src/main/java/uk/ac/cam/cl/dtg/segue/api/monitors/AuditMonitor.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Inject
+    public AuditMonitor(final UserAuthenticationManager userAuthenticationManager) {
+        this.userAuthenticationManager = userAuthenticationManager;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext containerRequestContext, ContainerResponseContext containerResponseContext) {
+        String plausibleUserId = userAuthenticationManager.getPlausibleUserIdentifierFromCookie(httpServletRequest);
+        log.trace(String.format("User (%s) requested %s \"%s\" with response: %d [%s]",
```
jsharkey13, line 34: Could this not be in some more machine-readable format? One thing that took ages parsing the Nginx logs the other day was corralling them into a format I could open with Excel and use in Python. Comma separated might be enough, since none of these things look to be allowed to have commas in.
I can't imagine we'll ever want to read it by hand, but I do think we should reivew it in an aggregated form regularly.

I was sort of hoping for `ip_address, jsessionid, segue_user_id, http_method, request_path, response_code`.

How will URL parameters interact? Search for instance might get messy if we log the URL param string. We might also want the canonical endpoint URL, like we have for Prometheus, so we can aggregate/search by endpoint more readily without having to worry about different values?

mlt47: OK yes that is sensible and achievable. I've noticed search in the perf logs sometimes being rather long! We should add a word limit to that ES query. Only recording the canonical endpoint would leave out information which might be important. Might be a good idea to move search from using a URL param string to query string.
Revisiting this will let me add documentation to some of these methods too.

jsharkey13: I actually meant to log both the canonical endpoint and the specific request URL path
`ip_address, jsessionid, segue_user_id, http_method, canonical_path, request_path, response_code`
since it's not _that_ much more additional space.

jsharkey13: I guess we could also log both the `segue_user_id` from the cookie and whether the HMAC was valid as separate items, in case we wanted to have logs that would let us investigate session-guessing brute force attacks / attempted spoofing. We still can't do session-reuse with an expired session token this way, but that's a much nicher and smaller attack surface.

`ip_address, jsessionid, segue_user_id, session_token, valid_hmac, http_method, canonical_path, request_path, response_code`

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -133,6 +122,33 @@ public UserAuthenticationManager(final IUserDataManager database,
...
+     * This method is used to extract user-identifying features of a request and return them in csv format.
+     * NOTE: The validity of the session token is not checked against the database.
+     *
+     * For interest the IP address format is described in the REMOTE_ADDR section of <a href="https://www.ietf.org/rfc/rfc3875">RFC 3875</>.
+     *
+     * @param request - http request from which to extract user identifying features.
+     * @return A string of comma-separated user identifying values from the request.
+     */
+    public String getUserIdentifierCsv(final HttpServletRequest request) {
+        String ipAddress = request.getRemoteAddr();
```
jsharkey13, line 135: On our servers (unlike local testing), this will always return the `10.x.x.x` IP address of the `isaac-router` proxy, since that is the other end of the connection. I recommend using `uk.ac.cam.cl.dtg.util.RequestIPExtractor::getClientIpAddr()`, which I built to do the right thing in all the messy cases we see!

mlt47: Nice! Hadn't seen that function, had wondered whether it would still know where it was coming from after the router/gateway.

jsharkey13 (approved):
I'm excited to see how this works in practice!
One last thing that occurs to me; does the AuditMonitor filter happen inside the PerformanceMonitor timings, or outside it? I.e. will we be able to see if this causes a performance overhead, or will that monitoring miss any delay this adds? Does the order the filters are registered matter?

mlt47:
> I'm excited to see how this works in practice!
> One last thing that occurs to me; does the AuditMonitor filter happen inside the PerformanceMonitor timings, or outside it? I.e. will we be able to see if this causes a performance overhead, or will that monitoring miss any delay this adds? Does the order the filters are registered matter?

Good question. It looks like we can apply a priority to the filter. Thankfully, from testing, priority 0 (or low) means apply first on request, last on response which is exactly what we want for the performance monitor.

jsharkey13 (approved):

PR #327, opened 2021-02-15
Impose 1000 character limit for search terms

bh412:
---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
@@ -189,9 +189,12 @@ public class IsaacController extends AbstractIsaacFacade {
             @DefaultValue(DEFAULT_START_INDEX_AS_STRING) @QueryParam("start_index") final Integer startIndex,
             @DefaultValue(DEFAULT_SEARCH_RESULT_LIMIT_AS_STRING) @QueryParam("limit") final Integer limit) {
 
+        // Impose 1000 character search string limit internally
+        String internalSearchString = searchString.substring(0, Math.min(searchString.length(), 1000));
+
         // Calculate the ETag on current live version of the content
         // NOTE: Assumes that the latest version of the content is being used.
-        EntityTag etag = new EntityTag(this.contentIndex.hashCode() + searchString.hashCode()
+        EntityTag etag = new EntityTag(this.contentIndex.hashCode() + internalSearchString.hashCode()
                 + types.hashCode() + "");
 
         Response cachedResponse = generateCachedResponse(request, etag);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
```diff
@@ -189,9 +189,12 @@ public final Response search(@Context final Request request, @Context final Http
             @DefaultValue(DEFAULT_START_INDEX_AS_STRING) @QueryParam("start_index") final Integer startIndex,
             @DefaultValue(DEFAULT_SEARCH_RESULT_LIMIT_AS_STRING) @QueryParam("limit") final Integer limit) {
 
+        // Impose 1000 character search string limit internally
+        String internalSearchString = searchString.substring(0, Math.min(searchString.length(), 1000));
```
jsharkey13, line 193: I think this should be a constant like `SEARCH_TEXT_CHAR_LIMIT` of 1000, and then if the `searchString` is longer than that we return a 400 Bad Request and not truncate it.
The front end can do something nicer like warn you, and have an even shorter limit.

bh412: Ahh I thought the idea was to silently truncate - I'll alter this

src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
```diff
@@ -189,6 +189,12 @@ public final Response search(@Context final Request request, @Context final Http
             @DefaultValue(DEFAULT_START_INDEX_AS_STRING) @QueryParam("start_index") final Integer startIndex,
             @DefaultValue(DEFAULT_SEARCH_RESULT_LIMIT_AS_STRING) @QueryParam("limit") final Integer limit) {
 
+        // Impose 1000 character search string limit internally
+        if (searchString.length() > SEARCH_TEXT_CHAR_LIMIT) {
```
jsharkey13, line 193: Does IntelliJ think this is safe? Is it that it wouldn't have matched the endpoint if searchString was null, and so you don't need a null check?

bh412: It infers the @notnull annotation - searchString has to be a string or it throws a 'can't find resource' when you try and hit it

jsharkey13 (approved):
Is there a corresponding front end PR to ensure this can't be hit by accident and that the error messages shown are obviously visible and explain the issue?

bh412:
I've added a front end PR with a search box max length of 255 characters. The returned toast seems clear enough to me

PR #332, opened 2021-02-22
Add question category option to gameboard filter

mlt47:
This PR depends on https://github.com/isaacphysics/isaac-react-app/pull/360

This will allow users to conditionally provide a "question category" parameter to the question filter. If present it will only return questions tagged with that question category. Not providing a question category is valid and is likely to be what CS want to use in the future.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
@@ -117,16 +117,18 @@ public class GameManager {
      *
      * @param title
      *            title of the board
-     * @param subjectsList
+     * @param subjects
      *            list of subjects to include in filtered results
-     * @param fieldsList
+     * @param fields
      *            list of fields to include in filtered results
-     * @param topicsList
+     * @param topics
      *            list of topics to include in filtered results
-     * @param levelsList
+     * @param levels
      *            list of levels to include in filtered results
-     * @param conceptsList
+     * @param concepts
      *            list of concepts (relatedContent) to include in filtered results
+     * @param questionCategories
+     *            list of question categories (i.e. problem_solving, book) to include in filtered results
      * @param boardOwner
      *            The user that should be marked as the creator of the gameBoard.
      * @return a gameboard if possible that satisifies the conditions provided by the parameters. Will return null if no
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -1235,14 +1243,9 @@ private IsaacWildcard getWildCardById(final String id) throws NoWildcardExceptio
...
         }
 
+        // handle exclusions
         List<String> tagsToExclude = Lists.newArrayList();
-        // add no filter constraint
-        tagsToExclude.add(HIDE_FROM_FILTER_TAG);
-
-        // Temporarily ignore book and quick_quiz question types.
-        // Strings hardcoded until question types are passed in as part of the query.
-        tagsToExclude.addAll(ImmutableList.of("quick_quiz", "book"));
```
bh412, line 1244: Does this rely on content having tagged all of the book/ quick_quiz questions not as problem_solving? Or is it a 'whitelist' and it's fine as it defaults to just be the problem_solving tagged ones now. I couldn't get it to cause an issue so I think it's fine

bh412 (commented):
Neat solution - refactoring all makes sense too - just the comment to clarify but all seems good

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
mlt47: Before this pull request, we did not return anything from the filter if it was tagged `quick_quiz` or `book` to allow the content team to add subject and topic tags to the books/quick_quizes.
Now we're flipping it. The accompanying frontend PR will add a `questionCategory=problem_solving` to the gameboard query. If that is present we will only return questions tagged `problem_solving`. Eventually they will be able to set `questionCategory=book,quick_quizes` and return those types of questions but we're not surfacing that to the users yet.

This relies on the content team having tagged all problem solving questions with the tag `problem_solving` from my understanding, and from inspection, this is done. They can now go ahead and tag the maths questions as both `problem_solving` and `book` (they are currently only `problem_solving` because having `book` excluded them from the results before this PR).

bh412: Sounds good - thanks for clarifying

PR #336, opened 2021-03-09
Quiz: tweak endpoint to make FE simpler

robinmessage:
---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizManager.java
@@ -25,6 +25,8 @@ import uk.ac.cam.cl.dtg.isaac.dos.IsaacQuiz;
 import uk.ac.cam.cl.dtg.isaac.dto.IHasQuizSummary;
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacQuizDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacQuizSectionDTO;
+import uk.ac.cam.cl.dtg.isaac.dto.QuizAssignmentDTO;
+import uk.ac.cam.cl.dtg.isaac.dto.QuizAttemptDTO;
 import uk.ac.cam.cl.dtg.segue.api.Constants;
 import uk.ac.cam.cl.dtg.segue.api.services.ContentService;
 import uk.ac.cam.cl.dtg.segue.dao.content.ContentManagerException;
@@ -114,13 +116,13 @@ public class QuizManager {
      *
      * @param quizId the id of the quiz.
      *
-     * @return The quiz or null.
+     * @return The quiz.
      */
     public IsaacQuizDTO findQuiz(final String quizId) throws ContentManagerException {
         Content cachedContent = contentManager.getContentDOById(contentIndex, quizId);
 
         if (cachedContent == null) {
-            return null;
+            throw new ContentManagerException("Couldn't find quiz with id " + quizId);
         }
 
         if (cachedContent instanceof IsaacQuiz) {
@@ -132,7 +134,7 @@ public class QuizManager {
                 throw new ContentManagerException("Expected an IsaacQuizDTO, got a " + contentDTO.getType());
             }
         } else {
-            throw new ContentManagerException("Expected an IsaacQuiz, got a " + cachedContent.getType());
+            throw new ContentManagerException("Expected an IsaacQuiz (id=" + quizId + "), got a " + cachedContent.getType());
         }
     }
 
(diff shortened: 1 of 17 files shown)


mlt47:
Not directly related to the changes in the PR but a problem in the API rather than the front-end is that if a user visits `/quizzes` before being assigned a quiz there is a database exception which is displayed to the user as an Internal Server Error.

robinmessage:
> Not directly related to the changes in the PR but a problem in the API rather than the front-end is that if a user visits `/quizzes` before being assigned a quiz there is a database exception which is displayed to the user as an Internal Server Error.

Good point, now fixed.

mlt47 (commented):
This code looks good to me, however I have been receiving the incorrect numbers for not attempted question parts in the teacher report, so there might be a bug there (perhaps related to trying a question more than once?). Let me know if you can't reproduce it.

robinmessage:
> This code looks good to me, however I have been receiving the incorrect numbers for not attempted question parts in the teacher report, so there might be a bug there (perhaps related to trying a question more than once?). Let me know if you can't reproduce it.

Hey Meurig, any idea what specifically caused this? I've tried to reproduce but not got anywhere I'm afraid. I could well be missing something obvious by doing the same things over and over.

mlt47:
> > This code looks good to me, however I have been receiving the incorrect numbers for not attempted question parts in the teacher report, so there might be a bug there (perhaps related to trying a question more than once?). Let me know if you can't reproduce it.
> 
> Hey Meurig, any idea what specifically caused this? I've tried to reproduce but not got anywhere I'm afraid. I could well be missing something obvious by doing the same things over and over.

I think it might be that `PgQuizQuestionAttemptPersistenceManager::getAllAnswersForQuizAssignments(...)` needs a compute if absent rather than compute
```
 Map<String, List<QuestionValidationResponse>> userAttemptsMap = resultsMap.computeIfAbsent(userId, (ignore) -> Maps.newHashMap());
```

robinmessage:
> I think it might be that `PgQuizQuestionAttemptPersistenceManager::getAllAnswersForQuizAssignments(...)` needs a compute if absent rather than compute
> 
> ```
>  Map<String, List<QuestionValidationResponse>> userAttemptsMap = resultsMap.computeIfAbsent(userId, (ignore) -> Maps.newHashMap());
> ```

Thanks Meurig! That's exactly it. I was spent a while thinking about duplicate answers to the same question I never considered that bug. I really should have considered what I expected to be passed in as `ignoreValue` for a new entry...
Anyway, I've been able to reproduce and confirm your fix is correct.

mlt47:
@robinmessage, I've just remembered I came across a separate small problem in the API while looking at this yesterday:
If one of the team where to (accidentally?) unpublish a quiz that had been assigned, `getQuizAssignments(...)` (and possibly others) would return a 500 internal server error from an exception in `augmentQuizSummary(...)`. This does sometimes happen, but I noticed it because we often switch between Physics and CS content during development.

robinmessage:
> @robinmessage, I've just remembered I came across a separate small problem in the API while looking at this yesterday:
> If one of the team where to (accidentally?) unpublish a quiz that had been assigned, `getQuizAssignments(...)` (and possibly others) would return a 500 internal server error from an exception in `augmentQuizSummary(...)`. This does sometimes happen, but I noticed it because we often switch between Physics and CS content during development.

Oh yeah, that's not a good behaviour. For now, I'll exclude those being visible as the front end can't do a lot without the quiz summary.

mlt47 (approved):

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizManager.java
```diff
@@ -128,23 +130,37 @@ public IsaacQuizDTO findQuiz(final String quizId) throws ContentManagerException
                 throw new ContentManagerException("Expected an IsaacQuizDTO, got a " + contentDTO.getType());
             }
         } else {
-            throw new ContentManagerException("Expected an IsaacQuiz, got a " + cachedContent.getType());
+            if (cachedContent != null) {
+                throw new ContentManagerException("Expected an IsaacQuiz (id=" + quizId + "), got a " + cachedContent.getType());
+            } else {
+                throw new ContentManagerException("Couldn't find quiz with id " + quizId);
+            }
```
mlt47, line 137: Have you noticed that the static analysis is saying warning of a useless null check? https://lgtm.com/projects/g/isaacphysics/isaac-api/rev/pr-3f4ff4ea4a09034c27f4a9c4b5e580137c11d2e1
Looks like it might be an odd merge as it is not true on this branch.

mlt47 (approved):

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/QuizManager.java
robinmessage: Yeap, it's from master. I think you noticed the NPE and fixed it on master, and I fixed it separately but in a slightly different way, which is now redundant.

However, I don't like the `null` return, so I've made a combined fix on this branch.

Also, I had named a commit incorrectly so I've rebased this branch.

mlt47: Great, thanks!

PR #337, opened 2021-03-15
Synchronise user list with MailJet

jsharkey13:
This is still a work in progress, since bad things will happen if MailJet is not configured. But a review at this stage would be informative!

---

**Pull Request Check List**
- [x] Security - New dependency - Searched for any know vulnerabilities
- [x] Security - New dependency - Added to dependency list
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -39,6 +39,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.managers.EventBookingManager;
 import uk.ac.cam.cl.dtg.segue.api.Constants.*;
+import uk.ac.cam.cl.dtg.segue.api.managers.ExternalAccountManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.SegueResourceMisuseException;
 import uk.ac.cam.cl.dtg.segue.api.managers.StatisticsManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
@@ -137,6 +138,7 @@ public class AdminFacade extends AbstractSegueFacade {
     private final AbstractUserPreferenceManager userPreferenceManager;
     private final EventBookingManager eventBookingManager;
     private final UserAssociationManager associationManager;
+    private final ExternalAccountManager externalAccountManager;
     private final IMisuseMonitor misuseMonitor;
 
     /**
@@ -169,7 +171,7 @@ public class AdminFacade extends AbstractSegueFacade {
                        final StatisticsManager statsManager, final LocationManager locationManager,
                        final SchoolListReader schoolReader, final AbstractUserPreferenceManager userPreferenceManager,
                        final EventBookingManager eventBookingManager, final UserAssociationManager associationManager,
-                       final IMisuseMonitor misuseMonitor) {
+                       final ExternalAccountManager externalAccountManager, final IMisuseMonitor misuseMonitor) {
         super(properties, logManager);
         this.userManager = userManager;
         this.contentManager = contentManager;
@@ -180,6 +182,7 @@ public class AdminFacade extends AbstractSegueFacade {
         this.userPreferenceManager = userPreferenceManager;
         this.eventBookingManager = eventBookingManager;
         this.associationManager = associationManager;
+        this.externalAccountManager = externalAccountManager;
         this.misuseMonitor = misuseMonitor;
     }
 
@@ -1612,6 +1615,28 @@ public class AdminFacade extends AbstractSegueFacade {
             } else {
                 return Response.ok(ImmutableMap.of("status", "Nothing to reset.")).build();
             }
+        } catch (NoUserLoggedInException e) {
+            return SegueErrorResponse.getNotLoggedInResponse();
+        }
+    }
+
+    /**
+     *  Manually trigger a sync. Unfinished.
+     */
+    @GET
+    @Path("/sync_external_accounts")
+    @ApiOperation(value = "Trigger an update for external providers where account details have changed.")
+    public Response syncExternalAccounts(@Context final HttpServletRequest httpServletRequest) {
+        //TODO - either remove this or add sensible error reporting!
+        try {
+            RegisteredUserDTO user = userManager.getCurrentRegisteredUser(httpServletRequest);
+            if (!isUserAnAdmin(userManager, user)) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+
+            externalAccountManager.processRecentlyUpdatedUsers();
+
+            return Response.ok().build();
         } catch (NoUserLoggedInException e) {
             return SegueErrorResponse.getNotLoggedInResponse();
         }
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -1622,13 +1622,13 @@ public class AdminFacade extends AbstractSegueFacade {
     }
 
     /**
-     *  Manually trigger a sync. Unfinished.
+     *  Manually trigger a sync for testing or debugging purposes. Minimal success or failure reporting.
      */
     @GET
     @Path("/sync_external_accounts")
     @ApiOperation(value = "Trigger an update for external providers where account details have changed.")
     public Response syncExternalAccounts(@Context final HttpServletRequest httpServletRequest) {
-        //TODO - remove this?
+        //TODO - automate this with Quartz, then review if this is still necessary?
         try {
             RegisteredUserDTO user = userManager.getCurrentRegisteredUser(httpServletRequest);
             if (!isUserAnAdmin(userManager, user)) {
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -40,6 +40,7 @@ import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.managers.EventBookingManager;
 import uk.ac.cam.cl.dtg.segue.api.Constants.*;
 import uk.ac.cam.cl.dtg.segue.api.managers.ExternalAccountManager;
+import uk.ac.cam.cl.dtg.segue.api.managers.ExternalAccountSynchronisationException;
 import uk.ac.cam.cl.dtg.segue.api.managers.SegueResourceMisuseException;
 import uk.ac.cam.cl.dtg.segue.api.managers.StatisticsManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
@@ -1627,18 +1628,21 @@ public class AdminFacade extends AbstractSegueFacade {
     @Path("/sync_external_accounts")
     @ApiOperation(value = "Trigger an update for external providers where account details have changed.")
     public Response syncExternalAccounts(@Context final HttpServletRequest httpServletRequest) {
-        //TODO - either remove this or add sensible error reporting!
+        //TODO - remove this?
         try {
             RegisteredUserDTO user = userManager.getCurrentRegisteredUser(httpServletRequest);
             if (!isUserAnAdmin(userManager, user)) {
                 return SegueErrorResponse.getIncorrectRoleResponse();
             }
 
-            externalAccountManager.processRecentlyUpdatedUsers();
+            externalAccountManager.synchroniseChangedUsers();
 
             return Response.ok().build();
         } catch (NoUserLoggedInException e) {
             return SegueErrorResponse.getNotLoggedInResponse();
+        } catch (ExternalAccountSynchronisationException e) {
+            return new SegueErrorResponse(Status.INTERNAL_SERVER_ERROR,
+                    "Database error while attempting to synchronise users!", e).toResponse();
         }
     }
 }
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -39,8 +39,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.managers.EventBookingManager;
 import uk.ac.cam.cl.dtg.segue.api.Constants.*;
-import uk.ac.cam.cl.dtg.segue.api.managers.ExternalAccountManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.ExternalAccountSynchronisationException;
+import uk.ac.cam.cl.dtg.segue.api.managers.IExternalAccountManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.SegueResourceMisuseException;
 import uk.ac.cam.cl.dtg.segue.api.managers.StatisticsManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
@@ -139,7 +139,7 @@ public class AdminFacade extends AbstractSegueFacade {
     private final AbstractUserPreferenceManager userPreferenceManager;
     private final EventBookingManager eventBookingManager;
     private final UserAssociationManager associationManager;
-    private final ExternalAccountManager externalAccountManager;
+    private final IExternalAccountManager externalAccountManager;
     private final IMisuseMonitor misuseMonitor;
 
     /**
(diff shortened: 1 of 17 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dos/users/UserExternalAccountChanges.java
```diff
@@ -0,0 +1,119 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.cam.cl.dtg.segue.dos.users;
+
+/**
+ *  A class to represent the state of a user that needs synchronising with an external provider.
+ */
```
sacummins, line 20: Just for clarity, is this used strictly as our internal DO or does it get serialised and traverse the network to Mailjet - don't mind if it does but could be worth putting in the javadoc for clarity?

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+    private final MailjetClient mailjetClient;
+    private final String newsListId;
+    private final String eventsListId;
+
+    /**
+     *  Wrapper for MailjetClient class.
+     * @param properties - needed to get the MailJet keys and list IDs
+     */
+    @Inject
+    public MailJetApiClientWrapper(final PropertiesLoader properties) {
```
sacummins, line 54: Rather than passing in the whole properties loader can we use Guice to magically create one of these with clear dependencies in the constructor params?

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+     */
+    public JSONObject getAccountByIdOrEmail(final String mailjetIdOrEmail) throws MailjetException {
+        if (null == mailjetIdOrEmail) {
+            return null;
+        }
+        MailjetRequest request = new MailjetRequest(Contact.resource, mailjetIdOrEmail);
+        MailjetResponse response = mailjetClient.get(request);
+        JSONArray responseData = response.getData();
+        if (response.getTotal() == 1) {
+            return responseData.getJSONObject(0);
```
sacummins, line 80: Would be nice if we could marshall this into an object that has types etc, although in the short term this is probably ok for early testing.

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+     * @param mailjetId - MailJet user ID
+     * @throws MailjetException  - if underlying MailjetClient throws an exception
+     */
+    public void permanentlyDeleteAccountById(final String mailjetId) throws MailjetException {
+        Validate.notNull(mailjetId);
+        MailjetRequest request = new MailjetRequest(Contacts.resource, mailjetId);
+        MailjetResponse response = mailjetClient.delete(request);
+        int responseStatus = response.getStatus();
+        if (!(responseStatus == Response.Status.OK.getStatusCode())) {
+            // TODO: Do we want to get any of the data from this request?
```
sacummins, line 97: Might be good to log that the delete was successful somewhere just in case we have to justify our request to delete something.

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+        try {
+            MailjetRequest request = new MailjetRequest(Contact.resource).property(Contact.EMAIL, email);
+            MailjetResponse response = mailjetClient.post(request);
+            // Get MailJet ID out:
+            JSONObject responseData = response.getData().getJSONObject(0);
+            return Integer.toString(responseData.getInt("ID"));
+        } catch (MailjetClientRequestException e) {
+            if (e.getMessage().contains("already exists")) {
+                // FIXME - we need to test that this response always comes back with "already exists" in the message
+                log.warn(String.format("Attempted to create a user with email (%s) that already existed!", email));
```
sacummins, line 122: In this event should we throw an exception or is the consumer likely not to care? In which case do we want to modify the method name to indicate it is addOrGetUser?

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param mailjetId - MailJet user ID
+     * @throws MailjetException  - if underlying MailjetClient throws an exception
+     */
+    public void updateUserProperties(final String mailjetId, final String firstName, final String role,
+                                      final String email_verification_status) throws MailjetException {
+        Validate.notNull(mailjetId);
+        MailjetRequest request = new MailjetRequest(Contactdata.resource, mailjetId)
+                .property(Contactdata.DATA, new JSONArray()
+                        .put(new JSONObject().put("Name", "firstname").put("value", firstName))
```
sacummins, line 145: Again it would be nice in the future to have DTOs or something to help make this look a bit nicer? Although not urgent for now.

src/main/resources/db_scripts/create_external_accounts_table.sql
```diff
@@ -0,0 +1,13 @@
...
+   user_id INTEGER NOT NULL,
+   provider_name TEXT NOT NULL,
+   provider_user_identifier TEXT,
+   provider_last_updated TIMESTAMP WITHOUT TIME ZONE,
+   CONSTRAINT external_accounts_pk PRIMARY KEY (user_id, provider_name),
+   CONSTRAINT external_accounts_fk FOREIGN KEY (user_id) REFERENCES public.users (id)
+       ON UPDATE CASCADE ON DELETE CASCADE
+);
+
+-- Add stub records to prevent synchronisation?
```
sacummins, line 11: Shoudl this just be a config value to turn sync on or off globally or is it important we can sync certain records and not others?

src/main/resources/db_scripts/postgres-rutherford-create-script.sql
```diff
@@ -109,6 +109,20 @@ ALTER TABLE public.event_bookings_id_seq OWNER TO rutherford;
...
+--
+-- Name: external_accounts; Type: TABLE; Schema: public; Owner: rutherford
+--
+
+CREATE TABLE public.external_accounts (
+    user_id integer NOT NULL,
+    provider_name text NOT NULL,
+    provider_user_identifier text,
+    provider_last_updated timestamp without time zone
+);
```
sacummins, line 121: This looks very similar to the thirdparty authenticators table. I wonder if in the future they could be combined or if we think that is a bad idea. Party of my likes to keep authentication stuff nice and separate but it does seem to be fulfilling the same purpose - i.e. linking an isaac user to another identity elsewhere

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
```diff
@@ -0,0 +1,181 @@
...
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.ac.cam.cl.dtg.util.PropertiesLoader;
+
+import javax.ws.rs.core.Response;
+
+import static uk.ac.cam.cl.dtg.segue.api.Constants.*;
+
+public class MailJetApiClientWrapper {
```
sacummins, line 42: I would maybe have considered adding some kind of interface to allow us to change our mind more easily on mail provider but this could be refactored later if we think there is a chance we will want to change.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
```diff
@@ -0,0 +1,143 @@
...
+import uk.ac.cam.cl.dtg.segue.dao.SegueDatabaseException;
+import uk.ac.cam.cl.dtg.segue.dao.users.IExternalAccountDataManager;
+import uk.ac.cam.cl.dtg.segue.dos.users.EmailVerificationStatus;
+import uk.ac.cam.cl.dtg.segue.dos.users.UserExternalAccountChanges;
+import uk.ac.cam.cl.dtg.util.email.MailJetApiClientWrapper;
+import uk.ac.cam.cl.dtg.util.email.MailJetSubscriptionAction;
+
+import java.util.List;
+
+public class ExternalAccountManager {
```
sacummins, line 34: In terms of function could be worth adding a bit of JavaDoc to explain what external accounts this thing is intended to manage. Is it really just email or do we think it would be easily extended to do other stuff in the future?

sacummins (approved):
Seems like a sensible implementation, might be good to have a couple of tests in terms of strange sync behaviours? I made a few comments but I think you already knew I would say those things - none of them are urgent? Happy to help refactor once it has gone in and been tested.

@jsharkey13 In terms of your question, about turning on / off sync per API a stub that doesn't do anything is fine for now. You might like to make it a config option in the properties or something and tell Guice to create objects accordingly. 

Do we think that MailJet will replace all email functionality in the site in the future? Or will we always want some kind of local email functionality?

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
```diff
@@ -0,0 +1,143 @@
...
+                            //    Expect: "deleted" but non-null "mailjet_id"
+                            //    Action: GDPR deletion, null out MailJet ID?, update provider_last_updated
+                            log.debug("Case: deletion.");
+                            deleteUserFromMailJet(mailjetId, userRecord);
+                        } else if (EmailVerificationStatus.DELIVERY_FAILED.equals(userRecord.getEmailVerificationStatus())) {
+                            // Case: DELIVERY_FAILED but already on MailJet
+                            //    Expect: DELIVERY_FAILED, but non-null "mailjet_id"
+                            //    Action: same as deletion? Or just remove from lists for now?
+                            log.debug("Case: delivery failed.");
+                            mailjetApi.updateUserSubscriptions(mailjetId, MailJetSubscriptionAction.REMOVE, MailJetSubscriptionAction.REMOVE);
```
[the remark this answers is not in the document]
jsharkey13: MailJet doesn't email those it marks as delivery failed, but we would like to remove them from the lists anyway.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
```diff
@@ -0,0 +1,143 @@
...
+                        } else if (!accountEmail.equals(mailjetDetails.getString("Email"))) {
```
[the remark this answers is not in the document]
jsharkey13: I'd discussed this with the others. Since between 1/3 and 1/2 of users do not verify their email addresses, we're including the not verified users too (with a note that they are not verified to allow segmenting). It's worth noting that to have _changed_ an account email does require verifying the new address so this particular code path shouldn't need to worry about that either way.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
```diff
@@ -0,0 +1,143 @@
...
+                            log.debug("Case: invalid/incorrect user to skip.");
+                            database.updateExternalAccount(userId, null);
+                        }
+                    }
+                    // Iff action done successfully, update the provider_last_updated time:
+                    log.debug("Update provider_last_updated.");
+                    database.updateProviderLastUpdated(userId);
+
+                } catch (SegueDatabaseException e) {
+                    log.error(String.format("Error storing record of MailJet update to user (%s)!", userId));
```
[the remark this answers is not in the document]
jsharkey13: All this code is asynchronous and if the last updated isn't modified will just run again the next time the sync is triggered.
Creation and updating are idempotent and so safe, I should check about deletion since that may throw an error if attempted twice.

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
```diff
@@ -1616,4 +1620,29 @@ public Response resetMisuseMonitor(@Context final HttpServletRequest request,
...
     }
+
+    /**
+     *  Manually trigger a sync. Unfinished.
+     */
+    @GET
+    @Path("/sync_external_accounts")
+    @ApiOperation(value = "Trigger an update for external providers where account details have changed.")
+    public Response syncExternalAccounts(@Context final HttpServletRequest httpServletRequest) {
+        //TODO - remove this?
```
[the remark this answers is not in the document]
jsharkey13: We do actually want to be able to trigger it this way for now, sadly. It just provides no meaningful progress or much error reporting without monitoring the server logs, hence the "Unfinished". Perhaps I will update the comment to clarify.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
```diff
@@ -0,0 +1,143 @@
...
+        MailJetSubscriptionAction eventsStatus = (userRecord.allowsEventsEmails() != null && userRecord.allowsEventsEmails()) ? MailJetSubscriptionAction.FORCE_SUBSCRIBE : MailJetSubscriptionAction.UNSUBSCRIBE;
+        mailjetApi.updateUserSubscriptions(mailjetId, newsStatus, eventsStatus);
+
+        database.updateExternalAccount(userId, mailjetId);
+    }
+
+    private void deleteUserFromMailJet(final String mailjetId, final UserExternalAccountChanges userRecord) throws SegueDatabaseException, MailjetException {
+        Long userId = userRecord.getUserId();
+        mailjetApi.permanentlyDeleteAccountById(mailjetId);
+        // Do we need to keep this row?
```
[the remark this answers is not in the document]
jsharkey13: Yes, and right now the code uses the row with null in it to judge whether to sync, so I might remove these comments.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgExternalAccountPersistenceManager.java
```diff
@@ -0,0 +1,115 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * This class is responsible for managing and persisting user data.
+ */
+public class PgExternalAccountPersistenceManager implements IExternalAccountDataManager {
```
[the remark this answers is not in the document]
jsharkey13: This whole PR is a bit of a mess of things that attempt to be generic, and things that are highly specialised to MailJet's weird API surface and requirements.
It isn't just email settings, it is also anything in the `users` table (but not groups, assignments, connections, etc).

With hindsight, I think the interface plan here was a poor choice and it should have been MailJet specific. I wouldn't like to force ourselves to use an inefficient approach for other providers that implement a more usable set of API calls.

src/main/java/uk/ac/cam/cl/dtg/segue/dos/users/UserExternalAccountChanges.java
jsharkey13: MailJet doesn't allow you to create a user and update their properties at the same time, and you cannot pass the JSON directly but JSON describing the JSON of the user; so this is entirely internal and split into pieces to be used in the various cases. I will update the JavaDoc.

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
jsharkey13: I was just trying to make it idempotent in case things went wrong, or the user has already been synced. Happy to rename it!

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
jsharkey13: Well their Java client requires `JSONObjects` in that format, so we'd have to make a DTO that could convert itself into a `JSONObject` of the right form. If we ever needed it in more than one place, it might be worth thinking about, but since each endpoint needs a different style of `JSONObject` so far it seemed unnecessary!

src/main/resources/db_scripts/create_external_accounts_table.sql
jsharkey13: This was mostly intended as a bit of a hack so we could test it with just staff users first; we will want to sync some accounts and not others and this uses the functionality built into the sync process for deleted or invalid accounts to avoid being synced.

src/main/resources/db_scripts/postgres-rutherford-create-script.sql
jsharkey13: I thought that too, but the code for SSO providers assumes all rows in this table correspond to valid `AuthenticationProvider` enum values, and trying to rewrite all that code seemed much too much effort.

src/main/java/uk/ac/cam/cl/dtg/util/email/MailJetApiClientWrapper.java
jsharkey13: Yes, this is a balance between baking into an interface the inefficient design decisions of MailJet's API (cannot bulk create users with IDs, cannot add users with properties except when bulk subscribing, etc, etc) or being flexible enough to be more efficient with another provider in the future.
In the same way that 5 years after we removed Mongo we're still stuck with the poor implementation decisions the old interfaces force onto us, which cannot make use of the efficiency of a relational database for doing the work under the hood.

Morpheu5 (approved):
Overall agree with Steve's comments which are now addressed so I guess this is good.

PR #340, opened 2021-03-26
Search school URN by exact match

Morpheu5:
This adds an exact match method to ISearchProvider so that we search for an exact match for a given needle in a given field. This helps with duplicate schools when searching by URN prefix.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (approved):
Still needs testing, but looks sensible!

PR #344, opened 2021-04-26
Allow additional group managers to remove themselves from a group

jsharkey13:
This does not notify anyone of the change, nor does it remove the permissions of the manager (in case they manager other groups with the students and consistent with the behaviour of leaving a group or being removed by the owner).

Could someone just double-check the logic of the permissions check, to make sure I am doing the right thing?


Morpheu5 (approved):
The logic looks fine to me.

PR #347, opened 2021-05-20
Catch DuplicateAccountException when signing up

bh412:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (commented):
We used to do:
```
    return new SegueErrorResponse(Status.BAD_REQUEST, "An account already exists with the e-mail address specified.").toResponse();
```
without the log statement. But maybe an info log wouldn't be a bad move.
I think `String.format("Duplicate account registration attempt for (%s)", email)` would be more helpful an internal log message.
I think the old error response message was better than the new one, and will at least mean the tests pass again.

bh412:
I haven't changed the error message from what it used to be, we just re-defined it here, with the 'return new...' you have above, when the DuplicateAccountException is thrown it sets the message to `"An account with that e-mail address already exists."` I'll alter the internal log to your suggestion

jsharkey13 (approved):
Yes, I should have read more carefully and that error message is fine as is. The log looks much better now, and sorry for saying "info" when I meant "warn"! 👍

PR #348, opened 2021-05-26
Add email type logging to customEmail sending

bh412:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


bh412:
If merged I'll update the logged data part of the wiki to include `numberFiltered` and `type`

jsharkey13 (approved):
This is fine! If I was being (needlessly!) pedantic, I might have moved the `.build()` to its own line so that the diff would be cleaner in future if it gets modified again. But it's not necessary!

I also wish I'd used `emailType` in the other log event when I added that, but we should stay consistent.

PR #350, opened 2021-05-27
Add recent questions and oldest incomplete questions to progress stats

jsharkey13:
To keep the response size manageable, only include the most recent 5 questions and oldest 5 incomplete questions. Use a FIFO queue to keep the most recent 5 questions, since this is more efficient than a list.
Doing this assumes the input Map is sorted, which it was not despite the results being returned from Postgres in order. LinkedHashMaps offer this order without much of an overhead and required minimal code change!

I also implemented my 4 year old TODO and removed the need for a local GameManager by using the static method and the existing DTOs, which should also be a performance boost by not loading everything from ElasticSearch twice.


jsharkey13:
I'm pretty pleased with how simple this was! 🎉 

And after a brief test with some high-use users I think that there won't be much of an impact so I added in some warning logging (and the old code logged a less helpful error for each missing question anyway that this now replaces with a slightly more accurate warning).

bh412 (commented):
The code style seems fine and the changes sensible, will test with some accounts before approving

bh412 (approved):
Works as expected, recent questions in the correct order

PR #354, opened 2021-06-16
Improve some of the error messages given by the quiz facade

Morpheu5:
Does what it says on the tin, does not affect functionality.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -304,7 +304,7 @@ public class QuizFacade extends AbstractIsaacFacade {
                 .cacheControl(getCacheControl(NUMBER_SECONDS_IN_ONE_HOUR, false)).tag(etag).build();
         } catch (ContentManagerException e) {
             log.error("Content error whilst previewing a quiz", e);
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -836,7 +836,7 @@ public final Response abandonQuizAttempt(@Context final HttpServletRequest httpS
             QuizAttemptDTO quizAttempt = this.quizAttemptManager.getById(quizAttemptId);
```
jsharkey13, line 836: Not quite releated to this PR, but what happens here is there is no attempt with that ID? Ideally we'd give the same error message as for the user ID mismatch to hide the internal details, does this throw the exception that's caught with an unhelpful error message below? Because otherwise there's an NPE if `quizAttempt` is null?

This probably applies more widely than this one, it's just it took me reading this far to think of it.

Morpheu5: This one throws a `SegueDatabaseException` so it's caught by the appropriate clause.

jsharkey13 (approved):
I still think the exceptions need better handling and phrasing; a quiz attempt not being found doesn't need a message about a SegueDatabaseException returned to the user.
But these changes themselves are fine, so maybe that comes later!

PR #355, opened 2021-06-18
Add audience and display to DOs and DTOs

mlt47:
Registered a new Orika converter to match the deeply nested audience data structure. Orika correctly uses this converter when dealing with the audience field.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
@@ -379,27 +379,13 @@ public class ContentMapper {
             log.info("Creating instance of content auto mapper.");
             MapperFactory mapperFactory = new DefaultMapperFactory.Builder().build();
 
-            ContentBaseOrikaConverter contentConverter = new ContentBaseOrikaConverter(this);
-            ChoiceOrikaConverter choiceConverter = new ChoiceOrikaConverter();
-            ItemOrikaConverter itemConverter = new ItemOrikaConverter();
-
-            QuestionValidationResponseOrikaConverter questionValidationResponseConverter 
-                = new QuestionValidationResponseOrikaConverter();
-
-            AnonymousUserQuestionAttemptsOrikaConverter anonymousUserOrikaConverter 
-                = new AnonymousUserQuestionAttemptsOrikaConverter();
-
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
```diff
@@ -379,27 +379,13 @@ public MapperFacade getAutoMapper() {
...
-            converterFactory.registerConverter("anonymousUserAttemptsToDTOConverter", anonymousUserOrikaConverter);
-
-            // special rules
-//            mapperFactory.classMap(AnonymousUser.class, AnonymousUserDTO.class).fieldMap("temporaryQuestionAttempts")
-//                    .converter("anonymousUserAttemptsToDTOConverter").add().byDefault().register();
+            converterFactory.registerConverter(new ContentBaseOrikaConverter(this));
+            converterFactory.registerConverter(new ChoiceOrikaConverter());
+            converterFactory.registerConverter(new ItemOrikaConverter());
+            converterFactory.registerConverter(new QuestionValidationResponseOrikaConverter());
+            converterFactory.registerConverter(new AnonymousUserQuestionAttemptsOrikaConverter());
```
jsharkey13, line 387: Does it matter that you removed the name/first argument string of this registration?

jsharkey13 (approved):
One minor question, since I don't have introspection on here I can't easily check the difference between the changed method calls!

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
mlt47: Good question. The first argument is the converter ID if you need to refer back to the converter again later.

The commented-out lines 401 and 402 used that ID to try and apply the particular converter for a particular field (following the documentation from https://orika-mapper.github.io/orika-docs/converters.html under the heading "Registering Converters at a Field Level").
As those lines were commented out and nowhere else in the repo uses that ID I thought I might as well reduce the number of lines and neaten it up a bit.

Having an ID does not affect the converter's likelihood of being selected as the appropriate converter for a field, the docs say: `Converters are ordered so that the more specific (in terms of class hierarchy) are tested before the more generic.` so I'm confident it won't change the mapping behaviour.

PR #357, opened 2021-06-22
Tell elasticsearch to treat the audience field as a nested object field

mlt47:
---

**Pull Request Check List**
- ~Security - Injection - everything run by an interpreter (SQL, OS...) is either validated or escaped~
- ~Security - Data Exposure - PII is not stored or sent unencrypted~
- ~Security - Data Exposure - Test any altered or created endpoints using [swagger](http://localhost:8080/isaac-api/api-docs/)~
- ~Security - Access Control - Check authorisation on every new endpoint~
- [ ] Peer-Reviewed

This does not affect existing queries.

The "nested object"-ness can be queried with something like this:
```
curl -X GET "localhost:9200/{SEARCH_INDEX_SHA}_content/_search?pretty" -H 'Content-Type: application/json' -d'
{
  "query": {
    "nested": {
      "path": "audience",
      "query": {
        "bool": {
          "must": [
            { "match": { "audience.stage.keyword": "A Level" }},
            { "match": { "audience.difficulty.keyword":  "Practice 1" }}
          ]
        }
      }
    }
  }
}
'
```

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/etl/ElasticSearchIndexer.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/etl/ElasticSearchIndexer.java
@@ -28,6 +28,7 @@ import uk.ac.cam.cl.dtg.segue.search.ElasticSearchProvider;
 import uk.ac.cam.cl.dtg.segue.search.SegueSearchException;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


bh412 (approved):
Changes appear sensible and match the suggested implementation

PR #361, opened 2021-07-02
Upgrade Jetty to 9.4

jsharkey13:
Please see the commit messages for a detailed breakdown of the necessary changes made. Note the changes to cookie format necessitated by the upgrade.

On reflection, I think we can actually cope by ignoring the node naming business entirely; it is no different to the current session stuff on multiple APIs, so we can just ignore the `node0` parts and I think it will all still work just fine.

### DIFF
--- a/Dockerfile-api
+++ b/Dockerfile-api
@@ -6,7 +6,7 @@ COPY . /isaac-api
 RUN mvn package -Dmaven.test.skip=true
 
 # create clean jetty docker container
-FROM jetty:9.3.29-jre8 as server
+FROM jetty:9.4.42-jre8 as server
 USER root
 COPY --from=target /isaac-api/target/isaac-api.war /var/lib/jetty/webapps/isaac-api.war
 RUN chmod 755 /var/lib/jetty/webapps/*
--- a/Dockerfile-etl
+++ b/Dockerfile-etl
@@ -6,7 +6,7 @@ COPY . /isaac-api
 RUN mvn package -Dmaven.test.skip=true -P etl
 
 # create clean jetty docker container
-FROM jetty:9.3.29-jre8 as server
+FROM jetty:9.4.42-jre8 as server
 USER root
 COPY --from=target /isaac-api/target/isaac-api.war /var/lib/jetty/webapps/isaac-api.war
 RUN chmod 755 /var/lib/jetty/webapps/*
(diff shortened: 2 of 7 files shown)


chrisjpurdy (approved):
Seems fine to me - I was unsure about `DATE_SIGNED` being deleted but I guess it isn't used anywhere otherwise the CI tests wouldn't pass?

sacummins:
This looks good to me and is a reasonable way of making the cookie compliant too. It would be good to chat about what is required to bring jetty completely up to date and whether I should spend a little time looking into this.

mlt47:
@jsharkey13, we talked in person about not following the JWT format because the algorithm and token type header was unnecessary and that Auth0 said there had been attacks involving deliberately specifying the incorrect algorithm.
I was thinking about it the other day and it might be useful if in the future we were needing to change the algorithm without downtime and therefore support two temporarily. We would then have to tie the implementation down to only use the one or two acceptable algorithms to avoid attacks but I would expect us to do that anyway. What do you think?

jsharkey13:
@chrisjpurdy Since Java is compiled, so long as you build ETL and the main API and run the tests, you can be sure it's not needed. IntelliJ said it was unused, and I happened to know that I had stopped using it.

@sacummins Jetty 10 and above need Java 11, which we have avoided because there's the licensing issues for newer Java versions and their short support lifetimes. We use OpenJDK for the Docker images, but there is no JDK8 build for Windows any more and so development is becoming trickier. We'd need to upgrade Java versions (probably to 11, which is LTS) if it was sensible and feasible.

@mlt47 I mean if Steve is okay with moving to a more JWT-style cookie `[base64 header JSON].[base64 body JSON].[base64 HMAC]` then I don't have objections, but I cannot forsee the cookie signing algorithm changing ever; HMAC-SHA256 is the only one anyone recommends any more unless you need third parties to be able to verify them without creating new ones (then you need asymmetric RSA/ECDSA so others can have the public key). 
The only thing likely to change is the body JSON format which isn't supposed to be specified in the header properties (and dealing with changing that would be the same as dealing with it in the present format) - so I don't think the JWT format gains anything and it adds a lot of complexity too. We don't need interop with anyone else, and without rewriting to use a JWT library entirely, there's no gain I can see to a standardised format.

sacummins (approved):

PR #362, opened 2021-07-05
Allow multiple boolean search clauses to be matched against the same field

mlt47:
To get the question filter to match questions with "statics" or "dynamics" and  of type "problem_solving" or "books" we need two separate boolean clauses on the "tags" field.
To achieve this I created a BooleanSearchClause record class (rather than including a tuples library) and told ElasticSearchProvider how to construct a bool match query from a list of these records.
That method is a lot more straightforward now and even the nested queries can be simplified by only treating them differently just before adding the clause to the master query.
I then followed the compiler's complaints to update the fieldsToMatch data structure to the new list of BooleanSearchClauses.

It would have been slightly nicer to review if I would have remembered to commit between the refactoring and the fixing of generateBoolMatchQuery, sorry!

I have tested our major routes to search content quite thoroughly and as the logic is now easier to comprehend I'm confident it is correct.
During the transforming of the clause data structure, the only section of work that was not checked by the compiler was making sure the correct boolean operator was copied over - I have manually double checked those and they are fine.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
@@ -27,8 +27,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.Constants;
-import uk.ac.cam.cl.dtg.isaac.api.Constants.GameboardItemState;
-import uk.ac.cam.cl.dtg.isaac.api.Constants.GameboardState;
 import uk.ac.cam.cl.dtg.isaac.dao.GameboardPersistenceManager;
 import uk.ac.cam.cl.dtg.isaac.dos.GameboardCreationMethod;
 import uk.ac.cam.cl.dtg.isaac.dos.IsaacQuestionPage;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -1210,17 +1191,18 @@ private IsaacWildcard getWildCardById(final String id) throws ContentManagerExce
...
-        // Filter on content tags
-        List<String> tagAnds = Lists.newArrayList();
-        List<String> tagOrs = Lists.newArrayList();
+        List<IContentManager.BooleanSearchClause> fieldsToMatch = Lists.newArrayList();
 
         // handle question categories
         if (null != gameFilter.getQuestionCategories()) {
-            tagAnds.addAll(gameFilter.getQuestionCategories());
+            fieldsToMatch.add(new IContentManager.BooleanSearchClause(
+                    TAGS_FIELDNAME, BooleanOperator.OR, gameFilter.getQuestionCategories()));
```
mlt47, line 1199: Intentionally switch from AND to OR here as one of our filter queries now search for `problem_solving` or `book` questions

src/main/java/uk/ac/cam/cl/dtg/segue/api/Constants.java
```diff
@@ -422,8 +420,6 @@ public PGInterval getPGInterval() {
     public static final String DIFFICULTY_FIELDNAME = "audience.difficulty";
     public static final String EXAM_BOARD_FIELDNAME = "audience.examBoard";
     public static final Set<String> NESTED_FIELDS =
-            // NOTE: if you are adding a nested field name that does not have the nested path audience, you will
-            // need to alter (at least) generateBoolMatchQuery to be a little smarter - possible, though
```
mlt47, line 426: This has now been fixed in generateBoolMatchQuery

PR #365, opened 2021-07-13
Add support for regex string match questions

chrisjpurdy:
Have added relevant DOs, DTOs and a new choice subclass. Most is just copy-pasted from string match question relevant files and nothing critical has changed as far as I can tell.

I also added github.com/google/re2j to the Maven dependency list, to replace the standard Java regex library (*this is removed in the most recent commits*)

---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13:
We discussed this, and decided not to use the Google library until we have some real-world test cases since it is much slower than the Java library for many cases that might be common for our usage (i.e. whilst the library's worst-case runtime is much much much better than the standard library, the common-case runtime may well be noticably worse). We've left the commit in the history so it is easy to go back to or find again later on.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacRegexStringMatchValidator.java
```diff
@@ -0,0 +1,131 @@
...
+                                   final Boolean multiLineRegex) {
+
+        if (null == trustedRegexPattern || null == userValue) {
+            return false;
+        }
+
+        // The pattern is case sensitive and single line by default
+        Pattern answerPattern = Pattern.compile(trustedRegexPattern,
+                (null != caseInsensitive && caseInsensitive ? Pattern.CASE_INSENSITIVE : 0)
+                        | (null != multiLineRegex && multiLineRegex ? Pattern.MULTILINE : 0));
```
jsharkey13, line 127: This bitwise OR hiding in between the flags is actually the way to combine multiple regex flags and not part of the logic of the inline conditionals. Our old CheckStyle was dead against inline conditionals, but I think these are probably fine.
Adding to the comment above that "the flags are combined with a bitwise or" could be helpful though.

src/main/java/uk/ac/cam/cl/dtg/segue/dos/content/RegexPattern.java
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright 2017 James Sharkey
```
jsharkey13, line 2: To fix.

src/main/java/uk/ac/cam/cl/dtg/segue/dto/content/RegexPatternDTO.java
```diff
@@ -0,0 +1,64 @@
+/**
+ * Copyright 2017 James Sharkey
```
jsharkey13, line 2: To fix.

jsharkey13 (commented):
Some fairly pedantic things to fix but no major changes needed. 

I do wonder whether `isaacRegexStringMatchQuestion`, aside from being a mouthfull, might be confused with `isaacStringMatchQuestion` in a way that `isaacRegexQuestion` or `isaacRegexMatchQuestion` wouldn't be? Maybe the others have opinions here? (It's just an internal name, so it's fine whatever in the end).

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacRegexStringMatchQuestionDTO.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package uk.ac.cam.cl.dtg.isaac.dto;
+
+import uk.ac.cam.cl.dtg.isaac.quiz.IsaacRegexStringMatchValidator;
+import uk.ac.cam.cl.dtg.segue.dos.content.JsonContentType;
+import uk.ac.cam.cl.dtg.segue.quiz.ValidatesWith;
+
+/**
+ * DTO for isaacRegexStringMatchQuestion.
```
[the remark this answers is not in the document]
chrisjpurdy: Is this wrong? It's certainly obsolete since it just reiterates the name, but the class name is correct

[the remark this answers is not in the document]
jsharkey13: Oh, yeah, this one is the right class name. The last one is not. If you remove the other one, remove this one. If you fix the other one, leave this and mark this thread resolved. Sorry!

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacRegexStringMatchValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        Validate.notNull(question);
+        Validate.notNull(answer);
+
+        if (!(question instanceof IsaacRegexStringMatchQuestion)) {
+            throw new IllegalArgumentException(String.format(
+                    "This validator only works with Isaac Regex String Match Questions... (%s is not regex string match)",
+                    question.getId()));
+        }
+
+        if (!(answer instanceof StringChoice)) {
```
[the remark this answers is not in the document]
chrisjpurdy: `StringChoice` seems a bit overkill here, as the `caseInsensitive` flag isn't considered for the submitted answer - maybe just a `Choice`? Or `TextChoice` (or similar) that's just a synonym for `Choice` but comes loaded with the assumption that text is being passed (maybe in this sense `StringChoice` is a good choice, even with the extra unneeded attribute).

chrisjpurdy:
Have now fixed the copyright issues and corrected/added to comments and JavaDoc in the most recent commit. 

I'm going to hold off on renaming `IssacRegexStringMatch` to `IssacRegexMatch` for now since I've littered multiple repositories with it.

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacRegexStringMatchQuestionDTO.java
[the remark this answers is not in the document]
chrisjpurdy: No worries!

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacRegexStringMatchValidator.java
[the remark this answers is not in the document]
chrisjpurdy: This approach is used in the `FreeText` question validation - as @jsharkey13 said it's more of an architecture level problem that it doesn't make sense to deal with at the moment.

Morpheu5 (approved):

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ChoiceOrikaConverter.java
```diff
@@ -72,6 +74,8 @@ public ChoiceDTO convertTo(final Choice source, final Type<ChoiceDTO> destinatio
             return super.mapperFacade.map(source, GraphChoiceDTO.class);
         } else if (source instanceof StringChoice) {
             return super.mapperFacade.map(source, StringChoiceDTO.class);
+        } else if (source instanceof RegexPattern) {
+            return super.mapperFacade.map(source, StringChoiceDTO.class);
```
jsharkey13, line 78: This isn't right. Nothing intentionally needs the DTO object, and so it shouldn't even be needed at all. Unfortunately, to serialise the question for the frontend, we do convert it to a DTO and need all the choices to be DTOs too. These are immediately `@JsonIgnore`-d for all question types except `isaacMultiChoiceQuestion`, but the type hierarchy means that they do have to exist in the DTO just to be immediately discarded.

I _think_, since it will never be used, making this map to the `RegexPatternDTO` is the 'right' thing to do here, since a string choice shouldn't work (but sometimes will because both usually only have the `value` field set and so the mapper does manage to coerce it; if a `RegexPattern` existed with `matchWholeString` set, this would likely explode).
But it is annoying to have to have and use a class that serves absolutely no useful purpose :/

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ChoiceOrikaConverter.java
```diff
@@ -105,6 +109,8 @@ public Choice convertFrom(final ChoiceDTO source, final Type<Choice> destination
             return super.mapperFacade.map(source, GraphChoice.class);
         } else if (source instanceof StringChoiceDTO) {
             return super.mapperFacade.map(source, StringChoice.class);
+        } else if (source instanceof RegexPatternDTO) {
+            return super.mapperFacade.map(source, StringChoice.class);
```
jsharkey13, line 113: See my comment above, but this I think must be `RegexPattern` too, for the same reason.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ChoiceOrikaConverter.java
jsharkey13: This shouldn't affect the validator's ability to expect an inbound `StringChoice` because we trust the user's declared `Choice` subtype, and we use the _question type's_ declared validator which is manually set with the annotation.

jsharkey13: The whole system is all such a mess . . . 
😞

jsharkey13 (changes requested):
Just to ensure this doesn't get merged in my absence without being fixed! See my comments about the choice mapping in Orika.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ChoiceOrikaConverter.java
jsharkey13: The fact I missed this twice also hints at the complexity! :/

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ChoiceOrikaConverter.java
chrisjpurdy: Fixed in most recent commit

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ChoiceOrikaConverter.java
chrisjpurdy: Fixed in most recent commit - this was just a typo that I missed.

jsharkey13 (approved):

PR #367, opened 2021-07-14
Item question feedback - added API support for matching subsets of an item choice

chrisjpurdy:
The precedence of matching choices is currently as follows:
1. Correct strict match
2. Incorrect strict match
3. Correct subset match
4. Incorrect subset match

This made sense to me on the grounds that if an incorrect choice is a subset of a correct choice where the user can choose , then that question is badly written, and correct choices should always be matched before incorrect ones. Incorrect subset should always be last since it's quite harsh: essentially "mark this question wrong if any of the following are ticked". (This is no longer the case since the requested changes below)

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidator.java
@@ -15,6 +15,8 @@
  */
 package uk.ac.cam.cl.dtg.isaac.quiz;
 
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import org.apache.commons.lang3.Validate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -88,12 +90,17 @@ public class IsaacItemQuestionValidator implements IValidator {
             if (!allowedItemIds.containsAll(submittedItemIds)) {
                 feedback = new Content("You did not provide a valid answer; it contained unrecognised items");
             }
+            if (submittedItemIds.size() != submittedChoice.getItems().size()) {
+                feedback = new Content("You did not provide a valid answer; it contained duplicate items");
+            }
         }
 
         // STEP 2: If they did, does their answer match a known answer?
 
         if (null == feedback && null != submittedItemIds) {
-            // Sort the choices so that we match incorrect choices last, taking precedence over correct ones.
+            /* Sort the choices so that we match incorrect choices last, taking precedence over correct ones, and
+               so that strict (entire set) match choices take precedence over subset match choices.
+             */
             List<Choice> orderedChoices = getOrderedChoices(itemQuestion.getChoices());
 
             // For all the choices on this question...
@@ -116,12 +123,22 @@ public class IsaacItemQuestionValidator implements IValidator {
                 }
 
                 // ... look for a match to the submitted answer:
-                if (itemChoice.getItems().size() != submittedChoice.getItems().size()) {
-                    continue;
-                }
+
                 Set<String> choiceItemIds = itemChoice.getItems().stream().map(Item::getId).collect(Collectors.toSet());
 
-                if (choiceItemIds.equals(submittedItemIds)) {
+                // Do not allow subset matching by default
+                boolean allowSubsetMatch = (null != itemChoice.isAllowSubsetMatch() && itemChoice.isAllowSubsetMatch());
+
+                /* If the intersection of the submitted and choice ids is equal to the submitted ones, then
+                   this means that:
+                    - submittedItemIds.size() <= choiceItemIds.size()
+                    - All submitted ids are within the set of choice ids
+                 */
+                if (allowSubsetMatch && Sets.intersection(submittedItemIds, choiceItemIds) == submittedItemIds) {
+                    responseCorrect = itemChoice.isCorrect();
+                    feedback = (Content) itemChoice.getExplanation();
+                    break;
+                } else if (choiceItemIds.size() == submittedItemIds.size() && choiceItemIds.equals(submittedItemIds)) {
                     // This is safe from duplication issues, since the list lengths are the same too.
                     responseCorrect = itemChoice.isCorrect();
                     feedback = (Content) itemChoice.getExplanation();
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dos/content/ItemChoice.java
```diff
@@ -42,4 +43,8 @@ public ItemChoice() {
     public void setItems(final List<Item> items) {
         this.items = items;
     }
+
+    public Boolean isAllowSubsetMatch() { return this.allowSubsetMatch; }
+
+    public void setAllowSubsetMatch(final boolean allowSubsetMatch) { this.allowSubsetMatch = allowSubsetMatch; }
```
jsharkey13, line 49: Could you lay this out on three lines? I know IntelliJ will collapse it for you, but it's not consistent with the rest of the codebase (and I don't like it either! :P )

jsharkey13: In fact, do this for all of them in the DO and DTO?

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidatorTest.java
```diff
@@ -97,6 +97,26 @@ public final void isaacItemQuestionValidator_CorrectItems_CorrectResponseShouldB
...
     }
 
+    /*
+       Test that exact correct answers are recognised even if the choice is subset match.
+    */
+    @Test
+    public final void isaacItemQuestionValidator_CorrectItemsSubsetMatchEnabled_CorrectResponseShouldBeReturned() {
+
+        // Enable subset matching on the second choice in the list (the correct one)
+        ((ItemChoice) someItemQuestion.getChoices().get(1)).setAllowSubsetMatch(true);
```
jsharkey13, line 107: This will affect other tests, not just this one, and JUnit makes no promises to the order in which they are run. You should keep the test side effect free, and either initialise a new question at the start that will be useful for subset tests, or make a new question inside this test as you did later on.

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidatorTest.java
```diff
@@ -152,6 +172,122 @@ public final void isaacItemQuestionValidator_CorrectChoicePrecedence_CorrectResp
...
 
+    /*
+       Test that correct subset choices take precedence over incorrect subset ones.
+    */
+    @Test
+    public final void isaacItemQuestionValidator_CorrectChoiceSubsetPrecedence_CorrectResponseShouldBeReturned() {
+
+        // Enable subset matching on both of the choices in the list
+        ((ItemChoice) someItemQuestion.getChoices().get(0)).setAllowSubsetMatch(true);
+        ((ItemChoice) someItemQuestion.getChoices().get(1)).setAllowSubsetMatch(true);
```
jsharkey13, line 183: These two lines also have side effects outside of this test function.

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidatorTest.java
```diff
@@ -152,6 +172,122 @@ public final void isaacItemQuestionValidator_CorrectChoicePrecedence_CorrectResp
...
+
+    /*
+       Test that if the correct choice is a strict subset of the submitted answer, it is not marked as correct,
+       if the correct choice is marked as subset match.
+    */
+    @Test
+    public final void isaacItemQuestionValidator_CorrectChoiceSubsetOfSubmitted_IncorrectResponseShouldBeReturned() {
+
+        // Enable subset matching on the correct choice
+        ((ItemChoice) someItemQuestion.getChoices().get(1)).setAllowSubsetMatch(true);
```
jsharkey13, line 203: Ditto for side effects here.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidator.java
```diff
@@ -133,4 +150,46 @@ public final QuestionValidationResponse validateQuestionResponse(final Question
...
+
+        /* First sort by whether subset matching is allowed or not - 'strict' match
+           item choices will appear before subset match ones.
+           Any Choices that are not ItemChoices will be ordered after 'strict' match
+           item choices.
+           Choices without a null value for allowSubsetMatch are considered 'strict'
+           for the ordering.
+         */
+        orderedChoices.sort((o1, o2) -> {
+            int o1Val = 1, o2Val = 1;
```
jsharkey13, line 165: Can we keep to one statement per line?

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidator.java
```diff
@@ -116,12 +123,22 @@ public final QuestionValidationResponse validateQuestionResponse(final Question
...
-                if (choiceItemIds.equals(submittedItemIds)) {
+                // Do not allow subset matching by default
+                boolean allowSubsetMatch = (null != itemChoice.isAllowSubsetMatch() && itemChoice.isAllowSubsetMatch());
+
+                /* If the intersection of the submitted and choice ids is equal to the submitted ones, then
+                   this means that:
+                    - submittedItemIds.size() <= choiceItemIds.size()
+                    - All submitted ids are within the set of choice ids
+                 */
+                if (allowSubsetMatch && Sets.intersection(submittedItemIds, choiceItemIds).equals(submittedItemIds)) {
```
jsharkey13, line 137: This feels the wrong way round to me?

If the question has options {A, B, C, D} and the correct answer is {A, B, C} and a known wrong answer that allows subset matches is just {D}, then a student answer of {A, B, C, D} should match this known wrong answer since that's what we need to allow to avoid writing every possible combination that includes D if we just want to say "D is definitely wrong".

I think this should be `Sets.intersection(submittedItemIds, choiceItemIds).equals(choiceItemIds)`, i.e. are all of the items from this choice in the submission?

jsharkey13: What you have here (and is demonstrated in your `isaacItemQuestionValidator_SubsetOfCorrectItems_CorrectResponseShouldBeReturned` test) is exactly the opposite of what this is meant to do. If we wanted to allow any subset of the options {A, B, C} to be marked as correct, then in "my" version you'd just add {A}, {B}, and {C} as correct choices that allow subset matches and that would work too. But that's not a case that I can see happening.

jsharkey13 (changes requested):
This feels like my fault, although I thought I explained it and highlighted the potential for the subset being the wrong way around - but your subset is the wrong way around 😅 

See my comment involving {A, B, C, D} for an example, but basically the submitted answer of {A, B, C} should match a known allowSubset choice {A}, but a submitted {A} *should not* match a known choice {A, B, C} with or without allowing subset matching.
This is what I meant when I said "allow subset match" could be confusing :P

Maybe `allowLooseMatch` is better, denoting that if all the things in the choice are in the student answer, it matches?

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacItemQuestionValidatorTest.java
chrisjpurdy: Why does this effect all tests? The @Before on the setUp function makes it run before each test, so `someItemQuestion` is initialised again - I wasn't sure either but I've just checked with an @After what the values of `allowSubsetMatch` are on the choices in `someItemQuestion`, and it doesn't affect the other tests as far as I can see

src/main/java/uk/ac/cam/cl/dtg/segue/dos/content/ItemChoice.java
chrisjpurdy: IntelliJ collapses the three lines into one visually, so the rest of the codebase looked like that to me (and personally I like this way of doing it anyway ':D) - I'll change it so that it's consistent.

jsharkey13 (approved):
🎉 

I added one last test, to push line coverage to 100%, that just checks that duplicate items (which should never happen in the wild) prevent an answer being marked as correct.

PR #368, opened 2021-07-16
Improved significant figures feedback

jsharkey13:
Instead of the old blanket sig fig response, now have a tag for too few sig figs which can be customised in the frontend if desired, but do not tag to warn about too many significant figures unless matched against a correct choice.

This extends #366 and so that should be merged first. Only the last commit is relevant to this specific change.


bh412 (approved):
Returns the right thing and the tests work. Given @mlt47 's feedback to the other PR, maybe check the comments?

jsharkey13:
I am happy with the comments as-is! 
This will need some frontend work to adjust the messaging and use the new tags, but it is safe to merge.

PR #370, opened 2021-07-20
Restore old desired behaviour to Orika Custom and Bidirectional mappers

jsharkey13:
After updating to v1.5.x, the default behaviour of mappers became much stricter and did not correctly map all of our polymorphic subclasses in all cases.
This issue is the one described in orika-mapper/orika#171, and the fix is to override canConvert for CustomConverter and BidirectionalConverter with the old code as suggested in that issue.

I can't be sure that _all_ of these Converters need to be the polymorphic supporting version - but Item, Choice and QuestionValidationResponse definitely need to be for questions to continue functioning.
I also must have tested with a question type that did still work (namely multiple choice that use Choice and not a subclass) and only _some_ cases of our use of Orika were broken by the changes :/

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/AbstractPolymorphicBidirectionalConverter.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/AbstractPolymorphicBidirectionalConverter.java
@@ -1,0 +1,34 @@
+package uk.ac.cam.cl.dtg.segue.dao.content;
+
+
+import ma.glasnost.orika.converter.BidirectionalConverter;
+import ma.glasnost.orika.metadata.Type;
+
+/**
+ * An Orika BidirectionalConverter that implements the old, polymorphic, canConvert check.
+ *
+ * @param <S> - source type
+ * @param <D> - destination type
+ */
+public abstract class AbstractPolymorphicBidirectionalConverter<S, D> extends BidirectionalConverter<S, D> {
+
+    @Override
+    public boolean canConvert(Type<?> sourceType, Type<?> destinationType) {
+        /* The behaviour of canConvert changed in Orika v1.5.0 to only convert exact class matches,
+           to fix an issue with converters acting too loosely on pairs of classes they were not
+           meant to convert.
+           The crux of the change was swapping from sourceType.isAssignableFrom() to sourceType.equals(),
+           which for our use case prevents the clever polymorphism we do with subtypes working in converters.
+
+           Since this is a bidirectional converter, we must check whether the reverse conversion is
+           possible too. It is not immediately obvious what the correct reverse check must be, and so this
+           check has been copied from the old Orika code.
+
+           See the commit that changed this behaviour here:
+           https://github.com/orika-mapper/orika/commit/554396579c96b3356c3c31ceb2e236cba0ffbaba
+         */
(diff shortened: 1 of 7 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/AbstractPolymorphicBidirectionalConverter.java
```diff
@@ -0,0 +1,34 @@
...
+           Since this is a bidirectional converter, we must check whether the reverse conversion is
+           possible too. It is not immediately obvious what the correct reverse check must be, and so this
+           check has been copied from the old Orika code.
+
+           See the commit that changed this behaviour here:
+           https://github.com/orika-mapper/orika/commit/554396579c96b3356c3c31ceb2e236cba0ffbaba
+         */
+        boolean forwardConvertable = this.sourceType.isAssignableFrom(sourceType) && this.destinationType.equals(destinationType);
+        boolean reverseConvertable = this.destinationType.isAssignableFrom(sourceType) && this.sourceType.equals(destinationType);
+        return forwardConvertable || reverseConvertable;
```
bh412, line 32: Just checking that this should be an or not and? Bidirectional to me would mean both ways (I don't claim to understand Orika so your or will most likely be correct)

bh412 (approved):
Appears to do what we need - Just one comment checking an operator, though I'm sure your implementation is correct

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/AbstractPolymorphicBidirectionalConverter.java
jsharkey13: Ah, yes, `Bidirectional` here means it will work forwards, or work in reverse; i.e. it can convert a `Choice` to a `ChoiceDTO` or the opposite way and convert back again.
The converter has `this.sourceType=Choice` and `this.destinationType=ChoiceDTO`, and we always call the mapper to convert to the parent class of either `Choice` or `ChoiceDTO` and our converter then picks the correct subclass.

`a.isAssignableFrom(b)` says "is a either the same class or a parent class of b".

If we have a specific example from a DO to a DTO where the `sourceType=StringChoice` and the `destinationType=ChoiceDTO`, then we do:

```
   forward = Choice.isAssignableFrom(StringChoice) && ChoiceDTO.equals(ChoiceDTO)
   reverse = ChoiceDTO.isAssignableFrom(StringChoice) && Choice.equals(ChoiceDTO)
```
and it is clear that the reverse conversion here cannot work, since the DTO and DO classes are not related at all. The forward case does work, since `Choice` is a parent class of `StringChoice` and this is what our converter is designed to do. But if we wanted to go the opposite way from DTO to DO, rather than needing a second converter, our bidirectional one can be used backwards:

`sourceType=StringChoiceDTO` and `destinationType=Choice` gives:
```
   forward = Choice.isAssignableFrom(StringChoiceDTO) && ChoiceDTO.equals(Choice)
   reverse = ChoiceDTO.isAssignableFrom(StringChoiceDTO) && Choice.equals(Choice)
```
and here the forward case is false, but the reverse case is now true since our mapper can do the backwards conversion.

So it must be OR, and not AND (since it's not possible to go both ways _at the same time_). It's also copied directly from the code in that GitHub link, if you scroll to see `BidirectionalConverter`.

PR #371, opened 2021-07-23
Add a displayUnit to numeric questions which overrides requiresUnits

jsharkey13:
The aim is that the displayUnit gets shown to the user and so we do not need to validate the units since they are not asked for. Guarding against cases where units have been specified anyway makes up most of the changes.
Two new content errors are added to highlight these issues.

I also removed sig figs from the DTO, since they were JsonIgnored and they were not being used and it simplified the file a lot.

This will need frontend and editor work to (a) use the display units in the frontend and stop showing the dropdown in that case, and (b) allow it to be specified in the content editor whilst trying to ensure requiresUnits is not set at the same time, and that choices provided do not have units.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacNumericQuestionDTO.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacNumericQuestionDTO.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2014 Stephen Cummins
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -36,9 +36,8 @@ import static uk.ac.cam.cl.dtg.isaac.api.Constants.NUMERIC_QUESTION_DEFAULT_SIGN
 @ValidatesWith(IsaacNumericValidator.class)
 public class IsaacNumericQuestionDTO extends IsaacQuestionBaseDTO {
     private Boolean requireUnits;
-    private Integer significantFiguresMin;
-    private Integer significantFiguresMax;
     private List<String> availableUnits;
+    private String displayUnit;
 
     /**
      * Gets the requireUnits.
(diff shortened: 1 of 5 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacNumericQuestionDTO.java
```diff
@@ -36,9 +36,8 @@
 @ValidatesWith(IsaacNumericValidator.class)
 public class IsaacNumericQuestionDTO extends IsaacQuestionBaseDTO {
     private Boolean requireUnits;
-    private Integer significantFiguresMin;
-    private Integer significantFiguresMax;
```
chrisjpurdy, line 40: Why can you delete these here? Are sig-figs only used in the API?

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidatorTest.java
```diff
@@ -605,6 +606,34 @@ public final void isaacNumericValidator_WrongChoiceType_ExceptionShouldBeThrown(
...
+    */
+    @Test
+    public final void isaacNumericValidator_TestInconsistentDisplayUnitsOverride_UnitsIgnored() {
+        // Set up the question object:
+        IsaacNumericQuestion someNumericQuestion = new IsaacNumericQuestion();
+        someNumericQuestion.setRequireUnits(true);
+        someNumericQuestion.setDisplayUnit("SOME-FAKE-UNIT");
+
+        List<Choice> answerList = Lists.newArrayList();
+        Quantity someCorrectAnswer = new Quantity(correctIntegerAnswer, "SOME-FAKE-UNIT");
```
chrisjpurdy, line 620: Doesn't the fact that you've set the display unit mean that you don't need to specify the correct unit in the Quantity answer here? Just checking to make sure I've understood the change.

chrisjpurdy (commented):
The logic checks out, and all of the changes make sense. I've left a couple of comments with general questions.

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacNumericQuestionDTO.java
jsharkey13: Yes, and more importantly they were `@JsonIgnore`-d to ensure they never got sent to the frontend since we do not want them to be there! I decided to remove them to check it didn't break anything, and since it didn't I left it in.

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidatorTest.java
jsharkey13: Yes. This tests that even though the unit is incorrect, the response is still correct because it is ignored. But actually, now you make me consider it again, we should give an error if you provide a unit when you shouldn't have done? And not just silently ignore it? An error is probably more consistent with everything else!

bh412: I think this test should stay as is, it's testing that a unit-less student answer is marked correct despite the fact a question specifies displeyUnit and requireUnits. You did throw a content error if requireUnits and displayUnit are both set stating requireUnits would be overwritten, so I think that's error enough?

bh412 (approved):
Looks good - I think the tests are sensible, explained in my comment

src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidatorTest.java
jsharkey13: Ah, yes, I think I muddled what I wanted to do and what I actually did! I agree with Ben that this is fine as-is!

PR #379, opened 2021-08-24
Feature/record gameboard contexts

mlt47:
---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
@@ -304,8 +304,8 @@ public class GameboardsFacade extends AbstractIsaacFacade {
                 return new SegueErrorResponse(Status.NOT_FOUND, "Gameboard id not a valid FastTrack gameboard id.")
                         .toResponse();
             }
-
             AbstractSegueUserDTO currentUser = this.userManager.getCurrentUser(httpServletRequest);
+            GameboardDTO gameboard = this.gameManager.getGameboard(gameboardId);
 
             Map<String, Map<String, List<QuestionValidationResponse>>> userQuestionAttempts =
                     this.questionManager.getQuestionAttemptsByUser(currentUser);
@@ -314,13 +314,13 @@ public class GameboardsFacade extends AbstractIsaacFacade {
             if (upperQuestionId.isEmpty()) {
                 List<FASTTRACK_LEVEL> upperAndLower = Arrays.asList(FASTTRACK_LEVEL.ft_upper, FASTTRACK_LEVEL.ft_lower);
                 conceptQuestionsProgress.addAll(fastTrackManger.getConceptProgress(
-                        gameboardId, upperAndLower, currentConceptTitle, userQuestionAttempts));
+                        gameboard, upperAndLower, currentConceptTitle, userQuestionAttempts));
             } else {
                 String upperConceptTitle = fastTrackManger.getConceptFromQuestionId(upperQuestionId);
                 conceptQuestionsProgress.addAll(fastTrackManger.getConceptProgress(
-                        gameboardId, Collections.singletonList(FASTTRACK_LEVEL.ft_upper), upperConceptTitle, userQuestionAttempts));
+                        gameboard, Collections.singletonList(FASTTRACK_LEVEL.ft_upper), upperConceptTitle, userQuestionAttempts));
                 conceptQuestionsProgress.addAll(fastTrackManger.getConceptProgress(
-                        gameboardId, Collections.singletonList(FASTTRACK_LEVEL.ft_lower), currentConceptTitle, userQuestionAttempts));
+                        gameboard, Collections.singletonList(FASTTRACK_LEVEL.ft_lower), currentConceptTitle, userQuestionAttempts));
             }
 
             return Response.ok(conceptQuestionsProgress).build();
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java
@@ -47,6 +47,7 @@ import uk.ac.cam.cl.dtg.segue.dto.ResultsWrapper;
 import uk.ac.cam.cl.dtg.segue.dto.content.ContentDTO;
 import uk.ac.cam.cl.dtg.segue.dto.users.RegisteredUserDTO;
 
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.sql.Array;
 import java.sql.Connection;
@@ -63,6 +64,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -186,11 +188,12 @@ public class GameboardPersistenceManager {
 		GameboardDO gameboardToSave = mapper.map(gameboard, GameboardDO.class);
 		// the mapping operation won't work for the list so we should just
 		// create a new one.
-		gameboardToSave.setQuestions(new ArrayList<String>());
+		gameboardToSave.setContents(Lists.newArrayList());
 
 		// Map each question into an IsaacQuestionInfo object
-		for (GameboardItem c : gameboard.getQuestions()) {
-			gameboardToSave.getQuestions().add(c.getId());
+		for (GameboardItem c : gameboard.getContents()) {
+			gameboardToSave.getContents().add(
+			        new GameboardContentDescriptor(c.getId(), c.getContentType(), c.getCreationContext()));
 		}
 
 		// This operation may not be atomic due to underlying DB. Gameboard create first then link to user view second.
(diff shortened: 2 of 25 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/GameboardsFacade.java
```diff
@@ -314,13 +314,13 @@ public final Response getFastTrackConceptFromHistory(@Context final Request requ
...
-                        gameboardId, upperAndLower, currentConceptTitle, userQuestionAttempts));
+                        gameboard, upperAndLower, currentConceptTitle, userQuestionAttempts));
             } else {
                 String upperConceptTitle = fastTrackManger.getConceptFromQuestionId(upperQuestionId);
                 conceptQuestionsProgress.addAll(fastTrackManger.getConceptProgress(
-                        gameboardId, Collections.singletonList(FASTTRACK_LEVEL.ft_upper), upperConceptTitle, userQuestionAttempts));
+                        gameboard, Collections.singletonList(FASTTRACK_LEVEL.ft_upper), upperConceptTitle, userQuestionAttempts));
                 conceptQuestionsProgress.addAll(fastTrackManger.getConceptProgress(
-                        gameboardId, Collections.singletonList(FASTTRACK_LEVEL.ft_lower), currentConceptTitle, userQuestionAttempts));
+                        gameboard, Collections.singletonList(FASTTRACK_LEVEL.ft_lower), currentConceptTitle, userQuestionAttempts));
```
mlt47, line 323: We now pass the gameboard as that allows us to use the gameboard's filter for creating an appropriate context (although this is not important for FastTrack at the moment).

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java
```diff
@@ -557,24 +566,34 @@ private GameboardDO saveGameboard(final GameboardDO gameboardToSave)
...
+                            + " ?::text::jsonb, ?);");
+
+            List<String> contentsJsonb = Lists.newArrayList();
+            for (GameboardContentDescriptor content: gameboardToSave.getContents()) {
+                contentsJsonb.add(objectMapper.writeValueAsString(content));
+            }
+            Array contents = conn.createArrayOf("jsonb", contentsJsonb.toArray());
+            // TODO MT AUDIENCE_CONTEXT CLEAN UP - Temporarily also store questions in DB for the benefit of old APIs
+            Array questionIds = conn.createArrayOf("varchar",
+                    gameboardToSave.getContents().stream().map(GameboardContentDescriptor::getId).toArray());
```
mlt47, line 581: This is to support the old API while it is still about but can then be removed along with the column

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/GameboardDO.java
```diff
@@ -277,10 +258,27 @@ public final void setTags(final Set<String> tags) {
...
+                ", title='" + title + '\'' +
+                ", contents=" + contents +
+                ", wildCard=" + wildCard +
+                ", wildCardPosition=" + wildCardPosition +
+                ", creationDate=" + creationDate +
+                ", gameFilter=" + gameFilter +
+                ", ownerUserId=" + ownerUserId +
+                ", creationMethod=" + creationMethod +
+                ", tags=" + tags +
+                ']';
```
mlt47, line 282: These were auto generated by IntelliJ, do we need to keep the same format? We don't do that much analysis of logs which is where consistency would be important. The main difference is the quoting of strings.

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/GameboardItem.java
```diff
@@ -57,6 +62,21 @@
...
     public GameboardItem() {}
 
+    /**
+     * Static factory method for constructing a minimal Gameboard Item from a GameboardContentDescriptor.
+     * Augmentation by mapping the content object to the instance will need to be done separately.
+     * @param contentDescriptor
+     *          a content descriptor to populate the gameboard item's fields.
+     * @return A Minimal gameboard item containing all of the information from the content descriptor.
+     */
+    public static GameboardItem fromContentDescriptorLight(final GameboardContentDescriptor contentDescriptor) {
```
mlt47, line 72: The "Light" here is to indicate that it is not a very full GameboardItem, open to better naming suggestions

src/main/java/uk/ac/cam/cl/dtg/segue/dos/AudienceContext.java
```diff
@@ -0,0 +1,68 @@
...
+                setDifficulty(gameFilter.getDifficulties().stream().map(Difficulty::valueOf).collect(Collectors.toList()));
+            }
+        }};
+    }
+
+    public AudienceContext() {}
+
+    public List<Stage> getStage() {
+        return stage;
+    }
```
mlt47, line 39: A silly decision while creating this structure in the editor was that the field name is singular but the value is a collection...
I am going to figure out if I can do a mass migration of the content to using the pluralised form of the field names before we start saving this in gameboard contexts everywhere but it will require switching off the beta feature for any users with it enabled on live (I believe only staff after a recent clean up) and for content editors to refresh their browsers so that they don't create structures in the old format... It gets a lot harder to do that if I don't manage to fix this before the release.

src/main/resources/db_scripts/postgres-rutherford-create-script.sql
```diff
@@ -132,6 +132,7 @@ CREATE TABLE public.gameboards (
     id character varying NOT NULL,
     title text,
     questions character varying[],
+    contents jsonb[] DEFAULT array[]::jsonb[] NOT NULL,
```
mlt47, line 135: I now need a migration script that updates the contents field for all gameboards from its question field. If you want to have a go at this, James, feel free! I'll do it otherwise

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/GameboardDO.java
jsharkey13: No, they are used for caching headers mainly, but so long as they are consistent for a given gameboard for a given API version it will be fine (we don't do cross-version API requests, so only cookies have to be consistent across versions). They rarely get used in logs. One recurring problem is that this style risks generating things like `{HashMap@10952}` for non-string things, which means the caching headers _always_ fail even though the contents of the map/object is identical (because the memory address won't be the same twice, even if the board is the same). I don't think we should be _too_ worried here right now

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/GameboardItem.java
jsharkey13: We use "Lightweight" elsewhere for the same idea, so there's probably nothing wrong with this!

jsharkey13: But `getLightweightItemFromContentDescriptor` or suchlike might be slightly clearer?

src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgUsers.java
```diff
@@ -704,10 +704,16 @@ private RegisteredUser createUser(final RegisteredUser userToCreate) throws Segu
                             + "date_of_birth, gender, registration_date, school_id, "
                             + "school_other, exam_board, last_updated, email_verification_status, "
                             + "last_seen, email_verification_token, email_to_verify, "
-                            + " registered_contexts, registered_context_last_confirmed) "
-                    + "VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);",
+                            + "registered_contexts, registered_context_last_confirmed) "
+                    + "VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?::text::jsonb[], ?);",
```
jsharkey13, line 708: Why is this JSONB[] -> TEXT -> JSONB[] cast necessary? JDBC seems to set it as a JSONB[] value on L733, so this looks like it ought to be an unnecessary overhead?

src/main/java/uk/ac/cam/cl/dtg/segue/dos/AudienceContext.java
jsharkey13: 😳

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/GameboardContentDescriptor.java
```diff
@@ -0,0 +1,52 @@
...
+        return context;
+    }
+
+    public void setContext(AudienceContext context) {
+        this.context = context;
+    }
+
+    @Override
+    public String toString() {
+        return "GameboardContentItem [" +
```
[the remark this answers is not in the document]
mlt47: Good spot

src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgUsers.java
mlt47: Hmm yes you're right. I remember trying it both ways in case it made a difference but must have not moved back. Like you say we explicitly create the array of the correct type so the prepared statement shouldn't need casting.

PR #381, opened 2021-09-13
Call synchroniseChangedUsers every 12 hours

bh412:
---

**Pull Request Check List**
- [ ] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/SyncMailjetUsersJob.java
```diff
@@ -0,0 +1,51 @@
...
+        externalAccountManager = Guice.createInjector(new SegueGuiceConfigurationModule()).getInstance(IExternalAccountManager.class);
+    }
+
+    @Override
+    public void execute(JobExecutionContext context) throws JobExecutionException {
+        try {
+            externalAccountManager.synchroniseChangedUsers();
+            log.info("Success: synchronised users");
+        } catch (ExternalAccountSynchronisationException e) {
+            log.error("Failed to synchronise users");
```
jsharkey13, line 47: Could this log the error that occurred? I do wonder if this will be obvious enough to catch the errors, if any occur, or if we need to think about emailing errors to ourselves like we do with the misuse monitors?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -914,7 +916,20 @@ private static SegueJobService getSegueJobService(final PostgresSqlDb database)
...
+                    "JavaJob",
+                    "Sync users to mailjet",
+                    "0 0 7,19 ? * * *");
+
+            segueJobService = new SegueJobService(Arrays.asList(PIISQLJob, cleanUpOldAnonymousUsers, cleanUpExpiredReservations, syncMailjetUsers));
+            
+            String mailjetKey = properties.getProperty(MAILJET_API_KEY);
+            String mailjetSecret = properties.getProperty(MAILJET_API_SECRET);
+
+            if (mailjetKey == null && mailjetSecret == null) {
```
jsharkey13, line 930: This would be much cleaner if the job was never added in the constructor in the first place, instead of trying to remove it if there is no key configured. But I think you are right that we do need to keep the removal of jobs, because Quartz retains jobs whether or not the API registers them and we don't want the MailJet job failing forever if (e.g. locally) it isn't wanted but was at some point registered.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -914,7 +916,20 @@ private static SegueJobService getSegueJobService(final PostgresSqlDb database)
                     "SQL scheduled job that deletes expired reservations for the event booking system",
                     "0 0 7 * * ?", "db_scripts/scheduled/expired-reservations-clean-up.sql");
 
-            segueJobService = new SegueJobService(Arrays.asList(PIISQLJob, cleanUpOldAnonymousUsers, cleanUpExpiredReservations));
+            SegueScheduledJob syncMailjetUsers = new SegueScheduledSyncMailjetUsersJob(
+                    "syncMailjetUsersJob",
+                    "JavaJob",
+                    "Sync users to mailjet",
+                    "0 0 7,19 ? * * *");
```
jsharkey13, line 923: I'm now wondering whether I was wrong when I suggested these times and we should just run it every 6 hours at 12am/6am/12pm/6pm; or even every 4 hours at 12am/4am/8am/12pm/4pm/8pm. Especially in September when we hit the rate limiting a lot, waiting 12 hours before rerunning would mean quite the queue of further changes would back up.

That would be simple to change; if we were to change the Cron string later on, would it update the existing job, or will it be ignored like some other changed properties of jobs are?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
bh412: I believe this is addressed now and it runs fine with just the removal (no config set) even when there is no syncMailjetUsers job to be removed.

PR #382, opened 2021-09-14
MailGun email bounce callback hook

jsharkey13:
It would be nice if there was some standardised format for bounce notifications, but there is not. This adds an endpoint that MailGun can use to notify us of bounce emails, separate so as to keep the code cleaner. I had to add some notes to distinguish the different endpoints.

---

**Pull Request Check List**


bh412 (approved):
Not sure what form the request will come from mailgun in but the code looks to do what we want provided it's the right format!

PR #383, opened 2021-09-15
Ensure that the SMTP settings configured per-email are actually used

jsharkey13:
Previously we used Session.getDefaultInstance which only initialised using the provided properties once and after that ignored the provided properties entirely. Creating a new Session object for _every_ email is very expensive however, so we can cache it because the properties don't differ that much between our emails.
This is a prerequisite for the overrideEnvelopeFrom email property to be in any way effective, which we are going to use to route emails in Exim.

To test this, you don't need to be able to send emails (configure an invalid mailer in the segue_config), but can just observe whether it prints a log message (`Creating new mail session with properties: ...`) for every email or just once per-config. Since none of our emails use this really right now, you should only see one log message.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/util/Mailer.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/util/Mailer.java
@@ -16,6 +16,8 @@
 package uk.ac.cam.cl.dtg.util;
 
 import org.apache.commons.lang3.Validate;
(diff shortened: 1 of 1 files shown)


jsharkey13:
I apologise in advance for the crazy lambda function I use to avoid making a Session more often than truly necessary!

daviesian (approved):

PR #384, opened 2021-09-16
Enforce "this event is in the past" in a better way when booking events

jsharkey13:
Previously we checked whether the end date had passed, or whether the start date had passed. But some long-running events have booking deadlines during the event, where the start date has passed but the end date has not.
It makes sense to allow these bookings, but the old code did not.

If an event has an end date, use this and ignore the start date when working out if an event is in the past. Only if it does not have an end date should the start date be used.

Ian wonders whether, if there is a booking deadline, we should only look at that and enforce the "booking deadline before end" as a content error. But that might be more likely to break older events?


bh412 (approved):
Looks to be a sensible change!

PR #386, opened 2021-09-22
Don't include completed questions in filter results

mlt47:
Small bug seems to have been introduced here: https://github.com/isaacphysics/isaac-api/commit/31472df63b38e4a463d7f56117cc27bdd672d1e9


jsharkey13 (approved):
This does seem to fix my mistake from when I altered this to use a set! Whoops!

PR #387, opened 2021-09-28
Improved email instrumentation

jsharkey13:
Add better logging of where emails are being sent via to Prometheus, and add more detail to the admin email preview response.
For Prometheus, I think this is correct. Apparently null label values are not allowed and will cause NPEs, and so I had to load in the default from address to use if the override value is null. Perhaps this metrics stuff would be better in the actual sender class EmailCommunicator, but it was probabaly added before the queuing in the Manager for a reason.

Also clean up the creation of an EmailManager in Guice because it was messy.


mlt47 (approved):
Prometheus section looks fine to me. We can test it locally through `docker compose -f compose-local-monitor.yml up -d` but you also need to replace YOUR_IP_ADDRESS in the prometheus.yml.

PR #396, opened 2021-11-10
Check properties of registeredContext aren't null before calling them

bh412:
This feels a bit wrong, but given we use the endpoint without knowing the site we have to return the examBoard even when it either isn't set (fine) or is set to null (bad) in the registeredContext on Phy. I added a null check for the stage too to be safe. Is this a reasonable way to do an inline null check in Java or should I use the nullable check from utils?


Morpheu5 (approved):
It solves the immediate issue but we may want to revisit the whole CSV situation to see if we could improve it.

PR #397, opened 2021-11-15
Add deprecated attribute to Content and all subclasses

chrisjpurdy:
Doing this highlighted a weird inconsistency - some places in the API map the JSON attribute `version` to the Content class attribute `published`, but it doesn't look like `version` is actually used anywhere. This probably needs some kind of clean up.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/dos/content/Content.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/dos/content/Content.java
@@ -48,6 +48,7 @@ public class Content extends ContentBase {
     protected String attribution;
     protected List<String> relatedContent;
     protected Boolean published;
+    protected Boolean deprecated;
     protected Integer level;
     protected String searchableContent;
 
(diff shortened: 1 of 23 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dos/content/Content.java
```diff
@@ -202,6 +205,18 @@ public final void setChildren(List<ContentBase> children) {
         this.children = children;
     }
 
+    public Boolean getDeprecated() {
+        if (null == deprecated) {
```
bh412, line 209: Do we need this check? I think it would be nicer if we had a front end undefined check so that this isn't nescessary?

bh412: In fact, the front end PR does have an undefined check :)

chrisjpurdy: I was copying `getPublished` for consistency, but to be fair it really doesn't need to be there (never a good idea to copy bad code anyway) - I've removed it 👍

bh412 (approved):
Works as intended with the change - happy for this to go in for the next release, then the app/editor will be waiting until the release after that

PR #398, opened 2021-11-19
Conditional Quartz startup

jsharkey13:
I had intended to prevent Quartz from attempting to start on the standby (and then failing, because of the read-only database) by using a configuration flag. But it occurred to me it would be more sensible to just find out directly if the database is readonly and not start if so. That avoids the risk of the config getting out of sync, and means it fails-safe if we forget about it; Quartz will start if it would be able to run successfully.

Obviously if it doesn't start at API startup, we then need some way of starting it later, hence an admin endpoint to do so. Note that that endpoint does not check the database status and just attempts to initialise Quartz regardless.
To ensure we cannot forget to start Quartz, or leave it not running for extended periods, also add an info endpoint that can be used to monitor the status of Quartz on the live server.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -1304,7 +1304,7 @@ public class AdminFacade extends AbstractSegueFacade {
     @Path("/start_quartz")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = "Start the Quartz Job Scheduler service if not already started.")
-    public Response resetMisuseMonitor(@Context final HttpServletRequest request) {
+    public Response startQuartzJobService(@Context final HttpServletRequest request) {
         try {
             RegisteredUserDTO user = userManager.getCurrentRegisteredUser(request);
             if (!isUserAnAdmin(userManager, user)) {
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -34,6 +34,7 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.impl.client.DefaultHttpClient;
 import org.jboss.resteasy.annotations.GZIP;
+import org.quartz.SchedulerException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.ac.cam.cl.dtg.isaac.api.managers.EventBookingManager;
@@ -42,7 +43,6 @@ import uk.ac.cam.cl.dtg.segue.api.managers.IExternalAccountManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.SegueResourceMisuseException;
 import uk.ac.cam.cl.dtg.segue.api.managers.StatisticsManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
-import uk.ac.cam.cl.dtg.segue.api.managers.UserAssociationManager;
 import uk.ac.cam.cl.dtg.segue.api.monitors.IMisuseMonitor;
 import uk.ac.cam.cl.dtg.segue.api.monitors.SegueMetrics;
 import uk.ac.cam.cl.dtg.segue.api.monitors.UserSearchMisuseHandler;
@@ -67,6 +67,7 @@ import uk.ac.cam.cl.dtg.segue.dto.users.RegisteredUserDTO;
 import uk.ac.cam.cl.dtg.segue.dto.users.UserIdMergeDTO;
 import uk.ac.cam.cl.dtg.segue.dto.users.UserSummaryForAdminUsersDTO;
 import uk.ac.cam.cl.dtg.segue.etl.GithubPushEventPayload;
+import uk.ac.cam.cl.dtg.segue.scheduler.SegueJobService;
 import uk.ac.cam.cl.dtg.segue.search.SegueSearchException;
 import uk.ac.cam.cl.dtg.util.PropertiesLoader;
 import uk.ac.cam.cl.dtg.util.RequestIPExtractor;
@@ -128,9 +129,9 @@ public class AdminFacade extends AbstractSegueFacade {
 
     private final AbstractUserPreferenceManager userPreferenceManager;
     private final EventBookingManager eventBookingManager;
-    private final UserAssociationManager associationManager;
     private final IExternalAccountManager externalAccountManager;
     private final IMisuseMonitor misuseMonitor;
+    private final SegueJobService segueJobService;
 
     /**
      * Create an instance of the administrators facade.
@@ -151,8 +152,6 @@ public class AdminFacade extends AbstractSegueFacade {
      *            - for looking up school information
      * @param eventBookingManager
      *            - for using the event booking system
-     * @param associationManager
-     *            - so that we can create associations.
      * @param misuseMonitor
      *            - misuse monitor.
      */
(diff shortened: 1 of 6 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
```diff
@@ -1300,4 +1299,28 @@ public Response syncExternalAccounts(@Context final HttpServletRequest httpServl
                     "Fatal error while attempting to synchronise users!", e).toResponse();
         }
     }
+
+    @POST
+    @Path("/start_quartz")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Start the Quartz Job Scheduler service if not already started.")
+    public Response resetMisuseMonitor(@Context final HttpServletRequest request) {
```
mlt47, line 1307: Although it doesn't matter what a facade endpoint method name is, probably worth changing it :)

jsharkey13: 🤦

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
```diff
@@ -1300,4 +1299,28 @@ public Response syncExternalAccounts(@Context final HttpServletRequest httpServl
...
+            RegisteredUserDTO user = userManager.getCurrentRegisteredUser(request);
+            if (!isUserAnAdmin(userManager, user)) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+            if (!segueJobService.isStarted()) {
+                segueJobService.initialiseService();
+                log.info(String.format("Admin user (%s) started Quartz scheduler successfully.", user.getEmail()));
+                return Response.ok(ImmutableMap.of("status", "Started successfully!")).build();
+            } else {
+                return Response.ok(ImmutableMap.of("status", "Already running.")).build();
```
mlt47, line 1318: Do we know what happens if you try to start it and it is already running?
For this to be properly correct you'd want to hold a lock to make sure that separate requests don't ask if it is started between 1313 and 1314 or initialiseService() setting the isStarted flag. A `synchronized(segueJobService) {...}` block would be my suggestion assuming it is a singleton.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -924,18 +925,17 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
                     "Sync users to mailjet",
                     "0 0 0/4 ? * * *");
 
-            Collection<SegueScheduledJob> configuredScheduledJobs = new ArrayList<>(Arrays.asList(PIISQLJob, cleanUpOldAnonymousUsers, cleanUpExpiredReservations));
+            List<SegueScheduledJob> configuredScheduledJobs = new ArrayList<>(Arrays.asList(PIISQLJob, cleanUpOldAnonymousUsers, cleanUpExpiredReservations));
```
mlt47, line 928: I don't think the new ArrayList<>() is necessary here as I would expect Arrays.asList(...) to return a list?

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
jsharkey13: What about making the actual method `public synchronized void initialiseService()` ?

src/main/java/uk/ac/cam/cl/dtg/segue/database/PostgresSqlDb.java
```diff
@@ -86,4 +86,20 @@ public void close() {
...
+     * @return whether the database is read only.
+     */
+    public boolean isReadOnlyReplica() {
+        try (Connection conn = getDatabaseConnection()) {
+            PreparedStatement pst;
+            pst = conn.prepareStatement("SELECT pg_is_in_recovery()");
+            ResultSet results = pst.executeQuery();
+            results.next();
+            return results.getBoolean(1);
+        } catch (SQLException e) {
```
mlt47, line 101: I'd be tempted to say that a 500 might be more informative in the case of an exception?

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
jsharkey13: From the bytecode decompilation it looks like a no-op if you call start and it is started. This was only supposed to be for info, and not a serious indicator. But I agree about parallel requests and synchronising something to prevent that simultaneous init case.

src/main/java/uk/ac/cam/cl/dtg/segue/database/PostgresSqlDb.java
mlt47: Ignore that, it is an internal method, it does not return a HTTP response. I'd still be tempted to pass on the exception and letting the calling method deciding how to handle it.

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
mlt47: Making the initialiseService() synchronized will only make the requests queue up rather than not attempt it if it is not initialised or in the process of being initialised. You need the lock to cover checking and setting.
If it is a no-op it doesn't matter unless they decide to do something very strange in the future! This is unlikely to be a problem in practice.

src/main/java/uk/ac/cam/cl/dtg/segue/database/PostgresSqlDb.java
jsharkey13: Well, this isn't called by and HTTP code so Internal Server Error doesn't help here. But also you can get an SQL exception if the database is readonly or the query interferes with recovery, so in those cases it is true.
I'd make it throw if you felt it was important that it did?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
jsharkey13: I'll remove it, but it's not like I added that particular oddity ¯\\\_(ツ)\_/¯

src/main/java/uk/ac/cam/cl/dtg/segue/database/PostgresSqlDb.java
mlt47: I guess the more correct version would then catch the specific read-only exceptions (using the union types notation stuff!) to return false and the others are left uncaught. But, you're right, probably not important!

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
mlt47: I now see that initialiseService() also checks whether or not it is already running, in that case synchronizing that method will do the trick 👍

src/main/java/uk/ac/cam/cl/dtg/segue/database/PostgresSqlDb.java
jsharkey13: That would work if those exceptions corresponded to Java exceptions, but the JDBC driver only throws them all as `SQLException` and I'd have to process the contained Postgres error to work out what it was. Just throwing the exception will do for now.

src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
jsharkey13: Yeah, I made the initialisation method idempotent so I also think this is all safe with regards to your concerns 🙌

src/main/java/uk/ac/cam/cl/dtg/segue/database/PostgresSqlDb.java
```diff
@@ -86,4 +86,18 @@ public void close() {
...
+
+    /**
+     * Check whether the database is a read only replica or not.
+     * @return whether the database is read only.
+     */
+    public boolean isReadOnlyReplica() throws SQLException {
+        try (Connection conn = getDatabaseConnection();
+             PreparedStatement pst = conn.prepareStatement("SELECT pg_is_in_recovery()");
+             ResultSet results = pst.executeQuery();
+        ) {
```
mlt47, line 98: Nice! I was going to have a go at fixing that LGTM error. I think it should apply to almost all our database code!

jsharkey13: This felt like a good test-case, since the statement contains no variables and there's nothing in the try beyond the expected auto-closeable resources. If LGTM doesn't comment below this commit, then success and a potential model going forward.

mlt47 (approved):

PR #399, opened 2021-11-19
Do not make additional status code checks on MailJet requests

jsharkey13:
MailJet's API actually already does this under the hood in [`MailjetResponseUtil::validateMailjetResponse`](https://github.com/mailjet/mailjet-apiv3-java/blob/mailjet-client-5.2.0/src/main/java/com/mailjet/client/MailjetResponseUtil.java#L28-L40) and we're blocking the 202 and 201 status codes they sometimes use with the old implementation we had!

This was my mistake, I asked Ben to add this and I shouldn't have.


bh412 (approved):
Stops the unnecessary response status checks and retains the deletion request - looks good!

PR #401, opened 2021-12-13
Add stage and difficulty stats to getUserQuestionInformation

bh412:
Adds stage and difficulty question information to user attempts in the StatisticsManager. It builds it in the same way as tags but because it's a nested map it's not the easiest to follow. If there's a simpler approach that is staring me in the face please let me know! At the moment the information is returned as a `Map<Stage, Map<Difficulty, Integer>>`

---
**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (changes requested):
This is so complicated and hard to read.

Could you not factor out `currentStage = audience.getStage().get(0)` and `currentDifficulty = audience.getDifficulty().get(0)` right after the second if (before the "Count attempts at question" comment), to at least improve readability? After that it might be more obvious how to improve the nested maps.

jsharkey13 (approved):
I really don't like the double-curly brackets initialisation business, but there's probably nothing wrong with it in this use-case since this is not a common code path and the associated memory demands will be small compared to the fact we have just copied all the question attempts for a user into memory,

jsharkey13 (approved):
🙈

PR #402, opened 2021-12-14
Add hiddenFromRoles property to quizzes to allow more flexible filtering

jsharkey13:
Currently there is only "visibleToStudents", this is the first step to deprecating that property. The changes attempt to be backwards compatible; if the content does not have the new property but only the old one (as it must for one release cycle) then the old behaviour should be unchanged.
We will then need to remove the old property from the content and then from the DOs and DTOs.

The new property is a list of roles (in all-caps as in our enums: `["STUDENT", "TEACHER"]`) who cannot see the quiz to attempt it freely, but only see it if assigned. There is currently no clever checking of roles, so it is possible to make a quiz that nobody can view or assign (by limiting all roles except students, say). Perhaps this may want to be a content error eventually, but for now it is okay.
The frontend and editor will need updating to use this new property. For the editor, I wonder if a radio button choice between "anyone" which leaves the property as `null` (i.e. anyone can see it), "only teacher assigned" which is `["STUDENT"]` (i.e. the current visibleToStudents=true), and "only staff" which is `["STUDENT", "TEACHER"]`. 
CS may eventually want to worry about EVENT_LEADERS who are TEACHERS but with some extra permissions, so maybe a multi-select will be necessary.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -169,8 +169,6 @@ public class QuizFacade extends AbstractIsaacFacade {
      * Get quizzes visible to this user, starting from index 0.
      *
      * Anonymous users can't see quizzes.
-     * Students can see quizzes with the visibleToStudents flag set.
-     * Teachers and higher can see all quizzes.
      *
      * @return a Response containing a list of ContentSummaryDTO for the visible quizzes.
      */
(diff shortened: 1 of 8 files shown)


bh412 (approved):
This all looks sensible - I think having the exclusion list seems the best, just as you say maybe make it hard in the editor implementation to exclude every role >= teacher so it can still be set.

chrisjpurdy (approved):
Looked though it while implementing the front-end and all seems good in this PR. I agree with the proposed editor change design, restricting content to being only hidden/shown to students and teachers is fine for the foreseeable future.

PR #403, opened 2021-12-17
Add status code to request duration metrics

jsharkey13:
I am unsure whether this is best, or whether a new metric with the path and the status code (without the timing information) might be better.

@mlt47 - opinions? If you think a new metric is better, please do close this. I don't want to make an enormous expansion in the number of tracked things by adding a third dimension if that's not wise!

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/monitors/PerformanceMonitor.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/monitors/PerformanceMonitor.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015 Stephen Cummins
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/monitors/PerformanceMonitor.java
```diff
@@ -90,8 +89,11 @@ public void filter(final ContainerRequestContext requestContext, final Container
 
         // Record for metrics
         REQUEST_LATENCY_HISTOGRAM
-                .labels(requestContext.getMethod(), monitorService.getPathWithoutPathParamValues(request.getUri()))
-                .observe((double)timeInMs / NUMBER_OF_MILLISECONDS_IN_A_SECOND);
+                .labels(
+                        requestContext.getMethod(),
+                        monitorService.getPathWithoutPathParamValues(request.getUri()),
+                        String.valueOf(responseContext.getStatus())
```
mlt47, line 95: Just a note that getStatus returns an int and therefore -1 when the status is not, if that was ever the case. This is probably a good thing here because String.valueOf(...) throws a null pointer exception if it is given null!

mlt47 (approved):
As the number of statuses are bounded and, in our case, has a small cardinality, I'm not worried about the expansion. Will be a useful addition to our monitoring and alerting.

PR #404, opened 2022-01-07
Mark correct numeric units as correct even for a known incorrect choice

jsharkey13:
This will only work in some circumstances:
 - the units provided match a known correct answer.
 - the value and units both match a known incorrect answer.

Then, and only then, can we safely mark the units as correct and the value as incorrect. Limited in usefulness, but addresses a noticed issue in the content without (I hope!) any side effects.
Also add a test for the noticed case for future reference.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
@@ -251,10 +251,11 @@ public class IsaacNumericValidator implements IValidator {
 
                 // What sort of match do we have:
                 if (numericValuesMatched && unitsFromUser.equals(unitsFromChoice)) {
-                    // Exact match: nothing else can do better.
+                    // Exact match: nothing else can do better, but previous match may tell us if units are also correct:
+                    Boolean unitsCorrect = (null != bestResponse && bestResponse.getCorrectUnits()) || quantityFromQuestion.isCorrect();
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -251,10 +251,11 @@ private QuantityValidationResponse validateWithUnits(final IsaacNumericQuestion
 
                 // What sort of match do we have:
                 if (numericValuesMatched && unitsFromUser.equals(unitsFromChoice)) {
-                    // Exact match: nothing else can do better.
+                    // Exact match: nothing else can do better, but previous match may tell us if units are also correct:
+                    Boolean unitsCorrect = (null != bestResponse && bestResponse.getCorrectUnits()) || quantityFromQuestion.isCorrect();
                     bestResponse = new QuantityValidationResponse(isaacNumericQuestion.getId(), answerFromUser,
                             quantityFromQuestion.isCorrect(), (Content) quantityFromQuestion.getExplanation(),
-                            quantityFromQuestion.isCorrect(), quantityFromQuestion.isCorrect(), new Date());
+                            quantityFromQuestion.isCorrect(), unitsCorrect, new Date());
```
bh412, line 258: This seems to make more sense than the `quantityFromQuestion...` that was here before

bh412 (approved):
This achieves what was aimed for, I don't think it will cause any unexpected negative behaviour

PR #405, opened 2022-01-19
Cloze question subset matching of item choices

chrisjpurdy:
This enables subset matching for cloze questions - if the question **doesn't have `withReplacement` enabled**, then it is possible to match a subset of an item choice.

This is useful for when a content creator wants to give feedback for an incorrect answer that contains a particular item. 
You could imagine one of the items being a completely made up word, and if the user's answer includes that word, the specific feedback might sarcastically ask the user to go and find a definition of it (might not be a practical example, but it's an example).

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacClozeValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacClozeValidator.java
@@ -139,7 +139,7 @@ public class IsaacClozeValidator implements IValidator {
                 Set<String> choiceItemIdSet = new HashSet<>(choiceItemIds);
                 /* If the intersection of the submitted and choice ids is equal to the choice ones, then
                    this means that:
-                    - choiceItemIds.size() <= submittedItemIds.size()
+                    - choiceItemIdSet.size() <= submittedItemIds.size()
                     - All choice ids are within the set of submitted ids
                  */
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacClozeValidator.java
```diff
@@ -148,4 +162,47 @@ public final QuestionValidationResponse validateQuestionResponse(final Question
         return new QuestionValidationResponse(question.getId(), answer, responseCorrect, feedback, new Date());
     }
 
+    @Override
```
bh412, line 165: With this all being duplicated from the `IsaacItemQuestionValidator`, would it work to not have this and instead have line 97 as `List<Choice> orderedChoices = new IsaacItemQuestionValidator().getOrderedChoices(clozeQuestion.getChoices());`? I'm not entirely sure this would even work though

chrisjpurdy: I was thinking of extending `IsaacItemQuestionValidator`, but since `validateQuestionResponse` is always declared `final` I wasn't able to do it that way. 
I've moved the code in `getOrderedChoices` to a static method in `IsaacItemQuestionValidator`. This lets me refer to it in both places without a problem.

bh412 (approved):
Looks good now some of the duplication has been removed

PR #406, opened 2022-01-19
Add additional event booking information auto deletion

bh412:
Deletes additional booking information from the `event_bookings` table 30 days after the event's end date (if present) or event date (if not). To prevent logging of previously removed additional booking information there is a check for any of the 4 fields being present and removed - perhaps this is too loose and might let some additional booking information remain? I'm not entirely sure how to prevent it running over all previous events with removed data otherwise, as `executeUpdate` returns the number of rows updated, even if there has been no change (e.g. the properties have already been removed)

---

**Pull Request Check List**
- [ ] Peer-Reviewed


bh412:
I set the cronString to `0 * * ? * * *` (once a minute) for testing locally

Morpheu5:
> perhaps this is too loose and might let some additional booking information remain? I'm not entirely sure how to prevent it running over all previous events with removed data otherwise, as `executeUpdate` returns the number of rows updated, even if there has been no change (e.g. the properties have already been removed)

As far as I can see, there is no other way than to examine every single booking and test for previously deleted information, then queue for updates only those who need updating.

In the end, does it even matter? Do we run (or will we run) stats on these deletion logs?

Otherwise, the code looks fine to me.

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/DeleteEventAdditionalBookingInformationJob.java
```diff
@@ -0,0 +1,120 @@
...
+                                    "UPDATE event_bookings SET additional_booking_information=jsonb_set(jsonb_set(jsonb_set(jsonb_set(additional_booking_information," +
+                                    " '{emergencyName}', '\"[REMOVED]\"'::JSONB, FALSE)," +
+                                    " '{emergencyNumber}', '\"[REMOVED]\"'::JSONB, FALSE)," +
+                                    " '{accessibilityRequirements}', '\"[REMOVED]\"'::JSONB, FALSE)," +
+                                    " '{medicalRequirements}', '\"[REMOVED]\"'::JSONB, FALSE)," +
+                                    " pii_removed=? " +
+                                    " WHERE event_id = ?" +
+                                    " AND additional_booking_information ??| array['emergencyName', 'emergencyNumber', 'accessibilityRequirements', 'medicalRequirements']" +
+                                    " AND pii_removed IS NULL");
+                            pst.setTimestamp(1, new Timestamp(new Date().getTime()));
```
mlt47, line 97: In this class we use a mix of the old style java.util.Date and new style Instant stuff from java.time which makes me uneasy but it is also a problem throughout the code (we also use the Joda-Time library which is basically the same as the java.time module I think). I will add it to the list of tech debt because time and date is complicated enough as it is without mixing with semi-deprecated classes!

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/DeleteEventAdditionalBookingInformationJob.java
```diff
@@ -0,0 +1,120 @@
...
+                    }
+                }
+            }
+            log.info("Ran DeleteEventAdditionalBookingInformationJob");
+        } catch (SQLException e) {
+            try {
+                throw new SegueDatabaseException("Postgres exception", e);
+            } catch (SegueDatabaseException ex) {
+                ex.printStackTrace();
+            }
```
mlt47, line 114: If you want to wrap it as a SegueDatabaseException I would have thought that `new SegueDatabaseEception("Postgres exception", e).printStackTrace();` is equivalent to this code

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/DeleteEventAdditionalBookingInformationJob.java
```diff
@@ -0,0 +1,120 @@
...
+        } catch (ContentManagerException e) {
+            log.error("Failed to delete event additional booking information");
+            e.printStackTrace();
```
mlt47, line 117: Is this what we want to do? I'm guessing the errors are piped into logs?

mlt47 (changes requested):
One change that needs to be made (the conditional deciding if the event is 30 days ago), the others are just suggestions and comments to be aware of.

bh412:
> > > perhaps this is too loose and might let some additional booking information remain? I'm not entirely sure how to prevent it running over all previous events with removed data otherwise, as `executeUpdate` returns the number of rows updated, even if there has been no change (e.g. the properties have already been removed)
> > 
> > 
> > As far as I can see, there is no other way than to examine every single booking and test for previously deleted information, then queue for updates only those who need updating.
> 
> I think the issue Ben mentioned has been addressed by the additional field in the database recording when the PII was updated. Is that correct, Ben, or have I misunderstood the problem?

Yep, this is addressed now! I'll have a look through your other comments, the factory issue may be from mirroring the mailjet sync job code

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/DeleteEventAdditionalBookingInformationJob.java
bh412: So for the failed to sync mailjet users exception we email the ticketing system - plus it is logged, do you think emailing is necessary here?

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/SegueScheduledDeleteEventAdditionalBookingInformationJob.java
```diff
@@ -0,0 +1,24 @@
+package uk.ac.cam.cl.dtg.segue.scheduler.jobs;
+
+import com.google.common.collect.Maps;
+import org.quartz.Job;
+import uk.ac.cam.cl.dtg.segue.scheduler.SegueScheduledJob;
+
+import java.util.Map;
+
+public class SegueScheduledDeleteEventAdditionalBookingInformationJob extends SegueScheduledJob {
```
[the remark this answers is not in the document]
bh412: I think I've addressed this! I could look at separately moving the classes that extend SegueScheduledJob to the same thing

jsharkey13 (approved):
I feel like we should have some unit tests here, this is one example where mocking and returning some faked events would be a good test of the logic. But it looks fine to me.

Morpheu5 (approved):

PR #408, opened 2022-02-04
Add name length and illegal character validation to user create/update endpoint

mwtrew:
---

**Pull Request Check List**
- [ ] Security - Injection - everything run by an interpreter (SQL, OS...) is either validated or escaped _(Not applicable)_
- [ ] Security - Access Control - Check authorisation on every new endpoint _(Not applicable)_
- [ ] Security - New dependency - configured sensibly not relying on defaults _(Not applicable)_
- [ ] Security - New dependency - Searched for any know vulnerabilities _(Not applicable)_
- [ ] Security - New dependency - Signed up team to mailing list _(Not applicable)_
- [ ] Security - New dependency - Added to dependency list _(Not applicable)_
- [ ] DB schema changes - postgres-rutherford-create-script updated _(Not applicable)_
- [ ] DB schema changes - upgrade script created matching create script _(Not applicable)_
- [ ] Updated Release Procedure & Documentation (& Considered Implications to Previous Versions) _(Not applicable)_
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
@@ -774,7 +774,7 @@ public class UserAccountManager implements IUserAccountManager {
         }
 
         if (!this.isUserNameValid(user.getFamilyName())) {
-            throw new InvalidNameException("The given name provided is too long or contains illegal characters.");
+            throw new InvalidNameException("The family name provided is too long or contains illegal characters.");
         }
 
         IPasswordAuthenticator authenticator = (IPasswordAuthenticator) this.registeredAuthProviders
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
@@ -770,11 +770,11 @@ public class UserAccountManager implements IUserAccountManager {
 
         // validate names
         if (!this.isUserNameValid(user.getGivenName())) {
-            throw new InvalidNameException("The given name provided is too long or contains illegal characters.");
+            throw new InvalidNameException("The given name provided is an invalid length or contains illegal characters.");
         }
 
         if (!this.isUserNameValid(user.getFamilyName())) {
-            throw new InvalidNameException("The family name provided is too long or contains illegal characters.");
+            throw new InvalidNameException("The family name provided is an invalid length or contains illegal characters.");
         }
 
         IPasswordAuthenticator authenticator = (IPasswordAuthenticator) this.registeredAuthProviders
@@ -857,11 +857,11 @@ public class UserAccountManager implements IUserAccountManager {
 
         // validate names
         if (!this.isUserNameValid(updatedUser.getGivenName())) {
-            throw new InvalidNameException("The given name provided is too long or contains illegal characters.");
+            throw new InvalidNameException("The given name provided is an invalid length or contains illegal characters.");
         }
 
         if (!this.isUserNameValid(updatedUser.getFamilyName())) {
-            throw new InvalidNameException("The family name provided is too long or contains illegal characters.");
+            throw new InvalidNameException("The family name provided is an invalid length or contains illegal characters.");
         }
 
         IPasswordAuthenticator authenticator = (IPasswordAuthenticator) this.registeredAuthProviders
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
```diff
@@ -1578,6 +1601,24 @@ private boolean isUserValid(final RegisteredUser userToValidate) {
...
     }
 
+    /**
+     * This function checks that the name provided is valid.
+     *
+     * @param name
+     *            - the name to validate.
+     * @return true if the name is valid, false otherwise.
+     */
+    public boolean isUserNameValid(final String name){
```
bh412, line 1611: This could be `final` to be more explicit but having looked at the methods around it, they aren't either...

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
```diff
@@ -1578,6 +1601,24 @@ private boolean isUserValid(final RegisteredUser userToValidate) {
...
+        boolean isValid = true;
+
+        Pattern illegalCharacters = Pattern.compile(USER_NAME_ILLEGAL_CHARS_REGEX);
+        if (name.length() > USER_NAME_MAX_LENGTH || illegalCharacters.matcher(name).find()) {
+            isValid = false;
```
bh412, line 1616: It might be a little nicer to have this as `return false;` and line 1619 as `return true;` removing the need for line 1612 - which matches most of the boolean returns we have, though the method above also doesn't follow this style.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
```diff
@@ -98,6 +99,10 @@
 
     private final ISecondFactorAuthenticator secondFactorManager;
 
+    private final int USER_NAME_MAX_LENGTH = 255;
```
bh412, line 102: I know it wasn't in the request but I think it would be a minor change to also add a `USER_NAME_MIN_LENGTH = 1;` and have it greater than or equal to that too? We do the front end 'this must have a value' check but nothing back end too...

bh412 (commented):
Does what it needs to and includes tests, would you mind adding the min check too? (which wasn't in the spec sorry)

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
jsharkey13: I think we should enforce that with a "not `.isEmpty()` check, rather than a hardcoded 1 value!

bh412: Yes, that would be far nicer wouldn't it, d'oh

mwtrew: Good point - I'll check for an empty name which should be equivalent.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
mwtrew: Added.

bh412 (approved):
Works as intended and covered by tests

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
```diff
@@ -763,6 +768,15 @@ public RegisteredUserDTO createUserObjectAndSession(final HttpServletRequest req
             throw new MissingRequiredFieldException("The email address provided is invalid.");
         }
 
+        // validate names
+        if (!this.isUserNameValid(user.getGivenName())) {
+            throw new InvalidNameException("The given name provided is an invalid length or contains illegal characters.");
```
jsharkey13, line 773: I think "forbidden characters" might have been more clear to a child than "illegal" which has other connotations outside Computer Science disciplines!
Also, unless we want to provide feedback about what specifically is wrong with the name, this message could be in a default constructor in the `InvalidNameException` class so it wasn't repeated.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
```diff
@@ -1568,14 +1591,26 @@ private RegisteredUser registerUserWithFederatedProvider(final AuthenticationPro
...
+
+    /**
+     * This function checks that the name provided is valid.
+     *
+     * @param name
+     *            - the name to validate.
+     * @return true if the name is valid, false otherwise.
+     */
+    public final boolean isUserNameValid(final String name){
+        Pattern illegalCharacters = Pattern.compile(USER_NAME_ILLEGAL_CHARS_REGEX);
```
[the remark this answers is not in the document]
mwtrew: Agreed, will address in a new PR.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAccountManager.java
mwtrew: Given name/family name aligns with the property names in the request, so I think this is a sensible response from the backend. The frontend will need to change to handle this consistently anyway (it currently also validates length but not special characters).

mwtrew:
@jsharkey13, I've made the suggested changes in [a new PR](https://github.com/isaacphysics/isaac-api/pull/414).

PR #410, opened 2022-02-08
Move PreparedStatements into try-with-resources blocks

jsharkey13:


chrisjpurdy (approved):
I couldn't find any errors! My eyes hurt from reviewing that, I can't imagine what making the changes in the first place was like 😅👀

PR #412, opened 2022-02-09
Ignore WebSocket idle timeout errors

jsharkey13:
These seem to happen a lot, and are likely caused by browsers unloading background tabs so that WebSockets do not respond to ping frames.
Since WebSockets attempt to reconnect when they can, this error is not that important.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
@@ -66,6 +66,7 @@ public class UserAlertsWebSocket implements IAlertListener {
     private Session session;
     private static ObjectMapper objectMapper = new ObjectMapper();
 
+    private static final String IDLE_TIMEOUT_MESSAGE = "Idle timeout expired";
     // Named unsafeConnectedSockets because, although non-aggregate operations on the concurrent hash map are fine,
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
```diff
@@ -299,7 +300,10 @@ public void onClose(final Session session, final int status, final String reason
     @OnWebSocketError
     public void onError(final Session session, final Throwable error) {
         long connectedUserId = connectedUser.getId();
-        log.warn(String.format("Error in WebSocket for user (%s): %s", connectedUserId, error));
+        String errorString = error.toString();
+        if (!errorString.contains(IDLE_TIMEOUT_MESSAGE)) {
```
mlt47, line 304: I would feel better checking the Exception's type rather than the message but can't find one which is more specific than WebSocketTimeoutException or concurrent.TimeoutException which would hide more than just the idle case.
Not an important enough problem to worry too much about it being accidentally reintroduced on library upgrade.

mlt47 (approved):

src/main/java/uk/ac/cam/cl/dtg/segue/api/userAlerts/UserAlertsWebSocket.java
jsharkey13: My thought process exactly! 😅

PR #415, opened 2022-02-11
Further try-with-resources conversions

jsharkey13:
Now that the bulk of them are sorted, it is much easier to see what is left and address those. This includes some CheckStyle changes too, but as separate commits so if reviewing all in one becomes too hard try commit by commit.

Again, review ignoring whitespace: https://github.com/isaacphysics/isaac-api/pull/415/files?w=1


chrisjpurdy (approved):
LGTM 👍

PR #418, opened 2022-02-24
PDF download endpoint

jsharkey13:
Currently blanket-limited to teacher accounts and above, with a new log type `DOWNLOAD_FILE` recording what path was downloaded.
Left the switch statement in, in case we add more supported types.

I am unsure how we make this more general; perhaps some files will want students to be able to download them too? But for now it will work.


chrisjpurdy (approved):
All absolutely fine! 👍

PR #419, opened 2022-03-04
Fix site-wide search doc type validation

mlt47:
This PR fixes the validation of document types sent to site-wide search.

The facade method returns a bad request if any of the document types are invalid.
The manager filters any invalid document types. If the list of valid document types are then empty (or the more common case of the user not specifying document types) then we search for all valid document types.

---

**Pull Request Check List**
_We do not tend to log bad requests_
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
@@ -205,6 +205,11 @@ public class IsaacController extends AbstractIsaacFacade {
                 showHiddenContent = isUserStaff(userManager, (RegisteredUserDTO) currentUser);
             }
             List<String> documentTypes = !types.isEmpty() ? Arrays.asList(types.split(",")) : null;
+            // Return an error if any of the proposed document types are invalid
+            if (documentTypes != null && !SITE_WIDE_SEARCH_VALID_DOC_TYPES.containsAll(documentTypes)) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
```diff
@@ -205,6 +205,11 @@ public final Response search(@Context final Request request, @Context final Http
                 showHiddenContent = isUserStaff(userManager, (RegisteredUserDTO) currentUser);
             }
             List<String> documentTypes = !types.isEmpty() ? Arrays.asList(types.split(",")) : null;
+            // Return an error if any of the proposed document types are invalid
+            if (documentTypes != null && !SITE_WIDE_SEARCH_VALID_DOC_TYPES.containsAll(documentTypes)) {
```
jsharkey13, line 209: Does this work if `documentTypes` is empty? I assume so?

jsharkey13 (approved):
That's one concise stream for `validDocumentTypes`! Looks correct and surprisingly readable though!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
jsharkey13: Great! I think the comma only case will cause two empty strings, and they aren't in the set so it is all okay 🎉

PR #424, opened 2022-03-16
Migrate segue DO and DTO classes to isaac

jsharkey13:
The `segue` - `isaac` divide has not been particularly useful, and this is a test step towards removing it.

Move the DO and DTO classes entirely into the `isaac` world, along with the `quiz` package.


jsharkey13 (commented):
I have manually reviewed all 366 files in the PR. IntelliJ did the refactoring well, and has also cleaned up imports of files in the codebase. In at least one case it has also adjusted the line endings, too; hopefully to be consistent with the majority.
Some imports that became unnecessary (i.e. importing from `segue.quiz` in `isaac.quiz` needed to be explicit, but now they are merged the name is in the same package and so can be an implicit name) have been removed too.

PR #425, opened 2022-03-28
Generate isaac-data-types.d.ts for the editor to use

robinmessage:
This adds a way to generate the DO types into Typescript for use in the new editor. Note the editor works on the DO types, not the DTO types.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


mlt47 (approved):

mlt47:
Will hold off merging into master until after tomorrow's release. Although it's clearly unlikely to cause a problem, our process currently is to freeze between setting the site up for full manual and automated regression testing (Tuesdays) until the release (Wednesdays).

robinmessage:
@jsharkey13 Thanks James, that makes sense and renaming it is fine as I just copy it into the front-end.

PR #429, opened 2022-03-30
Assignments endpoint optimisation

jsharkey13:
This PR contains three separate optimisations to reduce inefficiency in the `GET /assignments` endpoint:

 - Instead of searching ElasticSearch for questions matching the IDs of the contents of each gameboard, just get the objects directly by ID. This does less validation of the actual objects in question, but does then use the existing cache in the `GitContentManager`, which will speed up subsequent requests for the same set of questions. This may (dramatically?) increase heap allocation (and garbage collection?) pressure, since this is one of the most common endpoints.

 - No longer include `savedToCurrentUser` information on bulk gameboard requests; when loading an individual board, it makes sense to ask if it is connected to the user, but in bulk we do not use this information anywhere and adding it comes at the cost of a query per board which can be in the hundreds for many users!

 - Look up group details only once per group when adding group information to assignment summaries. We already did some caching of owner and manager detail lookups, so this was a logical and simple additional step.

For two frequent staff and teacher users, this speeds up the request duration by more than 10x if the cache is hit. It also reduces the number of database queries made per HTTP request by 10-25x. (Yes, nearly _thirty times_ fewer queries for one request, down from 3500 to 110 which is still a lot but not by comparison!).

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -273,10 +273,9 @@ public class AssignmentFacade extends AbstractIsaacFacade {
                 Collection<AssignmentDTO> allAssignmentsSetToGroup
                         = this.assignmentManager.getAssignmentsByGroup(group.getId());
 
-                // In order to get all the information about gameboard items, we need to use the method which augments
-                // gameboards with user attempt information. But we don't _want_ this information for real, so we won't
-                // do the costly loading of the real attempt information from the database:
-                // FIXME: would GameManager::getGameboards with only one argument work correctly / be better?
+                // In order to get all the information about parts and pass marks, which is needed for Assignment Progress,
+                // we need to use the method which augments gameboards with user attempt information.
+                // But we don't _want_ the attempt information itself for real, so we won't load it from the database:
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -272,11 +276,12 @@ public Response getAssigned(@Context final HttpServletRequest request,
                 // In order to get all the information about gameboard items, we need to use the method which augments
                 // gameboards with user attempt information. But we don't _want_ this information for real, so we won't
                 // do the costly loading of the real attempt information from the database:
+                // FIXME: would GameManager::getGameboards with only one argument work correctly / be better?
```
mwtrew, line 279: Minor, but I think this has potential to become an immortal 'FIXME' - is it worth following up before we merge?

mwtrew (approved):

Morpheu5 (approved):
The two comments can be ignored/postponed if in a rush, but it'd be nice to think about at least the first one (list queries).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
jsharkey13: Okay, I followed it up and the answer is "no". So I have updated the comment to better explain _why_ and removed the FIXME :P

PR #430, opened 2022-03-30
Add ETL timing logs and extra bulk indexing

chrisjpurdy:
The branch name is a bit ominous, but this PR just adds bulk indexing for units, published units and content errors (which speeds up ETL phy locally by about 8 seconds). 

Importantly, some logging is added that times various parts of the indexing process. This will be helpful in diagnosing what exactly is taking ages on the live server.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
@@ -102,33 +102,30 @@ public class ContentIndexer {
 
             long totalStartTime, startTime, endTime;
 
-            log.debug("Beginning to populate the Git content cache");
-
             totalStartTime = System.nanoTime();
             buildGitContentIndex(version, true, contentCache, tagsList, allUnits, publishedUnits, indexProblemCache);
             endTime = System.nanoTime();
 
-            log.debug("Finished populating Git content cache, took: " + ((endTime - totalStartTime) / 1000000) + " milliseconds");
-            log.debug("Beginning to record content errors");
+            log.info("Finished populating Git content cache, took: " + ((endTime - totalStartTime) / 1000000) + "ms");
+            log.info("Beginning to record content errors");
 
             startTime = System.nanoTime();
             recordContentErrors(version, contentCache, indexProblemCache);
             endTime = System.nanoTime();
 
-            log.debug("Finished recording content errors, took: " + ((endTime - startTime) / 1000000) + " milliseconds");
-            log.debug("Beginning to build elasticsearch index from Git content cache");
+            log.info("Finished recording content errors, took: " + ((endTime - startTime) / 1000000) + "ms");
 
             startTime = System.nanoTime();
             buildElasticSearchIndex(version, contentCache, tagsList, allUnits, publishedUnits, indexProblemCache);
             endTime = System.nanoTime();
-            log.debug("Finished indexing git content cache, took: " + ((endTime - startTime) / 1000000) + " milliseconds");
+            log.info("Finished indexing git content cache, took: " + ((endTime - startTime) / 1000000) + "ms");
 
             // Verify the version requested is now available
             if (!es.hasIndex(version, CONTENT_INDEX_TYPE.CONTENT.toString())) {
                 throw new Exception(String.format("Failed to index version %s. Don't know why.", version));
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
```diff
@@ -629,42 +636,63 @@ private synchronized void buildElasticSearchIndex(final String sha,
...
-                        objectMapper.writeValueAsString(ImmutableMap.of("cleanKey", k, "unit", allUnits.get(k))));
-            }
-            for (String k : publishedUnits.keySet()) {
-                es.indexObject(sha, CONTENT_INDEX_TYPE.PUBLISHED_UNIT.toString(),
-                        objectMapper.writeValueAsString(ImmutableMap.of("cleanKey", k, "unit", publishedUnits.get(k))));
-            }
+            startTime = System.nanoTime();
+            es.bulkIndex(sha, CONTENT_INDEX_TYPE.UNIT.toString(), allUnits.entrySet().stream().map(entry -> {
+                try {
+                    return immutableEntry((String) null, objectMapper.writeValueAsString(ImmutableMap.of("cleanKey", entry.getKey(), "unit", entry.getValue())));
```
mlt47, line 650: It's a shame to have to create an immutableEntry with null as the first value for these.
It would be nice to be able to override the `bulkIndex` method, one accepting a List of Strings and one accepting the current List of Entries. Unfortunately, we can't overload it as Java erases the type parameter in generics and so would not be able to decide which of the two methods to choose from during runtime. However it would be fine if one had a different name so I would suggest renaming the current one to something like `bulkIndexWithIDs(...)` and then creating a `bulkIndex(...)` which hands off the actual work to `bulkIndexWithIDs(...)` after transforming the `dataToIndex` using a stream like you have here.

src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
```diff
@@ -629,42 +636,63 @@ private synchronized void buildElasticSearchIndex(final String sha,
...
+                } catch (JsonProcessingException jsonProcessingException) {
+                    log.error("Unable to serialise unit entry for unit: " + entry.getValue());
+                    return null;
+                }
+            }).filter(Objects::nonNull).collect(Collectors.toList()));
```
mlt47, line 655: The filter is nice given that Java doesn't seem to have a particularly good way of combining checked exceptions and streams.

mlt47 (changes requested):
This looks good. Some small code style suggestions.

src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
chrisjpurdy: Agreed, I was trying to get an MVP working and felt that it might be okay to leave it as is, but it would much cleaner and clearer to have a `bulkIndexWithIDs(...)` as you suggest. Will get that working 👍

src/main/java/uk/ac/cam/cl/dtg/segue/etl/ContentIndexer.java
```diff
@@ -107,18 +100,32 @@ void loadAndIndexContent(String version) throws Exception, VersionLockedExceptio
             Map<String, String> publishedUnits = new HashMap<>();
             Map<Content, List<String>> indexProblemCache = new HashMap<>();
 
+            long totalStartTime, startTime, endTime;
+
+            totalStartTime = System.nanoTime();
             buildGitContentIndex(version, true, contentCache, tagsList, allUnits, publishedUnits, indexProblemCache);
+            endTime = System.nanoTime();
+
+            log.info("Finished populating Git content cache, took: " + ((endTime - totalStartTime) / 1000000) + "ms");
```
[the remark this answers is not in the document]
chrisjpurdy: Felt the same way but erred on the side of "Oh it'll be fine its fairly clear what's going on" - I'll put this in since I'm implementing the extra bulk index by ID function 👍

chrisjpurdy:
Would it be possible to test this locally when you review the new changes please @mlt47 - my local ETL is having issues connecting to the content repo and I'm fairly sure these changes have nothing to do with it.

PR #434, opened 2022-04-07
Prefer topic results when searching site

mlt47:
CS have requested that we rank topic summary pages higher.
From trial and error, boosting their match results by 10 seems to produce a subjectively better set of results.
This will not alter Physics' results as they do not have pages of this type.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (approved):
It's tricky to get the balance right here. Let's go with this!

PR #437, opened 2022-04-14
Feature/event pre post email

bh412:
Needs the `scheduled_emails` table adding:
```
CREATE TABLE scheduled_emails (
    email_id TEXT NOT NULL,
    sent TIMESTAMP WITHOUT TIME ZONE,
    CONSTRAINT scheduled_emails_pk PRIMARY KEY (email_id)
);
```
Some commenting is left in for debugging - still needs some general tidying / javadoc rewording - plus tests adding

---

**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13:
> Needs the `scheduled_emails` table adding:
> 
> ```
> CREATE TABLE scheduled_emails (
>     email_id TEXT NOT NULL,
>     sent TIMESTAMP WITHOUT TIME ZONE,
>     CONSTRAINT scheduled_emails_pk PRIMARY KEY (email_id)
> );
> ```

Could you ensure this is included as a migration SQL file in `src/main/resources/db_scripts`, please, so that it is not lost!

In fact, if you're feeling helpful, you could even make a new subfolder `migrations` and name the migration SQL file like `YYYY-MM-some_description_here.sql` and that can be the future plan for this sort of thing.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/ScheduledEmailManager.java
```diff
@@ -0,0 +1,176 @@
...
+        // Magic number
+        Integer limit = 10000;
+        Integer startIndex = 0;
+        Map<String, List<String>> fieldsToMatch = Maps.newHashMap();
+        Map<String, Constants.SortOrder> sortInstructions = Maps.newHashMap();
+        Map<String, AbstractFilterInstruction> filterInstructions = Maps.newHashMap();
+        fieldsToMatch.put(TYPE_FIELDNAME, Collections.singletonList(EVENT_TYPE));
+        sortInstructions.put(DATE_FIELDNAME, Constants.SortOrder.DESC);
+        ZonedDateTime now = ZonedDateTime.now();
+        ZonedDateTime threeDaysbehind = now.plusDays(-365);
```
bh412, line 104: These are wrong and useful for testing

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -943,24 +946,53 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
 
             List<SegueScheduledJob> configuredScheduledJobs = new ArrayList<>(Arrays.asList(PIISQLJob, cleanUpOldAnonymousUsers,
-                    cleanUpExpiredReservations, deleteEventAdditionalBookingInformation, deleteEventAdditionalBookingInformationOneYearJob));
+                    cleanUpExpiredReservations));
 
             if (mailjetKey != null && mailjetSecret != null) {
                 configuredScheduledJobs.add(syncMailjetUsers);
             }
 
+            if (eventPrePostEmails != null) {
```
jsharkey13, line 980: Maybe this should be `eventPrePostEmails != null && Boolean.parseBoolean(eventPrePostEmails);`? Then we can leave the key in but set it to false too, and if set it has to be explicitly `true` or `false`  for the API to even start at all.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -943,24 +946,53 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
+                configuredScheduledJobs.add(eventFeedbackEmail);
+            }
+
             segueJobService = new SegueJobService(configuredScheduledJobs, database);
 
             if (mailjetKey == null && mailjetSecret == null) {
                 segueJobService.removeScheduleJob(syncMailjetUsers);
             }
+
+            if (eventPrePostEmails == null) {
```
jsharkey13, line 991: Ditto: `eventPrePostEmails == null || !Boolean.parseBoolean(eventPrePostEmails);`

bh412:
Needs `EVENT_PRE_POST_EMAILS=false` adding to the segue-config.properties (true for CS)

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,170 @@
...
+        this.properties = properties;
+        this.contentManager = contentManager;
+        this.bookingManager = bookingManager;
+        this.userAccountManager = userAccountManager;
+        this.emailManager = emailManager;
+        this.pgScheduledEmailManager = pgScheduledEmailManager;
+    }
+
+    public void sendEmailForEvent(IsaacEventPageDTO event, String templateId) throws SegueDatabaseException {
+        List<DetailedEventBookingDTO> eventBookings = bookingManager.adminGetBookingsByEventId(event.getId());
```
jsharkey13, line 69: What booking statuses does this return? I fear this might also email cancelled and other non-confirmed bookings. Should we also be emailing reserved users? With the same email text?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,170 @@
...
+                                         final PgScheduledEmailManager pgScheduledEmailManager) {
+        this.properties = properties;
+        this.contentManager = contentManager;
+        this.bookingManager = bookingManager;
+        this.userAccountManager = userAccountManager;
+        this.emailManager = emailManager;
+        this.pgScheduledEmailManager = pgScheduledEmailManager;
+    }
+
+    public void sendEmailForEvent(IsaacEventPageDTO event, String templateId) throws SegueDatabaseException {
```
jsharkey13, line 68: I got confused by this method name, because it's singular and unclear who is being emailed. If (see below) we filter for all confirmed/reserved/non-cancelled bookings then the title should imply that?
`sendEmailForConfirmedBookings(...)`, `sendEmailToAttendees(...)` ?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/PgScheduledEmailManager.java
```diff
@@ -0,0 +1,67 @@
...
+        String query = "SELECT EXISTS(SELECT 1 FROM scheduled_emails WHERE email_id = ?)";
+        try (Connection conn = database.getDatabaseConnection();
+             PreparedStatement pst = conn.prepareStatement(query);
+        ) {
+            pst.setString(1, emailKey);
+            try (ResultSet results = pst.executeQuery()) {
+                if (results.next()) {
+                    return results.getBoolean(1);
+                } else {
+                    return true;
```
jsharkey13, line 42: Is this right? Don't we only follow this branch if there is no result in the ResultSet? (Which I could believe shouldn't ever be possible, but then we should log an error to be clear something is very wrong?).

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -943,24 +946,53 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
 
             List<SegueScheduledJob> configuredScheduledJobs = new ArrayList<>(Arrays.asList(PIISQLJob, cleanUpOldAnonymousUsers,
-                    cleanUpExpiredReservations, deleteEventAdditionalBookingInformation, deleteEventAdditionalBookingInformationOneYearJob));
+                    cleanUpExpiredReservations));
 
             if (mailjetKey != null && mailjetSecret != null) {
                 configuredScheduledJobs.add(syncMailjetUsers);
             }
 
+            if (Boolean.parseBoolean(eventPrePostEmails)) {
```
jsharkey13, line 980: Maybe you could do this parsing just once (i.e. make `eventPrePostEmails` itself the Boolean)?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -943,24 +946,53 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
+                configuredScheduledJobs.add(eventFeedbackEmail);
+            }
+
             segueJobService = new SegueJobService(configuredScheduledJobs, database);
 
             if (mailjetKey == null && mailjetSecret == null) {
                 segueJobService.removeScheduleJob(syncMailjetUsers);
             }
+
+            if (!Boolean.parseBoolean(eventPrePostEmails)) {
```
jsharkey13, line 991: Maybe you could do this parsing just once (i.e. make `eventPrePostEmails` itself the Boolean)?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
jsharkey13: Should the method perhaps take a list of booking statuses to use as a filter in the stream below? That might make it both more explicit what the caller wants, and force us to think about it more carefully? Or we could hard-code an explicit list initially so long as we know what we want is the same for both of these emails for now.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
jsharkey13: Perhaps to be even safer, it could be:
```
String eventPrePostEmails = properties.getProperty(EVENT_PRE_POST_EMAILS);
Boolean eventPrePostEmailsEnabled = !eventPrePostEmails.isEmpty() && Boolean.parseBoolean(eventPrePostEmails);
```
?

jsharkey13: Then Guice won't explode without the key being present in the config file, I think? (Might be worth testing if you do try that approach).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/PgScheduledEmailManager.java
bh412: Yes you're right, this is a catchall 'if for some reason it breaks don't try sending an email' - I'll change it to an error

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
bh412: Thanks for spotting this, it definitely should be filtered - I'm adding a hardcoded 'confirmed' or 'attended' list and double checking the intention with Fergus

mlt47 (commented):
There is one comment about a possible need for a lock and another about using the end date for reminder emails that is worth double-checking, but otherwise I only made minor suggestions and it looks like it'll work fine.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,175 @@
...
+            ResultsWrapper<ContentDTO> findByFieldNames = this.contentManager.findByFieldNames(
+                properties.getProperty(CONTENT_INDEX), ContentService.generateDefaultFieldToMatch(fieldsToMatch),
+                startIndex, limit, sortInstructions, filterInstructions);
+            for (ContentDTO contentResult : findByFieldNames.getResults()) {
+                if (contentResult instanceof IsaacEventPageDTO) {
+                    IsaacEventPageDTO event = (IsaacEventPageDTO) contentResult;
+                    String emailKey = String.format("%s@pre", event.getId());
+                    List<BookingStatus> bookingStatuses = Arrays.asList(BookingStatus.CONFIRMED, BookingStatus.ATTENDED);
+                    if (!pgScheduledEmailManager.scheduledEmailAlreadySent(emailKey)) {
+                        pgScheduledEmailManager.saveScheduledEmailSent(emailKey);
```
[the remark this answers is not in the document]
mlt47: You'll probably be able to decide whether or not to act from the success of just the insert, but it's the sort of thing that's worth double checking in the documentation.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -956,11 +978,22 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
             segueJobService = new SegueJobService(configuredScheduledJobs, database);
 
             if (mailjetKey == null && mailjetSecret == null) {
                 segueJobService.removeScheduleJob(syncMailjetUsers);
             }
+
+            if (!eventPrePostEmailsEnabled) {
+                segueJobService.removeScheduleJob(eventReminderEmail);
+                segueJobService.removeScheduleJob(eventFeedbackEmail);
+            }
```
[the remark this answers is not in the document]
bh412: If at some point down the line we want to stop it running, just removing the config property wouldn't work as it would still live in the qrtz_triggers / qrtz_jobs tables in the db, so it needs actually removing

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,175 @@
...
+        Map<String, List<String>> fieldsToMatch = Maps.newHashMap();
+        Map<String, Constants.SortOrder> sortInstructions = Maps.newHashMap();
+        Map<String, AbstractFilterInstruction> filterInstructions = Maps.newHashMap();
+        fieldsToMatch.put(TYPE_FIELDNAME, Collections.singletonList(EVENT_TYPE));
+        sortInstructions.put(DATE_FIELDNAME, Constants.SortOrder.DESC);
+        ZonedDateTime now = ZonedDateTime.now();
+        ZonedDateTime threeDaysAhead = now.plusDays(3);
+        DateRangeFilterInstruction
+            eventsWithinThreeDays = new DateRangeFilterInstruction(new Date(), Date.from(threeDaysAhead.toInstant()));
+        filterInstructions.put(ENDDATE_FIELDNAME, eventsWithinThreeDays);
```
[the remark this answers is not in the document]
bh412: This is true, lots of this logic is copied from the events facade / methods, which use the enddate - it may be worth investigating if they actually work properly.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,175 @@
...
+
+        try {
+            ResultsWrapper<ContentDTO> findByFieldNames = this.contentManager.findByFieldNames(
+                properties.getProperty(CONTENT_INDEX), ContentService.generateDefaultFieldToMatch(fieldsToMatch),
+                startIndex, limit, sortInstructions, filterInstructions);
+            for (ContentDTO contentResult : findByFieldNames.getResults()) {
+                if (contentResult instanceof IsaacEventPageDTO) {
+                    IsaacEventPageDTO event = (IsaacEventPageDTO) contentResult;
+                    String emailKey = String.format("%s@pre", event.getId());
+                    List<BookingStatus> bookingStatuses = Arrays.asList(BookingStatus.CONFIRMED, BookingStatus.ATTENDED);
```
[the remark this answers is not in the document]
bh412: This was a request from Fergus, they shouldn't be marked as attended before an event but if for some reason an event manager does so accidentally then they will still receive the reminded. (Similar logic follows for sending the feedback email to confirmed / attended in case they don't set them to attended right away)

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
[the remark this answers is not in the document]
jsharkey13: I suppose a comment before these removals wouldn't do any harm, to explain this and the other section like it for future reference!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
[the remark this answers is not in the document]
jsharkey13: The whole point of Quartz is that one and only one server will run each job, that's the only reason to use it and not just Cron. But you make a good point about races, since (hypothetically) the jobs could overlap.

I suppose `saveScheduledEmailSent` could return a boolean, based on whether the "insert on conflict no action" returned 1 rows or 0 rows added?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,175 @@
...
+        List<Long> ids = eventBookings.stream()
+                            .filter(DetailedEventBookingDTO -> bookingStatuses == null || bookingStatuses.contains(DetailedEventBookingDTO.getBookingStatus()))
+                            .map(DetailedEventBookingDTO::getUserBooked)
+                            .map(UserSummaryDTO::getId)
+                            .distinct().collect(Collectors.toList());
+        for (Long id : ids) {
+            try {
+                RegisteredUserDTO user = userAccountManager.getUserDTOById(id);
+                emailManager.sendTemplatedEmailToUser(user,
+                    emailManager.getEmailTemplateDTO(templateId),
```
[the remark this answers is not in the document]
jsharkey13: The underlying `getContentById` does use a cache, yes, so this is nice by not important.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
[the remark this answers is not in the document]
mlt47: Thought it might be something like that 👍

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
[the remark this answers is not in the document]
mlt47: OK sure, a short comment would be useful here to explain it.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
```diff
@@ -0,0 +1,175 @@
...
+                log.debug(String.format("Sent email to user: %s %s, at: %s", user.getGivenName(), user.getFamilyName(), user.getEmail()));
+            } catch (NoUserException e) {
+                log.error(String.format("No user found with ID: %s", id));
+            } catch (ContentManagerException e) {
+                log.error("Failed to add the scheduled email sent time: ", e);
+            }
+        }
+    }
+
+    public void sendReminderEmails() {
```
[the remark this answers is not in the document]
mlt47: Happy for this not to get actioned

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/EventNotificationEmailManager.java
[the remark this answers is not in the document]
bh412: Just using the `saveScheduledEmailSent` and having it return a boolean has been used

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/PgScheduledEmailManager.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param database
+     *            - the database reference used for persistence.
+     */
+    @Inject
+    public PgScheduledEmailManager(final PostgresSqlDb database) {
+        this.database = database;
+    }
+
+    public boolean saveScheduledEmailSent(String emailKey) throws SegueDatabaseException {
```
mlt47, line 32: Perhaps  `commitToSchedulingEmail(final String emailKey)` would be a more descriptive method name.

PR #438, opened 2022-04-19
Avoid repeated database queries for deleted assignment owners

jsharkey13:
The old method would query the database again and again for users who had been deleted, since we did not cache the "not found" null case.
Another option would be to add null values to the HashMap and use the "containsKey" method instead?

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2021 Raspberry Pi Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -27,7 +27,9 @@ import uk.ac.cam.cl.dtg.isaac.dto.users.UserSummaryDTO;
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 public class AssignmentService {
     private static final Logger log = LoggerFactory.getLogger(AssignmentService.class);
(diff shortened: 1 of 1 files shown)


jsharkey13:
I'm unsure if this is the "best" way to do this, but there are hundreds of errors for a prolific user who has been deleted and the assignments left behind. But for each one we go back to the database, since the error was not cached.

This approach seemed more obvious and safer, but it is less efficient than adding null values to the HashMap and using `containsKey` and not `get` to check if the owner has already been looked up. Would the other method be preferable?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
```diff
@@ -39,24 +41,25 @@ public AssignmentService(final UserAccountManager userManager) {
         this.userManager = userManager;
     }
 
-    public <T extends IAssignmentLike> void augmentAssignerSummaries(Collection<T> assignments) throws SegueDatabaseException {
+    public <T extends IAssignmentLike> void augmentAssignerSummaries(final Collection<T> assignments) throws SegueDatabaseException {
         Map<Long, UserSummaryDTO> userSummaryCache = new HashMap<>();
 
+        // Iterating over the owner IDs allows us to cache "no user found" errors without querying database each time:
+        List<Long> ownerUserIds = assignments.stream().map(IAssignmentLike::getOwnerUserId).distinct().collect(Collectors.toList());
```
Morpheu5, line 48: Does `distinct()` take care of the `null`s?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
```diff
@@ -39,24 +41,25 @@ public AssignmentService(final UserAccountManager userManager) {
...
+        for (Long ownerUserId: ownerUserIds) {
+            try {
```
Morpheu5, line 50: This doesn't try to get the summary from the cache anymore. Is this correct?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
jsharkey13: I'd not even considered that there might be `null` values. It shouldn't be possible for owner IDs to become null because there's a foreign key in Postgres with `ON DELETE CASCADE` (and we only soft-delete users anyhow now so a user ID can never vanish).

But I could add a `.filter(Objects::nonNull)` easily enough?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
jsharkey13: Erm, yeah, because of the `.distinct()` each user will only be looked up once, and this is how we populate the cache to use later when we loop by assignment (where there are lots of duplicated owner users a chache is good for).

I think this diff is best viewed separately, if I am honest; the combined diff muddles it all up horribly!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
Morpheu5: Seeing there is a null check in the removed piece of code, I thought maybe that value was somehow nullable, or there was a NPE somewhere upstream.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/services/AssignmentService.java
Morpheu5: Yeah, that does make it a bit more readable indeed…

PR #440, opened 2022-04-27
Update dependencies, April 2022

Morpheu5:
Brought some dependencies up to date to fix a critical severity security issue and a few high severity issues.

According to Dependency Check, there remains one high severity security issue with `asyncutils` which is a dependency of `resteasy` but no newer versions of `asyncutils` exist so it's possible `resteasy` 5 has moved on to use something else, and seems to introduce breaking changes so we may want to take a look at this in the next update cycle.

~There remains an odd issue with ETL being broken by something that I can't identify. The issue is JSch detecting an invalid private key, but I can't see anything wrong with it according to the portion of the Internet that I surveyed.~
The issue with ETL was due to JSch not liking certain types of SSH keys (ECDSA and EdDSA at least) so @mwtrew kindly provided a replacement.

# Direct dependencies

- RESTEasy `4.7.0.Final` => `4.7.6.Final`
- OAuth Client `1.31.0` => `0.33.3`
- Jetty `9.4.42.v20210604` => `9.4.46.v20220331`
- [NEW] OWASP Dependency Check
- [NEW] OSS Index

# Indirect dependencies

- Maven Core pinned to `3.8.5`
- gson pinned to `2.9.0`
- google-http-client-gson pinned to `1.41.8`
- okhttp pinned to `3.12.13`
- netty pinned to `4.1.76.Final`
- jackson-dataformat pinned to `2.13.2`


pom.xml
```diff
@@ -12,18 +12,19 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <segue.version>v2.21.4-SNAPSHOT</segue.version>
         <log4j.version>2.17.1</log4j.version>
-        <resteasy.version>4.7.0.Final</resteasy.version>
+        <resteasy.version>4.7.6.Final</resteasy.version>
         <guice.version>4.2.3</guice.version>
-        <oauth-client.version>1.31.0</oauth-client.version>
+        <oauth-client.version>1.33.3</oauth-client.version>
         <jackson.version>2.13.2</jackson.version>
         <jackson-databind.version>2.13.2.2</jackson-databind.version>
```
jsharkey13, line 19: Both of these can go to be `2.13.3`.

pom.xml
```diff
@@ -422,12 +423,23 @@
...
         </dependency>
-<!--        <dependency>-->
-<!--            <groupId>org.owasp</groupId>-->
-<!--            <artifactId>dependency-check-maven</artifactId>-->
-<!--            <version>7.0.4</version>-->
-<!--        </dependency>-->
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+        </dependency>
```
jsharkey13, line 429: This is new? And it doesn't have a version number? But below, in the pinning section, it is listed with a version number?!

pom.xml
```diff
@@ -422,12 +423,23 @@
...
 
+        <dependency>
+            <groupId>org.owasp</groupId>
+            <artifactId>dependency-check-maven</artifactId>
+            <version>7.1.0</version>
```
jsharkey13, line 434: This might get out of sync with the same version number below? A variable?

pom.xml
```diff
@@ -436,7 +448,12 @@
             <plugin>
                 <groupId>org.owasp</groupId>
                 <artifactId>dependency-check-maven</artifactId>
-<!--                <version>${dependency-check-maven.version}</version>-->
+                <version>7.1.0</version>
```
jsharkey13, line 451: See note above about a variable?

pom.xml
```diff
@@ -562,6 +579,76 @@
                 <artifactId>commons-codec</artifactId>
                 <version>1.15</version>
             </dependency>
+
```
jsharkey13, line 582: Maybe you could add a comment above this dependencyManagement block to say what it is?

Morpheu5: I am honestly not sure what this is and how it got here because it's not an indirect dependency unless it was one of some package that we stopped using recently…

Morpheu5: Ah, turns out this is a dependency of RESTEasy and it looks like there's no need to pin the version. Not sure why it ended up in the `pom.xml` file.

PR #442, opened 2022-05-04
Initialise Isaac on Jetty start not first HTTP request

jsharkey13:
Turns out that it had always been this simple and for years we could have fixed this issue with so little effort :upside_down_face:

This change means that the Isaac Application initialises its classes on Jetty startup and does not wait for the first matching HTTP request to do so. This means the API will (finally) start by itself without needing to be poked or prompted.


mlt47 (approved):
Works for me!

PR #443, opened 2022-05-06
Handle multiple assignments set in one request

chrisjpurdy:
This adds a new endpoint, and modifies an existing one. After the front-end change for this, the old endpoint will not be in use, and it's use case is subsumed by the new one anyway.

New DTOs are added to send back to the user info on which (if any) assignments to groups fail.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -31,9 +31,7 @@ import uk.ac.cam.cl.dtg.isaac.api.managers.AssignmentManager;
 import uk.ac.cam.cl.dtg.isaac.api.managers.DuplicateAssignmentException;
 import uk.ac.cam.cl.dtg.isaac.api.managers.GameManager;
 import uk.ac.cam.cl.dtg.isaac.api.services.AssignmentService;
-import uk.ac.cam.cl.dtg.isaac.dto.AssignmentDTO;
-import uk.ac.cam.cl.dtg.isaac.dto.GameboardDTO;
-import uk.ac.cam.cl.dtg.isaac.dto.GameboardItem;
+import uk.ac.cam.cl.dtg.isaac.dto.*;
 import uk.ac.cam.cl.dtg.segue.api.managers.GroupManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.QuestionManager;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
@@ -45,8 +43,6 @@ import uk.ac.cam.cl.dtg.segue.dao.SegueDatabaseException;
 import uk.ac.cam.cl.dtg.segue.dao.content.ContentManagerException;
 import uk.ac.cam.cl.dtg.isaac.dos.LightweightQuestionValidationResponse;
 import uk.ac.cam.cl.dtg.isaac.dos.QuestionValidationResponse;
-import uk.ac.cam.cl.dtg.isaac.dto.SegueErrorResponse;
-import uk.ac.cam.cl.dtg.isaac.dto.UserGroupDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.QuestionDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.users.RegisteredUserDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.users.UserSummaryDTO;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -939,105 +935,149 @@ public Response getAssignedGroupsByGameboards(@Context final HttpServletRequest
...
+                return new SegueErrorResponse(Status.FORBIDDEN, "You need a staff account to set assignments to more than one group at once!").toResponse();
             }
 
-            if (!GroupManager.isOwnerOrAdditionalManager(assigneeGroup, currentlyLoggedInUser.getId())
-                    && !isUserAnAdmin(userManager, currentlyLoggedInUser)) {
-                return new SegueErrorResponse(Status.FORBIDDEN,
-                        "You can only set assignments to groups you own or manage.").toResponse();
-            }
+            // Since all of the assignment notes in a single query are currently the same, we only have to do note size
+            // checks once, and we only need to fetch the gameboard once
```
jsharkey13, line 972: This seems unlikely to be the case generally, since there's no guarantee the request is coming from the frontend. I think the notes for each assignment need to be checked inside the loop.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -939,105 +935,149 @@ public Response getAssignedGroupsByGameboards(@Context final HttpServletRequest
...
                 // user is not staff but it is a teacher, if we got here unscathed
                 return new SegueErrorResponse(Status.BAD_REQUEST, "You are not allowed to add assignment notes.").toResponse();
             }
 
-            GameboardDTO gameboard = this.gameManager.getGameboard(assignmentDTOFromClient.getGameboardId());
+            if (firstAssignment.getGameboardId() == null) {
+                return new SegueErrorResponse(Status.BAD_REQUEST, "A required field was missing. Must provide gameboard id.").toResponse();
+            }
+
+            GameboardDTO gameboard = this.gameManager.getGameboard(firstAssignment.getGameboardId());
```
jsharkey13, line 991: What if each AssignmentDTO has a different gameboard? Again just because the frontend won't let you doesn't mean it won't happen.

jsharkey13: This validation is important for the later loop not to fail, so this also needs to be inside the loop (although I'd recommend a cache of valid gameboards to avoid looking up the same one again and again).

jsharkey13 (changes requested):
I think this assumes too much about the state of incoming requests. For security reasons we need to thoroughly validate each item in the submitted list, and conveniently this gives us the added bonus of flexibility for different groups/notes/etc later on should we want it.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -939,105 +935,149 @@ public Response getAssignedGroupsByGameboards(@Context final HttpServletRequest
...
      * @param request
      *            - so that we can identify the current user.
-     * @param assignmentDTOFromClient a partially completed DTO for the assignment.
-     * @return the assignment object.
+     * @param assignmentDTOsFromClient a list of partially completed DTO(s) for the assignment(s).
+     * @return a list of ids of successful assignments, and a list of failed (an AssignmentSettingResponseDTO)
      */
     @POST
-    @Path("/assign/")
+    @Path("/assign_many")
```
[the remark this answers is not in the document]
chrisjpurdy: I'm easy, so may as well make it consistent with REST nomenclature 👍

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
chrisjpurdy: To be honest, I didn't think about the fact that any old garbage could be in the gameboard and notes part of the other `AssignmentDTO`s, so yeah good spot.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -939,105 +941,152 @@ public Response getAssignedGroupsByGameboards(@Context final HttpServletRequest
...
+    /**
+     * Allows a user to assign a gameboard to group of users.
+     *
+     * @param request
+     *            - so that we can identify the current user.
+     * @param assignmentDTOFromClient a partially completed DTO for the assignment.
+     * @return an AssignmentSettingResponseDTO (see assignGameBoards)
+     */
+    @POST
+    @Path("/assign/")
```
Morpheu5, line 1081: If I understand this correctly, this endpoint is going to go in a few releases time, right? So, should we just mark this as deprecated so we remember to remove it in the next release or something like that?

Morpheu5 (commented):
Aside from that one comment, this looks sensible. I haven't tested this yet.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
jsharkey13: I don't see why we would deprecate this? Especially since this single-assignment request will still be the majority of cases! There's no additional maintenance burden in maintaining it or anything, so why remove it?

chrisjpurdy: In the frontend, setting a single assignment sends a singleton list to the multiple assignment endpoint, so nowhere in the frontend actually sends a single-assignment request. The only reason I can think of keeping this endpoint is so the API is more flexible, but the frontend definitely isn't using this flexibility at the moment.

chrisjpurdy: I feel like it's best to deprecate code whose functionality is subsumed by other code, just so that there's less redundant stuff when browsing the codebase.

Morpheu5: That's what I meant, yes. I couldn't find any mention of `/assign/` in the front-end and, given the singleton list use case, I thought deprecating the old endpoint reduced maintenance and attack surface.

jsharkey13: Okay, fine. I think it's no longer actually the same endpoint anyway, since the value it returns is now very different! It's a breaking change.
Let's leave it marked this way.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -939,105 +941,154 @@ public Response getAssignedGroupsByGameboards(@Context final HttpServletRequest
...
+                } catch (DuplicateAssignmentException e) {
+                    assigmentErrors.add(new AssignmentErrorDTO(assignmentDTO.getGroupId(), e.getMessage()));
+                } catch (SegueDatabaseException e) {
+                    log.error("Database error while trying to assign work", e);
+                    assigmentErrors.add(new AssignmentErrorDTO(assignmentDTO.getGroupId(), "Unknown database error."));
                 }
             }
-
-            return Response.ok(assignmentDTOFromClient).build();
+            return Response.ok(new AssignmentSettingResponseDTO(assigmentSuccessfulGroupIds, assigmentErrors)).build();
```
jsharkey13, line 1067: I don't think this list of group IDs is enough to correctly tell the user which assignments have been set and which have failed. Assume that all the assignments submitted are to the same group, but some are duplicates and some gamebaords don't exist; only one group ID will appear some number of times in this list, and there will be some number of errors in the other list.

Is that really enough information to inform the user what has actually happened and what has not? I've not looked at the frontend code to see the feedback it gives the user, so if it _is_ enough, then that's okay!

jsharkey13: I think an ordered list of statuses, one per submitted assignment, would work otherwise. (Like a list of `AssignmentStatusDTO` objects, instead of a wrapper object overall, and instead of an error object, that has `assignmentId` non-null if the assignment was set, and `errorMessage` non-null if it failed?).

But if there's enough as-is to do clear feedback, then maybe stick with this.

jsharkey13 (commented):
I changed some names and changed the comment about computeIfAbsent (it can't be used because the method we need there throws a checked exception).

I still have one question about feedback and knowing what has actually been done by the request.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
chrisjpurdy: Yeah having one feedback object per assignment/group pair would be better, I'll make that change tomorrow, it should take about 30 mins so it's worth it for more sensible code. 
As the frontend is currently, you can't set multiple assignments to a single group, only a single assignment to multiple groups - the error objects and list of "successful" ids tell you all you need to know for useful feedback.

PR #444, opened 2022-05-09
Do not mangle URL email template parameters

jsharkey13:
I noticed that verification URLs in our emails were ending up like: `/verifyemail?userid=XXXXXX&amp;token=YYYYYYY` with an encoded `&amp;` rather than the raw ampersand. In the plain-text versions of the email this prevents the link working at all. In the HTML emails, it _should_ be the case that the browser fixes this, but it is not obvious that it will happen.

We don't need to be so zealous at sanitising email parameters whose names end `URL` since they are allowed to be HTML-like.


bh412 (approved):
Concise and a sensible alteration

PR #445, opened 2022-05-09
Replace outdated JSch for etl

Morpheu5:
JSch doesn't like ECDSA keys so we are replacing it with something that does like them.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/database/GitDb.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/database/GitDb.java
@@ -31,11 +31,19 @@ import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.api.errors.NoHeadException;
 import org.eclipse.jgit.errors.RevisionSyntaxException;
-import org.eclipse.jgit.lib.*;
+import org.eclipse.jgit.lib.Constants;
+import org.eclipse.jgit.lib.ObjectId;
+import org.eclipse.jgit.lib.ObjectLoader;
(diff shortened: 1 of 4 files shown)


Morpheu5 (commented):
Looks good, tests pass, ETL works for me once again, and we swapped an old library with one that is more up to date. Win/win/win/win.

src/main/java/uk/ac/cam/cl/dtg/segue/database/GitDb.java
```diff
@@ -374,8 +345,14 @@ protected JSch getJSch(final OpenSshConfig.Host hc, final org.eclipse.jgit.util.
                     log.info("Fetched latest from git. Latest version is: " + this.getHeadSha());
                 }
             }
-
-
+        } catch (TransportException e) {
+            log.error("Failed to authenticate with the remote content repository via SSH. Ensure the 'Git' section of "
+                    + "segue-config[.cs].properties has valid values, particularly that the key at "
```
jsharkey13, line 350: There's actually a `getPropertiesFile()` method of the Properties object that will give you the file path for the actual in-use file which may be more useful than this generic name?

jsharkey13: But if not, I'd prefer that the `[.cs]` bit wasn't there, since the API isn't supposed to know about either physics or CS where possible. Just `segue.properties` is fine.

jsharkey13 (commented):
One very minor thing, that isn't really that bad.

jsharkey13 (approved):

PR #447, opened 2022-05-10
Better Caching Headers

jsharkey13:
If a response is explicitly marked as not-public, or has a cache header added but is not designated public; mark it as private. This should help cases where man-in-the-middle monitoring software acts as a HTTP cache and introduces security issues.

Technically this change may prevent the altered endpoints that may otherwise have been cached from being cached (due to
the max-age=0 bit), but this is fine for these cases.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
@@ -43,10 +43,7 @@ import javax.ws.rs.core.Response.Status;
 import java.util.Collections;
 import java.util.List;
 
-import static uk.ac.cam.cl.dtg.segue.api.Constants.BooleanOperator;
-import static uk.ac.cam.cl.dtg.segue.api.Constants.CONTENT_INDEX;
-import static uk.ac.cam.cl.dtg.segue.api.Constants.DEFAULT_RESULTS_LIMIT;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
```diff
@@ -43,10 +43,7 @@
 import java.util.Collections;
 import java.util.List;
 
-import static uk.ac.cam.cl.dtg.segue.api.Constants.BooleanOperator;
-import static uk.ac.cam.cl.dtg.segue.api.Constants.CONTENT_INDEX;
-import static uk.ac.cam.cl.dtg.segue.api.Constants.DEFAULT_RESULTS_LIMIT;
-import static uk.ac.cam.cl.dtg.segue.api.Constants.TYPE_FIELDNAME;
+import static uk.ac.cam.cl.dtg.segue.api.Constants.*;
```
Morpheu5, line 46: !

Morpheu5 (approved):

src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
jsharkey13: This is the preferred import style for the two constants classes, static and starred.

PR #448, opened 2022-05-12
Clone IsaacEventPageDTO objects to avoid cache poisoning

jsharkey13:
We mutate the object returned from this method to add booking information; but since all that is returned is a reference to the DTO in the ContentManager cache, we thus mutate the cached object too! When we the use this cached object for subsequent requests, we risk returning outdated data.
Instead, use the mapper to deep-copy the DTO.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
@@ -1721,12 +1721,10 @@ public class EventsFacade extends AbstractIsaacFacade {
 
             try {
                 RegisteredUserDTO user = userManager.getCurrentRegisteredUser(request);
-                page.setUserBooked(this.bookingManager.isUserBooked(page.getId(), user.getId()));
-                page.setUserOnWaitList(this.bookingManager.hasBookingWithStatus(page.getId(), user.getId(), BookingStatus.WAITING_LIST));
                 page.setUserBookingStatus(this.bookingManager.getBookingStatus(page.getId(), user.getId()));
             } catch (NoUserLoggedInException e) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -1718,12 +1721,10 @@ private IsaacEventPageDTO augmentEventWithBookingInformation(final HttpServletRe
 
             try {
                 RegisteredUserDTO user = userManager.getCurrentRegisteredUser(request);
-                page.setUserBooked(this.bookingManager.isUserBooked(page.getId(), user.getId()));
-                page.setUserOnWaitList(this.bookingManager.hasBookingWithStatus(page.getId(), user.getId(), BookingStatus.WAITING_LIST));
                 page.setUserBookingStatus(this.bookingManager.getBookingStatus(page.getId(), user.getId()));
             } catch (NoUserLoggedInException e) {
                 // no action as we don't require the user to be logged in.
-                page.setUserBooked(null);
+                page.setUserBookingStatus(null);
```
Morpheu5, line 1727: I assume this (and the other related changes) is OK because we don't use these properties anymore in the front-end, right?

Morpheu5 (approved):

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
jsharkey13: Ah, my commit message was very clear that I had checked this, but the PR doesn't say it. Yes, neither of these properties appear in the frontend types and they are not used by name anywhere either!

mwtrew (approved):

PR #453, opened 2022-06-14
Send tags to frontend indicating numeric question answer is invalid format

chrisjpurdy:
And move feedback text to frontend

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
@@ -208,12 +208,13 @@ public class IsaacNumericValidator implements IValidator {
         } catch (NumberFormatException e) {
             log.debug("Validation failed for '" + answerFromUser.getValue() + " " + answerFromUser.getUnits() + "': "
                     + "cannot parse as number!");
-
-            String feedback = "Your answer is not in a format we recognise, please enter your answer as a decimal number.";
+            HashSet<String> responseTags = new HashSet<>(ImmutableList.of("nq_format"));
             if (answerFromUser.getValue().matches(INVALID_NEGATIVE_STANDARD_FORM)) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -208,12 +208,13 @@ public final QuestionValidationResponse validateQuestionResponse(
         } catch (NumberFormatException e) {
             log.debug("Validation failed for '" + answerFromUser.getValue() + " " + answerFromUser.getUnits() + "': "
                     + "cannot parse as number!");
-
-            String feedback = "Your answer is not in a format we recognise, please enter your answer as a decimal number.";
```
jsharkey13, line 212: I think we should still keep this feedback for this error and not use the default validation response as you've changed to below.
Given you override it in the frontend now, it doesn't matter so much, but it seems silly to add a less useful error message instead of this one.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -208,12 +208,13 @@ public final QuestionValidationResponse validateQuestionResponse(
...
+            HashSet<String> responseTags = new HashSet<>(ImmutableList.of("nq_format"));
             if (answerFromUser.getValue().matches(INVALID_NEGATIVE_STANDARD_FORM)) {
-                feedback += "<br>When writing standard form, you must include a `^` between the 10 and the exponent.";
```
jsharkey13, line 214: (But you can leave this line out now, and just add the tag, I think that's fine).

jsharkey13 (commented):
This is good, but I don't think the tags should be so terse/cryptic. I think `unrecognised_format` and `invalid_std_form` would be clearer.

jsharkey13 (approved):

PR #460, opened 2022-07-01
Use separate web.xml files for local and production?

jsharkey13:
Remove localhost from the CORS domains for production, which will prevent the development version of the content editor from accessing staging but tightens up the allowed origins.
@mlt47 - will this be okay? Ideally the editor running on localhost would use localhost for the tags/glossary queries it currently makes to staging, but I know that it is simpler to not need an API locally to develop the editor.

Allow any 4-digit localhost port to access the local version of the API. The local configuration is a copy of the old API one but with more ports allowed on localhost.

---

If we don't think it's important enough to warrant separate configuration (and I do wonder whether listing localhost in production _really_ matters, because anyone making CORS requests from a localhost site is running software that could bypass CORS anyhow), then I can revert to a single API file but leave this neatened format for `pom.xml` file.


Morpheu5 (approved):

web-api-local.xml
```diff
@@ -0,0 +1,56 @@
...
+    <filter>
+        <filter-name>cross-origin</filter-name>
+        <filter-class>org.eclipse.jetty.servlets.CrossOriginFilter</filter-class>
+        <init-param>
+            <param-name>allowedMethods</param-name>
+            <param-value>GET,POST,DELETE,PUT,HEAD</param-value>
+        </init-param>
+        <init-param>
+            <param-name>allowedOrigins</param-name>
+            <param-value>(https://*.isaac(physics|computerscience).org|http://localhost:[0-9]{4})</param-value>
```
mlt47, line 14: How about `[0-9]+` for the ports (or `\d+`) as I don't think there is a particular reason for it to enforce 4 digits.

mlt47 (approved):
I don't mind whether you go one config or separate config. One seems simpler.
Easy enough to edit the editor to point locally if you choose the other option though so don't let that stop you.

PR #466, opened 2022-07-21
Get rid of `getAssignedGroupsByGameboard`

chrisjpurdy:
The endpoint was horrendously slow if used with many gameboard ids, and its only use is in "Set Assignments" which should be populated using the `getAssigned` endpoint instead.

**Should not be merged** without the frontend PR that changes which endpoint "Set Assignments" uses (will be linked to here once it is done)

## Merge only AFTER these two:

- https://github.com/isaacphysics/isaac-api/pull/467
- https://github.com/isaacphysics/isaac-react-app/pull/610


jsharkey13 (approved):
This seems fine, but requires https://github.com/isaacphysics/isaac-api/pull/467 and https://github.com/isaacphysics/isaac-react-app/pull/610 to be merged first or things will stop working.

Morpheu5 (approved):

PR #467, opened 2022-07-21
Augment returned assignments from `/assignments/assign` with group name

chrisjpurdy:


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -254,8 +254,10 @@ public Response getAssigned(@Context final HttpServletRequest request,
 
             if (null == groupIdOfInterest) {
                 List<UserGroupDTO> allGroupsOwnedAndManagedByUser = this.groupManager.getAllGroupsOwnedAndManagedByUser(currentlyLoggedInUser, false);
-                return Response.ok(this.assignmentManager.getAllAssignmentsForSpecificGroups(allGroupsOwnedAndManagedByUser))
-                        .cacheControl(getCacheControl(NEVER_CACHE_WITHOUT_ETAG_CHECK, false)).build();
+                Map<Long, String> groupIdToName = allGroupsOwnedAndManagedByUser.stream().collect(Collectors.toMap(UserGroupDTO::getId, UserGroupDTO::getGroupName));
+                List<AssignmentDTO> assignments = this.assignmentManager.getAllAssignmentsForSpecificGroups(allGroupsOwnedAndManagedByUser);
+                assignments.stream().forEach(assignment -> assignment.setGroupName(groupIdToName.get(assignment.getGroupId())));
+                return Response.ok(assignments).cacheControl(getCacheControl(NEVER_CACHE_WITHOUT_ETAG_CHECK, false)).build();
```
jsharkey13, line 260: I don't think the Facade is the right place to do this augmenting. It might also be useful elsewhere, and so I think it can live inside the manager method `getAllAssignmentsForSpecificGroups` in `AssignmentManager.java`.

E.g. in that method have:

```java
        Map<Long, String> groupIdToName = groups.stream().collect(Collectors.toMap(UserGroupDTO::getId, UserGroupDTO::getGroupName));
        List<AssignmentDTO> assignments = this.assignmentPersistenceManager.getAssignmentsByGroupList(groupIdToName.keySet());
        assignments.forEach(assignment -> assignment.setGroupName(groupIdToName.get(assignment.getGroupId())));
        return assignments;
```

I did look at where else used this method, and it's not much of an overhead for them. Using the keySet avoids another stream, and I think works because we just need a Collection and not a list.

chrisjpurdy: Done!

jsharkey13 (approved):

PR #470, opened 2022-07-26
Remove version parameter from content manager methods

mwtrew:
These methods still use the SHA to build cache keys, but ES queries use version aliases (live, latest) exclusively.

Other changes:
 * Removes IContentManager interface and replaces references with the implementation (GitContentManager)
 * Removes unused methods required by the interface

---

**Pull Request Check List**
- [ ] Peer-Reviewed


Morpheu5 (approved):
Seems to work fine and the changes are relatively straightforward.

jsharkey13 (approved):
This looks fine. One thing I note is that we've lost a great deal of the JavaDoc that used to be in `IContentManager` and was inherited by the `GitContentManager`; now there's no inheritance but the JavaDoc hasn't been copied over.

I don't know how I feel about _having_ to have JavaDoc for every public method, but it is the only thing to note before we go with this.

jsharkey13:
I've gone back and reviewed the old JavaDoc, and it's not worth copying. We can write better stuff as a future project, but let's not delay this and make merging it a pain later!

PR #471, opened 2022-07-29
Update to Jetty 11 and Jakarta namespace

mwtrew:
Also updates some other dependencies to resolve incompatibilities and removes those with `javax` dependencies (as far as possible).

A few notes:
 * The latest available version of Guice has not moved to the `jakarta` namespace and still has `javax` dependencies, but this doesn't seem to cause any problems.
 * Orika produces a "reflective access warning" as described [here](https://github.com/orika-mapper/orika/issues/280), however this appears to be advisory and doesn't have any effect (until the next Java upgrade, when it will stop working). 
 * Apparently there is no `jakarta` equivalent to the `@Immutable` annotation, which was only used in one place and is effectively documentation, so I've just removed that.

---

**Pull Request Check List**
- [ ] Peer-Reviewed


pom.xml
```diff
@@ -389,30 +354,46 @@
...
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+            <version>3.0.2</version>
+        </dependency>
+
+        <dependency>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <version>2.1.0</version>
         </dependency>
```
jsharkey13, line 397: Do we need to explicitly add these dependencies now? Previously they were transitively provided by Jetty?

pom.xml
```diff
@@ -250,10 +210,15 @@
...
-            <version>3.6</version>
+            <version>4.3</version>
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.7</version>
+        </dependency>
```
jsharkey13, line 221: Why have we gained another logging framework as a dependency? Is it that we can no longer transitively depend on it via the log4j interface to slf4j?

mwtrew:
> This has totally broken Swagger, and I've spent an hour looking but am unsure how to proceed.
I didn't notice that, will take a look.

pom.xml
mwtrew: We used to get this transitively from `jgit`. The new version of `jetty-servlets` brings in version 2.x which supersedes that and spews forth error messages when used with `log4j` at the moment due to [this](https://issues.apache.org/jira/browse/LOG4J2-3139).

Another option would be excluding `slf4j-api` from `jetty-servlets`, so the `jgit`-imported version is effective once again. Both approaches seem to be practically equivalent, but I think the explicit import is preferable.

pom.xml
mwtrew: Looks like you're right, updated.

PR #475, opened 2022-08-09
Safely remove Quartz jobs

jsharkey13:
If the `SegueGuiceConfigurationModule` attempts to remove jobs on servlet startup on a standby server with a read-only database, it will fail and in doing so cause Guice to fail to provision, breaking the entire servlet.

Instead, keep a list of jobs that should be removed and only attempt to remove them when attempting to start the scheduler (this is done if the database is not read-only, or when an ADMIN hits the endpoint to start Quartz).

This might be easiest to review commit by commit.


src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -1011,18 +1010,18 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
+                scheduledJobsToRemove.add(syncMailjetUsers);
             }
 
             if (!eventPrePostEmailsEnabled) {
-                segueJobService.removeScheduleJob(eventReminderEmail);
-                segueJobService.removeScheduleJob(eventFeedbackEmail);
+                scheduledJobsToRemove.add(eventReminderEmail);
+                scheduledJobsToRemove.add(eventFeedbackEmail);
             }
+            segueJobService = new SegueJobService(database, configuredScheduledJobs, scheduledJobsToRemove);
```
mlt47, line 1024: Just a thought: It's a bit annoying that Quartz not working meant that the API didn't work.
Should we have a try catch around this rather than throwing the SchedulerException?
In my head they are separate services that should be monitored separately but that does increase the complexity of the architecture a little.

mlt47 (approved):
This looks like it should work.

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -1011,18 +1010,18 @@ private static SegueJobService getSegueJobService(final PropertiesLoader propert
...
             }
 
-            segueJobService = new SegueJobService(configuredScheduledJobs, database);
-
-            // Simply removing the following jobs from the configuredScheduledJobs wouldn't remove them from the qrtz_job_details
-            // page so they need properly removing.
+            // Simply removing jobs from configuredScheduledJobs won't de-register them if they
+            // are currently configured, so the constructor takes a list of jobs to remove too.
+            List<SegueScheduledJob> scheduledJobsToRemove = new ArrayList<>();
             if (mailjetKey == null && mailjetSecret == null) {
```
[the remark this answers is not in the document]
jsharkey13: Ah, yes, that was what I wanted when Ben added it, but they couldn't be because the scheduler needed to have been created between the first if and this one. Now that none of it is being done synchronously this should be possible!

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
jsharkey13: It must be said that no other provider in the GuiceConfigurationModule looks to throw a checked exception. Both Postgres and ElasticSearch throw unchecked exceptions, which have the same end effect.

I suppose we could catch and return null. And UptimeRobot does monitor the Quartz status, so we probably would notice in that case. We don't explicitly use the provider method, only via `injector.getInstance(SegueJobService.class);`, and so returning `null` (which other providers do when they fail to init) seems safe enough.

OTOH, Quartz is now critical to the normal function of the website, so perhaps failing hard and fast is safer overall?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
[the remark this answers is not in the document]
jsharkey13: This technically changes the logic slightly (added if both non-null, removed if both null ==> added if both non-null, removed if either null), but that seems safest, since we can't actually run the job if either config entry is null anyhow!

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
mlt47: Let's leave it as is for now 👍

PR #476, opened 2022-08-15
Allow tests to be submitted up to midnight of the due date

chrisjpurdy:


jsharkey13 (approved):
I feel like, if the due date is a timestamp, not a date with no time, then it's meant to be a timestamp; i.e. mysteriously rounding it to midnight might be quite confusing.
But this is fine for our use-case, I think. Perhaps in an ideal world the JavaDoc of `QuizAssignmentDO` and `QuizAssignmentDTO` could note that elsewhere we assume that the due date has no time and is only a day - but I doubt anyone would read it even if it did say that.

PR #477, opened 2022-08-15
Unlock code that prevents users from attempting a question through the regular question attempt endpoint if the question is part of a quiz

Morpheu5:
**Pull Request Check List**
- [ ] Peer-Reviewed


jsharkey13 (approved):
This seems to work as expected, and it's been working fine on non-production sites for some time.

Annoyingly, we don't already have the question object cached because loading a question page uses `findSingleResult` and not `getContentDOById` (because it mutates the page with attempt information). But the overhead isn't too great, and I can't think of circumstances where a question page might vanish without the part also vanishing.

PR #479, opened 2022-08-18
Add validation for due date/scheduled start date combination

mwtrew:
Other than the change referenced in the title, this PR involves some fairly significant changes to how tests work, which I will summarise here (see [this commit](https://github.com/isaacphysics/isaac-api/pull/479/commits/3e0285bb8ac02bebf025e6ecc4a384eee87164e8) for more detail):
 * Introduces [Failsafe](https://maven.apache.org/surefire/maven-failsafe-plugin/) which runs integration tests as part of the Maven `verify` target. The `test` target now only runs unit tests.
     * This requires integration tests to be named `*IT.java` hence the re-naming of existing tests.
     * Configures GitHub Actions to use the `verify` target instead of `test`, so integration tests are still run.
 * Integration tests are run in sequence and containers are re-created per test class, giving them more isolation across tests and lower peak memory usage. This also makes them a bit slower.
 * Unit tests re-use forked JVM instances making them faster.
 * Adds the integration test config path as a Failsafe option, so it doesn't need to be supplied on the command line.

Also adds some `@PowerMockIgnore`s to existing unit tests to remove some spam from the logs.

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/test/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacadeIT.java
+++ b/src/test/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacadeIT.java
@@ -79,9 +79,9 @@ public class EventsFacadeIT extends IsaacIntegrationTest {
     // DELETE /events/{event_id}/bookings/cancel -> EventsFacade::cancelBooking(request, eventId)
     public void getBookingByIdTest() throws NoCredentialsAvailableException, NoUserException, SegueDatabaseException, AuthenticationProviderMappingException, IncorrectCredentialsProvidedException, AdditionalAuthenticationRequiredException, InvalidKeySpecException, NoSuchAlgorithmException, MFARequiredButNotConfiguredException, SQLException {
         // --- Login as a student
-        LoginResult studentLogin = loginAs(httpSession, IntegrationTestConstants.TEST_STUDENT_EMAIL, IntegrationTestConstants.TEST_STUDENT_PASSWORD);
+        LoginResult studentLogin = loginAs(httpSession, ITConstants.TEST_STUDENT_EMAIL, ITConstants.TEST_STUDENT_PASSWORD);
         // --- Login as an event manager
-        LoginResult eventManagerLogin = loginAs(httpSession, IntegrationTestConstants.TEST_EVENTMANAGER_EMAIL, IntegrationTestConstants.TEST_EVENTMANAGER_PASSWORD);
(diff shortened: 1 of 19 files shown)


src/test/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacadeIT.java
```diff
@@ -38,7 +36,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-public class EventsFacadeTest extends IsaacIntegrationTest {
+public class EventsFacadeIT extends IsaacIntegrationTest {
```
Morpheu5, line 39: I'm not sure i *like* the name change, but I suppose `EventsFacadeIntegrationTest` is worse.

src/test/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacadeIT.java
```diff
@@ -81,9 +79,9 @@ public void getEventsTest() {
     // DELETE /events/{event_id}/bookings/cancel -> EventsFacade::cancelBooking(request, eventId)
     public void getBookingByIdTest() throws NoCredentialsAvailableException, NoUserException, SegueDatabaseException, AuthenticationProviderMappingException, IncorrectCredentialsProvidedException, AdditionalAuthenticationRequiredException, InvalidKeySpecException, NoSuchAlgorithmException, MFARequiredButNotConfiguredException, SQLException {
         // --- Login as a student
-        LoginResult studentLogin = loginAs(httpSession, Constants.TEST_STUDENT_EMAIL, Constants.TEST_STUDENT_PASSWORD);
+        LoginResult studentLogin = loginAs(httpSession, IntegrationTestConstants.TEST_STUDENT_EMAIL, IntegrationTestConstants.TEST_STUDENT_PASSWORD);
```
Morpheu5, line 82: Although at this point might as well call these ITConstants too…

Morpheu5 (approved):

src/test/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacadeIT.java
mwtrew: Nor do I, but [Failsafe expects this](https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html) by default - it would be possible to change that with a bit of fiddly configuration however. Looks like they do this to avoid overlapping with Surefire, which runs anything matching `*Test.java` by default.

PR #481, opened 2022-09-01
Fix caching of available quizzes/tests

jsharkey13:
Prior to this change, tests were hard cached for an hour, always.

I swapped it to NEVER_CACHE_WITHOUT_ETAG_CHECK to enable checking the ETag, but since we never generated a 304 this was pointless.
Adding in the ETag check (via generateCachedResponse) fixed this, but still left the issue that if the startIndex was specified, or if the user's role changed (i.e. on a shared browser during testing), then we used the incorrect cached value anyway.
So finally I updated the ETag to include the role and startIndex too. Adding the values together is safe, since ETag collisions are unlikely and even malicious adjusting of startIndex still can't leak any new data, only generate a useless "304 Not Modified".

I had to change the name of a variable used in the AbstractFacadeTest class, since we needed both a `Request` and an `HttpServletRequest`, but the latter was until now just named `request` which was too confusing! This means some tests are changed in this PR which seem unrelated, but were necessary since the ETag check required adding another property to the method call which broke the tests.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -175,8 +175,8 @@ public class QuizFacade extends AbstractIsaacFacade {
     @Path("/available")
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(summary = "Get tests visible to this user, from index 0.")
-    public final Response getAvailableQuizzes(@Context final HttpServletRequest request) {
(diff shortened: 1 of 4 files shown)


Morpheu5 (approved):

jsharkey13:
We just need some plausibly unique value, which differs correctly when the different settings are changed. Accidental duplicate ETag values will just cause users to see older quizzes for longer (which is what always happens without these changes at the moment). So I'm not fussed about monitoring it! We do record the status codes for each endpoint, so we can look at the levels of 200s vs 304s if we wanted to.

PR #483, opened 2022-09-05
Update some dependencies to address security vulnerabilities

Morpheu5:
- asyncutil is going nowhere but at least we are safe because the vulnerability only affects Java <= 1.6
- snakeyaml was vulnerable but an update to jackson fixed it
- the rest is fairly mild and also don't have immediate fixes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
         <resteasy.version>6.0.3.Final</resteasy.version>
         <guice.version>5.1.0</guice.version>
         <oauth-client.version>1.33.3</oauth-client.version>
-        <jackson.version>2.13.3</jackson.version>
+        <jackson.version>2.13.4</jackson.version>
         <jackson-databind.version>2.13.3</jackson-databind.version>
         <powermock.version>2.0.9</powermock.version>
         <swagger-ui-version>4.13.2</swagger-ui-version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -15,7 +15,7 @@
         <resteasy.version>6.0.3.Final</resteasy.version>
         <guice.version>5.1.0</guice.version>
         <oauth-client.version>1.33.3</oauth-client.version>
-        <jackson.version>2.13.3</jackson.version>
+        <jackson.version>2.13.4</jackson.version>
         <jackson-databind.version>2.13.3</jackson-databind.version>
```
jsharkey13, line 19:
```suggestion
        <jackson-databind.version>2.13.4</jackson-databind.version>
```
The `jackson` and `jackson-databind` versions should be updated at the same time, @Morpheu5, unless databind is ahead by a patch (which it sometimes must be, hence the separate variable).

Morpheu5: > The `jackson` and `jackson-databind` versions should be updated at the same time, @Morpheu5, unless databind is ahead by a patch (which it sometimes must be, hence the separate variable).

Well spotted, I don't see any reasons why this shouldn't be the case.

jsharkey13 (approved):

PR #485, opened 2022-09-22
Allow cached image use after content version change

jsharkey13:
Currently the cache ETag of images is based on the content SHA and file path; a neat trick to avoid having to load the image from disk each time if the user has a cached image and the content version hasn't changed.

If the content has changed, currently we always send the full image anew regardless. However by adding a second part to the ETag that is based on the file contents, we can still send a Not Modified even when the content version changes but the file does not (and this is the most common case for any given content commit change).

Sadly browsers will not update the ETag of a cached item if we return the "new" ETag after a "late" match (because there might be multiple cached matches and if the ETag changes it doesn't know which existing value matched), so we will still have to load the image from disk after a content push - but this now allows us at least to save users and our own bandwidth with the tiny Not Modified response.

As with all things cache, if we don't match even though we could have (e.g. multiple If-None-Match headers) then the user just gets a fresh copy and all is good; and if we do match even though we shouldn't have, then the user just sees a slightly out-of-date image. I read through the generateCachedResponse -> evaluatePreconditions code just to be sure this is equivalent, and it is for GET requests (which this endpoint is).

Browsers don't cache 404's and so I removed the cache stuff from those errors to make it simpler to read.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
@@ -83,6 +83,7 @@ import static uk.ac.cam.cl.dtg.segue.api.Constants.*;
 @Tag(name = "/")
 public class IsaacController extends AbstractIsaacFacade {
     private static final Logger log = LoggerFactory.getLogger(IsaacController.class);
+    private static final String ETAG_SEPARATOR = "/";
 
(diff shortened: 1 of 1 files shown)


jsharkey13:
We ought to have enough logging already to monitor the ratios of 200s to 304s pre and post this change, too!

Morpheu5 (approved):

PR #489, opened 2022-10-04
Always augment `AssignmentDTO`s with very lightweight `GameboardDTO`s

chrisjpurdy:
This is to allow basic gameboard information for assignments to be displayed in the frontend without having to individually request each gameboard of interest.

If only we used GraphQL 😅

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -312,7 +312,6 @@ public class AssignmentFacade extends AbstractIsaacFacade {
             if (null == groupIdOfInterest) {
                 List<UserGroupDTO> allGroupsOwnedAndManagedByUser = this.groupManager.getAllGroupsOwnedAndManagedByUser(currentlyLoggedInUser, false);
                 List<AssignmentDTO> assignments = this.assignmentManager.getAllAssignmentsForSpecificGroups(allGroupsOwnedAndManagedByUser, true);
-                // Add lightweight gameboard and assigner summary information to each assignment
                 Set<String> gameboardIds = assignments.stream().map(AssignmentDTO::getGameboardId).collect(Collectors.toSet());
                 List<GameboardDTO> liteGameboards = this.gameManager.getLiteGameboards(gameboardIds);
                 // Remove unneeded data from the gameboards - very messy but not as messy as feasible alternatives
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -310,8 +311,25 @@ public Response getAssigned(@Context final HttpServletRequest request,
...
+                liteGameboards.forEach(g -> {
+                    g.setContents(null);
+                    g.setWildCard(null);
+                    g.setWildCardPosition(null);
+                    g.setGameFilter(null);
+                    g.setLastVisited(null);
+                    g.setPercentageCompleted(null);
+                    g.setCreationMethod(null);
+                    g.setCreationDate(null);
+                });
```
Morpheu5, line 327: OK, this works just fine, it's a bit messy but gets the job done and is explicit enough. However, given that we use mappers elsewhere to do a similar thing, wouldn't it be more idiomatic (I'm not saying "better" on purpose here :) to make a `LiteGameboardDTO` and use a mapper to get all the unnecessary data stripped?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -348,6 +348,19 @@ public GameboardDTO getLiteGameboard(final String gameboardId) throws SegueDatab
...
+    /**
+     * Get a list of gameboards by their ids.
+     *
+     * @param gameboardId
+     *            - the ids to find.
+     * @return a list of lite gameboards without their question data fully expanded.
+     * @throws SegueDatabaseException
+     *             - a database error has occurred.
+     */
+    public List<GameboardDTO> getLiteGameboards(final Collection<String> gameboardId) throws SegueDatabaseException {
```
Morpheu5, line 360: Or, even better, if we eventually want to move towards a progressive enhancement approach rather than progressive destruction (…) approach[^1], to make a new query at the persistence level that just doesn't query unnecessary data, and returns a collection of `LiteGamboardDO`s instead?

[^1]: This is only tangentially related to this approach, but once you pick an idiom, it's a good idea to not stray too far away from it.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java
```diff
@@ -154,7 +154,22 @@ public List<GameboardDTO> getGameboardsByIds(final List<String> gameboardIds) th
...
+     * resolve any fine grain details about the boards. E.g. no question details
+     * will be retrieved.
+     *
+     * @param gameboardIds
+     *            - to retrieve.
+     * @return a list of lightly populated gameboards.
+     * @throws SegueDatabaseException
+     *             - if there are problems with the database.
+     */
+    public List<GameboardDTO> getLiteGameboardsByIds(final Collection<String> gameboardIds) throws SegueDatabaseException {
```
Morpheu5, line 171: Yeah, the previous comment should have gone here but hey.

jsharkey13 (approved):
I think we just have to try this in the wild now, to see what impact it has.

PR #490, opened 2022-10-06
Cloze question detailed item feedback

jsharkey13:
This extends #487, and should be merged after it. The diff might be easier to view just [between this PR and the previous one](https://github.com/isaacphysics/isaac-api/compare/feature/cloze-validator-improvements...feature/cloze-question-item-feedback) rather than onto master directly yet.

It is opt-in, via the detailedItemFeedback property of Cloze Questions.

The only nasty thing is that the QuestionValidationResponseDeserializer doesn't have any type information; we don't store that in the question attempts table. That means it has to work out what to deserialise something into based on the JSON content. For QuantityVRs, we use the fact that the submitted answer will be a Quantity; but for cloze questions we re-use ItemChoices and so we cannot do that. Instead just check if the unique property of ClozeVRs is set.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dos/ClozeValidationResponse.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dos/ClozeValidationResponse.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2022 James Sharkey
+ *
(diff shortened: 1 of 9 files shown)


jsharkey13:
It might be nice to add some tests for this little extra bit of behavior before this is merged. But this ought to be enough for someone to build the frontend UI for this feature.

PR #493, opened 2022-10-24
Consider cancelled booking status in events facade and manager methods

chrisjpurdy:
I started this but am not sure I can finish it because of a few unanswered questions at the moment, including:

- Should users be emailed when an event is cancelled? (I think they should)
- Should `EventBookingManager::createBooking` throw errors/fail nicely if the event being booked onto is cancelled or closed? (It doesn't currently; I think it should throw an error, requiring the caller to ensure the event is open)


jsharkey13:
I do not think that we should send emails when an event is cancelled, partly because this is hard (events get cancelled by content changes, and we have no state to know when it has changed from version to version) and partly because it's very easy already to email all users a custom email that gives more detail than an automatic one could.

I _do_ think the BookingManager should throw an `EventCancelled` exception or suchlike, since there are plenty of other endpoints (reservations, admin booking, etc) that should also check cancellation status.

What about the event reminder and feedback emails, do they also respect cancellation status?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
```diff
@@ -1034,6 +1041,11 @@ public final Response createBookingForMe(@Context final HttpServletRequest reque
                     .toResponse();
             }
 
+            if (EventStatus.CANCELLED.equals(event.getEventStatus())) {
+                return new SegueErrorResponse(Status.BAD_REQUEST, "Sorry this event is cancelled. Please try again later.")
```
mwtrew, line 1045: Although I doubt anyone will take this literally, I don't think "please try again later" makes much sense here. Most events probably won't be un-cancelled once cancelled.

mwtrew (approved):
Looks good overall. Emailing confirmed users automatically on cancellation would be nice, but admins have the option of emailing confirmed users relatively easily through the admin tools so that's not critical. Perhaps a template email for that would be useful.

jsharkey13:
We (or at least I) decided explicitly agains automatic emails, since it is both very hard to detect a cancellation since that's just a content push, and because using the standard event email feature is to easy. Cancellations usually happen for various reasons, so I don't know that a template would be much use (it would need too much personalising for each reason: is it postponed, if so when to; or cancelled altogether, etc).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java
chrisjpurdy: Good point 😆 I'll remove it

PR #496, opened 2022-10-27
Allow assignments to be scheduled on the same date as they are due

chrisjpurdy:
~~By rounding down the scheduled start date to the nearest day in order to perform the comparison.~~ This now rounds the due date up to the beginning of the next day, using `isBeforeDueDate` to perform comparisons.

Should be merged with [the related app PR](https://github.com/isaacphysics/isaac-react-app/pull/663).

This also fixes an edge case where assignment emails might not be sent if an assignment start date is set to a given day or before it.

Care should be taken when reviewing to make sure the date/time logic is correct - I tried to make sure it was but these things are prone to syntactically subtle errors. 
**I think the test is failing because it was testing incorrect logic - I need a second opinion on this before I change it.**

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -393,7 +393,7 @@ public class QuizFacade extends AbstractIsaacFacade {
             }
 
             // Check the due date hasn't passed
-            if (quizAssignment.getDueDate() != null && !quizAssignment.isBeforeDueDate(new Date())) {
+            if (quizAssignment.getDueDate() != null && !quizAssignment.dueDateIsAfter(new Date())) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -683,7 +683,7 @@ public final Response markIncompleteQuizAttempt(@Context final HttpServletReques
                     "You can only mark assignments incomplete for groups you own or manage.").toResponse();
             }
 
-            if (assignment.getDueDate() != null && assignment.getDueDate().before(new Date())) {
+            if (assignment.getDueDate() != null && assignment.isBeforeDueDate(new Date())) {
```
chrisjpurdy, line 686: As far as I can tell, this line is causing the tests to fail. However, surely a test is "still due" if the current time is before the due date, i.e. the due date hasn't passed yet?

jsharkey13 (approved):

PR #497, opened 2022-10-27
Update Elasticsearch client

mwtrew:
This PR moves from the outdated Transport client to the less outdated High Level REST Client, and to the latest minor version of ES itself (7.17). The API is fairly similar except for removal of mapping types, which luckily we did not rely on - each type already had its own index.

This PR also introduces password authentication, which is not terribly important for our use-case but the new client emits warnings on every request otherwise. This will require a config/environment variable update when deployed.

---

**Pull Request Check List (todo)**
- [ ] Peer-Reviewed


mlt47 (approved):
The changes all look fine to me. Not too big a translation between the client types, thankfully.
I will think through the deployment of an ES node with an updated version to the cluster next week, do it, and document it for us to follow in the future.

PR #499, opened 2022-10-31
Detailed item feedback for cloze questions

jsharkey13:
This PR is blocked on #498 having been released, since as soon as question attempts using this new validation response are stored in the database they will be deserialised by older API versions which will fail to parse them. We need to release #498, and then wait a release cycle to merge this.

This adds a new `itemsCorrect` property to the validation response which is a list of correctnesses (true or false) for each of the submitted items.


chrisjpurdy (approved):
Code looks good to me, and I can confirm that it works well after testing another change dependant on this one extensively 👍

PR #500, opened 2022-10-31
Allow additional managers to edit groups that they manage

chrisjpurdy:
This PR enables a few things:

- Group owners explicitly giving additional managers extra privileges to rename and archive group, as well as managing other managers assignments
- Group owners transferring ownership to an additional manager
- Sending emails to additional managers to notify them if the above things occur
---
Content know about and are happy with this change: Nicki in particular gave me pointers on the best approach for this.


src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
```diff
@@ -400,9 +400,22 @@ public Response editGroup(@Context final HttpServletRequest request, final UserG
                 return new SegueErrorResponse(Status.NOT_FOUND, "Group specified does not exist.").toResponse();
             }
 
+            // If user is not the owner...
             if (!existingGroup.getOwnerId().equals(user.getId())) {
-                return new SegueErrorResponse(Status.FORBIDDEN,
-                        "The group you have attempted to edit does not belong to you.").toResponse();
+                // Check if an additional manager with privileges is allowed to make these changes
+                boolean ownerOnlyFieldChanged = !existingGroup.getOwnerId().equals(groupDTO.getOwnerId())
```
chrisjpurdy, line 406: This is kind of nasty, but I can't see a reason why it wouldn't work. 
I think that in edge cases where this returns `true` when it probably shouldn't, something weird has happened in the front-end to mutate the group object.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/GroupManager.java
```diff
@@ -410,6 +420,50 @@ public UserGroupDTO addUserToManagerList(final UserGroupDTO group, final Registe
...
+        // Change old and new owners additional manager status if appropriate
+        if (!group.getAdditionalManagersUserIds().contains(oldOwner.getId())) {
+            this.groupDatabase.addUserAdditionalManagerList(oldOwner.getId(), group.getId());
+        }
+        if (group.getAdditionalManagersUserIds().contains(newOwner.getId())) {
+            this.groupDatabase.removeUserFromAdditionalManagerList(newOwner.getId(), group.getId());
+        }
+
+        // ! We are mutating this group object, but this particular mutation should be safe !
+        group.setOwnerId(newOwner.getId());
```
chrisjpurdy, line 457: This is safe because `editUserGroup(group)` transforms the `UserGroupDTO` into a DO to persist it to the DB. `UserGroupDO`s don't store any information about additional managers, so it doesn't matter that we don't update the additional manager list in `group`.

chrisjpurdy: Looking at this again - it is a bit of a code smell, as it relies on internal DB logic and what happens inside `editUserGroup`, but provided we don't change how we store additional managers, it will work.

chrisjpurdy (commented):
Looked over it again to double check stuff, seems fine!

src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
mwtrew: Looking into the DTO pattern [^1][^2] a bit, they exist to hold only the fields necessary for a particular interaction - that sounds relevant here as there are fields we should not be able to update from this endpoint at all, and some that depend on the user/group relationship. Perhaps creating new DTOs reflecting only the fields that can be updated in each situation would be cleaner, along with versions of `groupManager::editUserGroup()` to process them. That would be a bigger change of course!

[^1]: https://auth0.com/blog/automatically-mapping-dto-to-entity-on-spring-boot-apis/#What-is-a-DTO
[^2]: https://snyk.io/blog/how-to-use-java-dtos/

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/GroupManager.java
mwtrew: I think this is fine. It would be nice if it was more obvious from here that `editUserGroup()` is a simple "map to DO and persist" and therefore other fields like additional managers would be lost anyway, but that's not a problem specific to this change.

mwtrew (approved):
Looks good, one thought about the `editGroup()` checks in the facade. Nice to see the integration tests.

src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
chrisjpurdy: I fully agree, but it would be a bigger change. The immense amount of boilerplate required for simple things like this is annoying to say the least. Since (I think) we want admins to be able to edit group objects as they please, we would need a different endpoint that accepts the smaller DTO, and as you say new versions of `editUserGroup` etc. 
Considering there are quite a few endpoints where we make this mistake already, I'd rather make a card for targeting all of them in one clean refactor?

PR #508, opened 2023-01-23
Replace apostrophes with tildes before ordering user names

chrisjpurdy:
A bit hacky, but there doesn't seem to be a better option.
 
The reason for this fix is so that ordering of user name strings in output CSVs match the default order of strings in Excel.


jsharkey13 (approved):
I'm not the biggest fan of this, but let's see what happens when we deploy it in the real world.

PR #512, opened 2023-02-03
Subscribe all users to "Legal notices" MailJet list

jsharkey13:
This new list will receive all users who have valid email addresses.
Additionally, compare the lowercase version of the email to the MailJet email, since they lowercase their emails.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
@@ -68,7 +68,6 @@ public class ExternalAccountManager implements IExternalAccountManager {
 
                     String accountEmail = userRecord.getAccountEmail();
                     boolean accountEmailDeliveryFailed = EmailVerificationStatus.DELIVERY_FAILED.equals(userRecord.getEmailVerificationStatus());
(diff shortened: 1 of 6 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/ExternalAccountManager.java
```diff
@@ -87,7 +86,7 @@ public synchronized void synchroniseChangedUsers() throws ExternalAccountSynchro
                             //    Action: same as deletion? Or just remove from lists for now?
                             log.debug("Case: delivery failed.");
                             mailjetApi.updateUserSubscriptions(mailjetId, MailJetSubscriptionAction.REMOVE, MailJetSubscriptionAction.REMOVE);
-                        } else if (!accountEmail.equals(mailjetDetails.getString("Email"))) {
+                        } else if (!accountEmail.toLowerCase().equals(mailjetDetails.getString("Email"))) {
```
mlt47, line 89: From this, I'm assuming MailJet always stores emails lower case? I guess in the past if a user had upper-case letters in their email each synchronize changed user would have meant the creation of a new user and deletion of another rather than updating the user. As those operations are the same (except for maybe some history loss) we don't have to worry about prior users who incorrectly fell down this branch.

chrisjpurdy (approved):
Seems good to me!

mlt47 (approved):
Looks like it should do the correct thing to me.

PR #514, opened 2023-02-17
Remove hardcoded regex restricting signup emails

jsharkey13:
Move this check that users are not signing up with restricted emails (i.e. those of the project domain, since no verification is done on signup of these emails) into a config property.


mlt47 (approved):
This looks good.
How serious is it if this was misconfigured? I'm guessing it would just be an inconvenience more than anything if someone was to sign up with one of these email domains - perhaps they would not be counted in some site stats and it might catch a mistake on the part of the user. If there is any greater risk, I would be tempted to log an error if the property is not set to catch accidental misconfiguration.

jsharkey13:
It's not serious in any practical sense, this is more about policy than security; no priviliges are granted based on email addresses, it may just slightly increase the risk of a phishing-style attack on the live site if unset. It's very easy to test with an end-to-end test, at least.

PR #520, opened 2023-03-24
Numeric question sig-fig warnings

chrisjpurdy:
Adds a couple of content errors, and refactors where the default sig-fig bound is applied if either are `null`. 
Also allows numeric questions with exact match but malformed sig-fig rules to be answered.


mwtrew (approved):
It would be nice to see a couple of quick unit tests for the validator changes as an act of kindness to your beleaguered reviewer, but the changes look good.

PR #521, opened 2023-03-24
Add an endpoint to shutdown the Quartz scheduler

jsharkey13:
The way we use Quartz means that we cannot restart it again after shutting it down this way, so the endpoint makes it clear that the operation is permanent.
Sadly the "isStarted" method just checks that the scheduler has ever been started, not whether is has subsequently been shutdown. I altered the existing methods to better reflect this.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
@@ -1304,7 +1304,7 @@ public class AdminFacade extends AbstractSegueFacade {
     }
 
     @POST
-    @Path("/start_quartz")
+    @Path("/quartz/start")
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(summary = "Start the Quartz Job Scheduler service if not already started.")
     public Response startQuartzJobService(@Context final HttpServletRequest request) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/AdminFacade.java
```diff
@@ -1304,7 +1304,7 @@ public Response syncExternalAccounts(@Context final HttpServletRequest httpServl
     }
 
     @POST
-    @Path("/start_quartz")
+    @Path("/quartz/start")
```
jsharkey13, line 1307: I think this change is safe. It's an admin-only endpoint, and we don't automate this at all.

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/SegueJobService.java
```diff
@@ -180,6 +193,17 @@ public synchronized void initialiseService() throws SchedulerException {
...
     }
 
+    /**
+     * Attempt to complete running jobs, then shut down the scheduler.
+     * @throws SchedulerException on failure to terminate.
+     */
+    public synchronized void shutdownService() throws SchedulerException {
+        if (wasStarted()) {
+            log.info("Shutting down Segue Job Service");
+            this.scheduler.shutdown(true);
```
mlt47, line 203: Cool, and passing in `true` means it will wait for all running jobs to complete first. (Your javadoc does convey this).

mlt47 (approved):

PR #522, opened 2023-03-24
Numeric question exact match unit check improvement

chrisjpurdy:
Allows incorrect units to be marked as such in numeric q "exact match" checks.

`/questions/orbiting_moon` is a good test case - 1410 kJ is a known wrong answer, but kJ is never a correct unit (occurs in none of the correct answers). However when you give 1410 kJ as the answer, the unit isn't marked as incorrect. Funnily enough, the unit _is_ marked as incorrect if you answer with 1410.0 kJ.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
@@ -513,10 +513,19 @@ public class IsaacNumericValidator implements IValidator {
                                                         final Quantity answerFromUser) {
 
         log.debug("\t[exactStringMatch]");
+        boolean wasACorrectAnswerWithUsersSelectedUnit = false;
 
-        for (Choice c : isaacNumericQuestion.getChoices()) {
+        // Goes through correct answers first, in order to work out whether any correct answers match the units that
+        // the user selected.
+        List<Choice> orderedChoices = getOrderedChoices(isaacNumericQuestion.getChoices());
+        for (Choice c : orderedChoices) {
             if (c instanceof Quantity) {
                 Quantity quantityFromQuestion = (Quantity) c;
+
+                // Record whether the units the user selected match the units of a correct answer
+                if (isaacNumericQuestion.getRequireUnits()) {
+                    wasACorrectAnswerWithUsersSelectedUnit |= quantityFromQuestion.getUnits().equals(answerFromUser.getUnits()) && quantityFromQuestion.isCorrect();
+                }
 
                 StringBuilder userStringForComparison = new StringBuilder();
                 userStringForComparison.append(answerFromUser.getValue().trim());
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -513,11 +513,25 @@ private QuantityValidationResponse exactStringMatch(final IsaacNumericQuestion i
                                                         final Quantity answerFromUser) {
 
         log.debug("\t[exactStringMatch]");
+        boolean wasACorrectAnswerWithUsersSelectedUnit = false;
```
jsharkey13, line 516: Is this variable necessary? Or, rather, since this is necessary, why is `unitsCorrect` still necessary? See the comment further down about the `unitsCorrect` variable.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -531,14 +545,14 @@ private QuantityValidationResponse exactStringMatch(final IsaacNumericQuestion i
                 }
 
                 if (questionAnswerString.toString().trim().equals(userStringForComparison.toString().trim())) {
-                    Boolean unitFeedback = null;
+                    Boolean unitsCorrect = null;
                     if (isaacNumericQuestion.getRequireUnits()) {
-                        unitFeedback = quantityFromQuestion.getUnits().equals(answerFromUser.getUnits());
+                        unitsCorrect = wasACorrectAnswerWithUsersSelectedUnit || quantityFromQuestion.isCorrect();
```
jsharkey13, line 550: Why is this necessary? `wasACorrectAnswerWithUsersSelectedUnit` has already been OR-ed with `quantityFromQuestion.isCorrect()` above so this is always redundant.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -531,14 +545,14 @@ private QuantityValidationResponse exactStringMatch(final IsaacNumericQuestion i
...
                     }
 
                     return new QuantityValidationResponse(isaacNumericQuestion.getId(), answerFromUser,
                             quantityFromQuestion.isCorrect(), (Content) quantityFromQuestion.getExplanation(),
-                            quantityFromQuestion.isCorrect(), unitFeedback, new Date());
+                            quantityFromQuestion.isCorrect(), unitsCorrect, new Date());
```
jsharkey13, line 555: Why not just use `wasACorrectAnswerWithUsersSelectedUnit` here? Or, better, rename that outer variable to `unitsCorrect` and ditch this inner variable entirely.

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java
```diff
@@ -513,11 +513,25 @@ private QuantityValidationResponse exactStringMatch(final IsaacNumericQuestion i
                                                         final Quantity answerFromUser) {
 
         log.debug("\t[exactStringMatch]");
+        boolean wasACorrectAnswerWithUsersSelectedUnit = false;
 
-        for (Choice c : isaacNumericQuestion.getChoices()) {
+        // Goes through correct answers first, in order to work out whether any correct answers match the units that
+        // the user selected.
+        List<Choice> orderedChoices = getOrderedChoices(isaacNumericQuestion.getChoices());
```
jsharkey13, line 520: Doing this sorting is wasteful, since we don't ever _use_ the result of `exactStringMatch` unless the answer is a match to a known incorrect choice; for every other case (i.e. the vast majority), we then do the normal `validateWith[out]Units` which does this sorting again.

I see that in order for the `wasACorrectAnswerWithUsersSelectedUnit` variable to be useful we need this sorting, but I wonder at this point if it is time to just remove the `exactStringMatch` check entirely - since it was only added to be a very quick very simple short circuit long before most of the rest of the complexity was added on.

jsharkey13 (changes requested):
I am not sure that this is the right approach to fixing this issue. I agree it does fix it, and (give or take an unnecessary variable) it does so with minimal changes. 

It seems to me a better fix, requiring no additions at all, would be to just remove `exactStringMatch` entirely. It has been a long time since that code was helpful; we no longer use the result of it at all except in the rare case of an exact string match to a known incorrect answer. Otherwise all we are doing is making it more complicated to provide feedback we already generate correctly in the main `validateWithUnits`, and comitting to keeping the two consistent when one is barely ever used.

I've reviewed this validator most often, so I probably have the best insight into why `exactStringMatch` exists (it is the oldest part of the class, pre-dating significant figures!) and I think it can go. But I'm open to discuss it if anyone else has an opinion!

chrisjpurdy:
Closed in favour of https://github.com/isaacphysics/isaac-api/pull/527

PR #527, opened 2023-04-17
Remove exact match pathway for numeric question validation

chrisjpurdy:
Alternate (and preferred) fix to [this PR](https://github.com/isaacphysics/isaac-api/pull/522). See James' comment there for why this is okay to do.


jsharkey13 (approved):
Sometimes removing code can be quite satisfying! Especially really old code, that complicates things. 
I take it your are happy with my suggestion then?

chrisjpurdy:
Very happy with the suggestion! Shorter code that does the same thing is better code IMO if we don't mind too much about efficiency.

PR #529, opened 2023-04-28
Cookie updates

jsharkey13:
This moves all our cookies to be set with the Secure and HttpOnly flags (previously the JSESSIONID was not, nor were local dev cookies), and sets SameSite=Strict on all the cookies (previous we were implicitly using SameSite=Lax in modern browsers).

I think this is fine; the difference between Lax (what we have been implicitly using before now) and Strict is about top-level navigations and third-party triggered GET requests.
We don't use any cookies on the very-top-level GET; that goes to the frontend which is stateless.
I also do not think that third-party origins should be making authenticated GET requests (our CORS would stop the response being used in most cases anyhow), so that should also be safe.
The [SameSite attribute is supported](https://caniuse.com/same-site-cookie-attribute) on all the browsers we care about.

If this breaks, moving to SAME_SITE_LAX_COMMENT is the next one to try.

This change to `web-api-local.xml` may break local development using Safari, if it [still does not do the sensible thing](https://bugs.webkit.org/show_bug.cgi?id=218980) Chrome and Firefox do by allowing insecure "Secure" cookie to be set on localhost origins. It won't affect the live sites at all, and if we _really_ need to test something in Safari it's straightforward to remove temporarily.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
@@ -80,6 +80,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.eclipse.jetty.http.HttpCookie.SAME_SITE_STRICT_COMMENT;
 import static uk.ac.cam.cl.dtg.segue.api.Constants.*;
 
 /**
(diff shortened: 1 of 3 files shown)


mwtrew (approved):
Looks good, but it does break for Safari on `localhost`. Perhaps a comment in `web-api-local.xml` explaining that would be useful.

jsharkey13:
Let's discuss whether we need to add a comment, and if the comment would help a developer more than just the DevTools warning would, later.

PR #530, opened 2023-05-03
YAML configuration and override support

mwtrew:
This PR adds support for YAML configuration files, replacing `segue-config.properties`. This is required to use SOPS (which doesn't support `.properties` files) for secrets management. 

The properties file loader is kept because we still use it for `content-indices.properties`. As you can see this touches quite a few files, but most of these are changing the type of `PropertiesLoader` references to the new `AbstractConfigLoader`, which Guice now binds to the `YamlConfigLoader` for the main config file.

Additionally, it is now possible to specify a comma-separated list of config files (e.g. `-Dconfig.location=/.../segue-config.yaml,/.../segue-config-overrides.yaml`). In the case of conflicting keys, keys from the files later in the list will be chosen. This allows for user-specific "override" config files which are helpful for local development.

---

**Pull Request Check List**
- ~~Security - Data Exposure - Test any altered or created endpoints using [swagger](http://localhost:8080/isaac-api/api-docs/)~~
-~~Security - Access Control - Check authorisation on every new endpoint~~
- [x] Security - New dependency - Searched for any know vulnerabilities
- [x] Security - New dependency - Added to dependency list
- [x] Peer-Reviewed


pom.xml
```diff
@@ -44,6 +44,11 @@
         </repository>
     </repositories>
     <dependencies>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>2.0</version>
+        </dependency>
```
mwtrew, line 51: This was already an indirect `jackson` dependency.

src/main/java/uk/ac/cam/cl/dtg/util/WriteablePropertiesLoader.java
jsharkey13, line 1: Did you intend to change the permission bits on this file? Seems unlikely that this needs to be executable!

src/main/java/uk/ac/cam/cl/dtg/util/YamlLoader.java
```diff
@@ -0,0 +1,100 @@
...
+
+            // check to see if this a resource or a file somewhere else
+            if (getClass().getClassLoader().getResourceAsStream(this.configPath) == null) {
+                File file = new File(path);
+                try (FileInputStream ioStream = new FileInputStream(file)) {
+                    // then we have to look further afield
+                    subConfig = yaml.load(ioStream);
+                }
+            } else {
+                subConfig = yaml.load(getClass().getClassLoader().getResourceAsStream(path));
```
jsharkey13, line 87: Does this not load the file twice if it's a resource (i.e. if it falls through to the else, it gets loaded again)?

src/main/java/uk/ac/cam/cl/dtg/util/WriteablePropertiesLoader.java
mwtrew: I'm set up to ignore filemode changes usually, this must have slipped through since it's a new file.

src/main/java/uk/ac/cam/cl/dtg/util/YamlLoader.java
mwtrew: This is the same as the existing `.properties` logic so I didn't check this in detail, I assume that's a reasonable way of checking whether the config is a resource or not. I noticed another issue here though, so have updated the PR with a couple of tests covering this.

PR #543, opened 2023-06-21
Improve bulk actions for group management

jsharkey13:
This adds two new database methods for loading bulk group information: one to bulk load groups themselves, and one to load a user's group membership status specifically rather than group by group.

Loading the groups in bulk, and only optionally augmenting them with owner and manager information, removes the need for many database queries: before the /assignments endpoint required `n_groups x (2 + n_managers_per_group)` queries to load group data, from which only the group name was used. Now it requires just 1 query since most of the information was not needed and can be skipped.

We also then filtered the assignments by the user's group membership status, but the only way to do this was to load the statuses for _all_ users in a group, for each group. This adds a new endpoint to load a specific user's group status for all their groups, reducing the number of queries required down to always 1.

For /assignments this reduces the total number of queries from `~ (n_groups x (3 + n_managers_per_group)) + n_teachers + 4` down to `6 + n_teachers` which is a vast improvement!

(I re-ordered some imports, since IntelliJ's "Optimise Imports" was unhappy with their ordering, but this are safe to ignore).

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
@@ -17,25 +17,11 @@ package uk.ac.cam.cl.dtg.isaac.api;
 
 import com.google.api.client.util.Lists;
 import com.google.api.client.util.Maps;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import com.opencsv.CSVWriter;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.tags.Tag;
-import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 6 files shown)


jsharkey13:
The test is failing because `filterItemsBasedOnMembershipContext` is explicitly not mocked, since it needs to be tested.
But I have moved it from using other methods of its own class to using a method from one of its properties. Since we don't mock the constructor, we can't alter this property's behaviour. If we do mock the constructor, all the other methods fail to be mocked correctly. I can't see how to fix this, I have tried several times and failed.

chrisjpurdy:
Would it be possible to have a stub/mock implementation of `IUserGroupPersistenceManager`, that we add to Guice instead of `PgUserGroupPersistenceManager` when running the test suite?
If not then we probably just need to add tests for that functionality to the integration test suite, and remove the failing ones, so that nothing needs to be mocked?

jsharkey13:
I think I found (yet another) way of doing partial mocks that I hadn't already tried, and after some false starts I think I have the tests working as intended now! 🤞

PR #545, opened 2023-07-04
Avoid caching mutable objects

jsharkey13:
DTO objects exist to be transferred to the client, and we were caching them because it is expensive to load them from ElasticSearch and also expensive (but less so) to map them from DO to DTO form.

However, despite the warnings in the JavaDoc, we keep mutating the DTOs returned by the GCM without cloning them. This is understandable; they exist to contain user data for transfer.
Rather than ditch the cache entirely, we remove the cache for DTOs but continue to cache DOs, since the latter are actually supposed to be unchanging anyway and we do not mutate them anywhere at present.
This adds ~1-3ms onto a cache hit, but it prevents all sorts of data leaking from being possible. Since a cache miss is ~10ms, hopefully this extra time will be okay.

---

There are two methods that load objects in bulk. The results from these methods are currently not modified, and the performance penalty from removing the DTO cache is large enough to warrant keeping the methods. Instead of removing them, name them in a way that makes things unambiguous.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java
@@ -737,7 +737,7 @@ public class GameboardPersistenceManager {
             // Search for questions that match the ids.
             ResultsWrapper<ContentDTO> results;
             try {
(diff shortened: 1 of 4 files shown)


chrisjpurdy (approved):
Both Meurig and I approved this 👍

PR #546, opened 2023-07-12
Improved page loading

jsharkey13:
None of these methods used question attempts, and locally there is not much difference between the speed of the old method and the new one. Sometimes it could be 75% faster, sometimes the same time; it was never worse which is good enough for me. I expect better results on the server (where the cache is helpful and other things are happening in the API).

We now consistently used `getContentById` for loading pages and fragments of all kinds, and so `findSingleResult` can and has been removed. The code is easier to read since all the methods are very similar to one another. A later improvement might be to reduce the code duplication.


mlt47 (commented):
Looks good. Shared DO cache between users should hopefully make a visible difference to response times for these endpoints.

mlt47 (approved):

PR #547, opened 2023-07-13
Use DOs for assignment and gameboard calculations

jsharkey13:
In these cases we were not using the question DTOs for transfer, but as reference for what parts a question contains. We can use DOs for this since they are the authoritative version of questions, and we do not plan to use the objects for anything which is transferred to the client. This avoids the need to create new DTO objects every time, and can use the DO cache.


mlt47 (commented):
It's a shame that there's so much duplicated logic!

PR #548, opened 2023-07-14
Stop using token owner for group join permissions

jsharkey13:
For some reason both groups and group join tokens have ownership information. We added support for altering the owner of a group, but this did not alter the owner of the token.
When users tried to join a group with a changed owner, the new owner was not listed amongst the users to approve, but the old token owner was. This meant that the group owner was not given permission.

This commit stops using the token ownership information for anything; tokens really belong to groups, not to users, and the old approach of tokens belonging to users has not been correct in some time. As such, the test that group tokens could work without a group is not useful; they cannot work that way since the introduction of additional managers.

This commit also purges methods for creating an association based on a token, since that method did not include group managers; it was misleading and required group managers to be dealt with separately.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dos/AssociationToken.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dos/AssociationToken.java
@@ -72,6 +72,7 @@ public class AssociationToken {
      * 
      * @return the ownerUserId
      */
+    @Deprecated
     public Long getOwnerUserId() {
         return ownerUserId;
     }
@@ -82,6 +83,7 @@ public class AssociationToken {
      * @param ownerUserId
      *            the ownerUserId to set
      */
+    @Deprecated
     public void setOwnerUserId(final Long ownerUserId) {
         this.ownerUserId = ownerUserId;
     }
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAssociationManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAssociationManager.java
@@ -234,15 +234,11 @@ public class UserAssociationManager {
 
         AssociationToken lookedupToken = associationDatabase.lookupAssociationToken(token);
 
-        if (null == lookedupToken) {
+        if (null == lookedupToken || lookedupToken.getGroupId() == null) {
             throw new InvalidUserAssociationTokenException("The group token provided does not exist or is invalid.");
         }
 
         UserGroupDTO group = userGroupManager.getGroupById(lookedupToken.getGroupId());
-
-        if (lookedupToken.getGroupId() == null) {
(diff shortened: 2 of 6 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/dos/AssociationToken.java
chrisjpurdy, line 1: Is this entire class deprecated now?

chrisjpurdy (approved):
I'm glad we have tests for this because it's not code I've looked at in detail before. Looks good in general though 👍 
Edit: I spotted what might be a bug that the tests didn't catch - see comments below

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAssociationManager.java
```diff
@@ -238,25 +238,24 @@ public AssociationToken createAssociationWithToken(final String token, final Reg
...
         }
 
-        // add owner association
-        if (!associationDatabase
-                .hasValidAssociation(lookedupToken.getOwnerUserId(), userGrantingPermission.getId())) {
-            associationDatabase.createAssociation(lookedupToken, userGrantingPermission.getId());
+        UserGroupDTO group = userGroupManager.getGroupById(lookedupToken.getGroupId());
+
+        if (lookedupToken.getGroupId() == null) {
+            throw new InvalidUserAssociationTokenException("The group token provided does not exist or is invalid.");
```
chrisjpurdy, line 244: Why is this check here rather than before the previous line? We could possibly skip a database access in the error case

chrisjpurdy: Actually, I think this should 100% be before the previous line given that you'll run into a null pointer exception otherwise :/

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/AssociationToken.java
jsharkey13: Nope, we still want group tokens! We just don't want them to have owners any longer. The rest of the fields are correct and map to database columns that will stay!

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAssociationManager.java
chrisjpurdy: On the earlier `null == lookedupToken` check you could change it to `null == lookedupToken || null == lookedupToken.getGroupId()`

jsharkey13: The (reverse of the) check was after the group lookup before, if you look at the old code. You're right, though, that it's not in the right place. I also think it's checking the wrong thing anyhow. Will go and take a detailed look! 🙌

jsharkey13:
These integration test failures are false positives in completely unrelated code 😣

PR #549, opened 2023-07-20
Remove credentials of deleted users

jsharkey13:
After we moved to soft-deletion of users, the rows of this table stopped being removed by the ON DELETE CASCADE. But we have no cause for keeping old credentials and so should remove them.
For existing deleted users we can remove the credentials with an SQL script.
This forms important groundwork for archiving of user accounts.


mlt47 (approved):
This looks good to me and the comments indicate that you've thought about the 3rd-party auth case.
The only thing I would remind you of was of Alastair's suggestion to maybe record the date of the deletion in the password field.
If you want to make that change `concat('DELETED@', CURRENT_DATE)` works.

jsharkey13:
We already record the date a user is deleted in both the `logged_events` table, and also as the `last_updated` time of the deleted `users` row. We agreed we likely did not need another copy of the date!

PR #563, opened 2023-10-10
Unit tests for PR #520

skyepurchase:
Adds unit tests for handling significant figures when they prevent a question from being answerable. Verifies that the correct warnings are sent and that these are not sent when the significant figures are meant to be disregarded.


mwtrew (changes requested):
Nice tests - a couple of minor comments around documentation.

mwtrew (approved):

PR #566, opened 2023-11-01
Stop mergeuser script skipping data on duplicates

jsharkey13:
This issue came about because I misunderstood how the PL/pgSQL EXCEPTION WHEN unique_violation THEN handling works. I assumed that it skipped rows that caused unique constraint violations. But of course it is obvious in hindsight that it instead aborts the whole BEGIN ... END block when an error occurs. This means that if _any_ conflicting item (say group membership) exists for both accounts in a merge, then all other group membership for the "to delete" user is skipped, leading to data loss!

Fix this using either row-based skipping with ON CONFLICT DO NOTHING, or more fine-grained updating using the DO UPDATE syntax. The only issue this really introduces is the fragility of having to hard-code the columns in the merge script; but many of these tables have not had schema changes in years.


mlt47 (approved):
This all looks correct to me.

PR #568, opened 2023-11-23
Update sketcher question DO/DTO to support axis labels

jacbn:
Adds axisLabelX and axisLabelY fields to the IsaacGraphSketcherQuestion DO and DTO for use with the new content editor update enabling this functionality.


mlt47 (approved):
Because this alters the content model and we run the system to have two version of the API but a shared ETL and ElasticSearch DB we should merge and release this API but wait a release cycle before updating ETL, so that all APIs can handle the new fields.

PR #571, opened 2024-01-17
ETL: halt on exceptions

jsharkey13:
I am sure there could be a subtype of Exception created here, but we do not use the type of the exception anywhere, and the parent methods also throw Exception 🤷

These exceptions prevent indexing from proceeding on error, which has happened several times recently with fairly undesired consequences. Hard failures like these could leave the indices in invalid states,
many of which will be dealt with by the expungeAnyContentTypeIndicesRelatedToVersion check at the start of buildElasticSearchIndex. I have added one further attempt to purge indices on error, though if the error is with ES, then this will fail too.


mlt47 (approved):
The important thing is that now `ETLManager.setNamedVersion(...)` will also raise an exception before moving the alias to point to the failed indices.

PR #572, opened 2024-01-25
Group code cleanup

jsharkey13:
The convertGroupsToDTOs used to have a lookup cache for group owners, but not additional managers. However, the method loaded the manager IDs and then the user data for each manager _one_ by _one_, group by group; this could make hundreds of database queries, even if they were no duplicate users in the lists.

This PR adds a new method for bulk loading additional manager data for multiple groups at once, then refactors the group conversion to use the new method. The most efficient approach is to make a unique set of all owners and managers and load their user data once, so that is what the code does. Even though the streams approach means we now loop though the "groups" list ~5 times rather than once, the reduction in database queries from 2*n_groups to 2 is a massive performance improvement.

The tightening from Iterable to Collection was necessary to use Streams, and since we only pass in collections anyhow, caused no issues. Also suppress the warnings about missing users to debug.

The method does try a new approach to the existing "query multiple" problem; rather than using "... IN (?, ?, ?, ...)", this uses ARRAYs and does "... = ANY(?)". The query plan is identical, ~~but we have not used java.sql.Array for selection before so this is an experiment~~. (We [have actually used this before](https://github.com/isaacphysics/isaac-api/blob/v3.12.3/src/main/java/uk/ac/cam/cl/dtg/isaac/dao/GameboardPersistenceManager.java#L842)!)

---

There are a few other cleanup commits to either use consistent methods or remove unnecessary code.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
@@ -301,7 +301,7 @@ public class GroupsFacade extends AbstractSegueFacade {
 
             RegisteredUserDTO userOfInterest = userManager.getUserDTOById(userId);
 
-            List<UserGroupDTO> groups = groupManager.getGroupsByOwner(userOfInterest);
+            List<UserGroupDTO> groups = groupManager.getAllGroupsOwnedAndManagedByUser(userOfInterest, false);
             return Response.ok(groups).cacheControl(getCacheControl(NEVER_CACHE_WITHOUT_ETAG_CHECK, false)).build();
         } catch (NoUserLoggedInException e) {
             return SegueErrorResponse.getNotLoggedInResponse();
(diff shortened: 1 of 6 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/api/GroupsFacade.java
```diff
@@ -301,7 +301,7 @@ public Response getGroupsForGivenUser(@Context final HttpServletRequest request,
 
             RegisteredUserDTO userOfInterest = userManager.getUserDTOById(userId);
 
-            List<UserGroupDTO> groups = groupManager.getGroupsByOwner(userOfInterest);
+            List<UserGroupDTO> groups = groupManager.getAllGroupsOwnedAndManagedByUser(userOfInterest, false);
```
jsharkey13, line 304: The API doc said this got all groups owned or managed by the user, but it did not. We do not use this endpoint anywhere in the frontend at present, so it may be a candidate for removal.

skyepurchase (approved):

PR #574, opened 2024-01-26
Ada improved teacher signup flow

mwtrew:
---

**Pull Request Check List**
- [x] Peer-Reviewed


src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -496,13 +503,31 @@ public RegisteredUser createUserSession(final HttpServletRequest request, final
...
+            IOException {
+        Map<String, String> session = getSegueSessionFromRequest(request);
+        String caveats = session.get(SESSION_CAVEATS);
+
+        if (null == caveats) {
+            return null;
+        }
+
+        ArrayList<String> caveatFlags = serializationMapper.readValue(caveats, new TypeReference<ArrayList<String>>(){});
+        caveatFlags.remove(caveatToRemove.toString());
```
jsharkey13, line 530: Should we do something if the session did not have such a caveat? If we were in that situation, something would be very wrong and it would seem worthwhile noting it?

`.remove()` returns false if nothing was removed, so that might be an option.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -922,12 +949,12 @@ private void createSession(final HttpServletRequest request, final HttpServletRe
             sessionInformationBuilder.put(SESSION_TOKEN, userSessionToken);
             sessionInformationBuilder.put(DATE_EXPIRES, sessionExpiryDate);
 
-            if (partialLoginFlag) {
-                partialLoginFlagString = String.valueOf(true);
-                sessionInformationBuilder.put(PARTIAL_LOGIN_FLAG, partialLoginFlagString);
+            for (AuthenticationCaveat c : authenticationCaveats) {
+                caveatFlags.add(c.toString());
```
jsharkey13, line 959: `authenticationCaveats` is a Set, which you then iterate over to create the list `caveatFlags`. There's no guarantee of ordering here, which matters when this is passed into `calculateSessionHMAC` which then uses the list in order and will fail to recognise two identical sets of caveats if the Set happens to order them differently sometimes.

Whilst at present there is at most one caveat, if we're going to write it in a way that accepts more than one caveat, we might as well future-proof it. Something like:
```java
caveatFlags = authenticationCaveats.stream().map(AuthenticationCaveat::toString).sorted().collect(Collectors.toList());
```
would at least impose a consistent order. Better, perhaps, would be to make `calculateSessionHMAC` take a Set and do the ordering itself so we don't have to worry about all places that use it?

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -1020,19 +1047,21 @@ private boolean isValidUsersSession(final Map<String, String> sessionInformation
...
      * @param sessionToken
      *            - a token allowing session invalidation
-     * @param partialLoginFlag
+     * @param caveatFlags
      *            - Boolean data to encode in the cookie - true if a partial login
      * @return HMAC signature.
      */
     private String calculateSessionHMAC(final String key, final String userId, final String currentDate, final String sessionToken,
-                                        @Nullable final String partialLoginFlag) {
+                                        @Nullable final ArrayList<String> caveatFlags) {
```
jsharkey13, line 1061: `ArrayList` seems needlessly specific? Can this not be `List` or `Collection`?

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
```diff
@@ -1020,19 +1047,21 @@ private boolean isValidUsersSession(final Map<String, String> sessionInformation
...
         StringBuilder sb = new StringBuilder();
         sb.append(userId);
         sb.append("|").append(currentDate);
         sb.append("|").append(sessionToken);
 
-        if (partialLoginFlag != null) {
-            sb.append("|").append(partialLoginFlag);
+        if (null != caveatFlags) {
+            for (String c : caveatFlags) {
```
jsharkey13, line 1068: This List was generated from a Set on at least one code path on the way here; it will not recognise two identical sets of caveats if the order happens to be different. If we're going to allow multiple caveats by using a List/Set/Collection, we should do so in a future-proof way.

Perhaps sorting the `caveatFlags` collection just before this loop is simplest and most effective?

jsharkey13 (commented):
Nothing major, I don't think.

Changing the cookie key from `PARTIAL_LOGIN_FLAG` to `SESSION_CAVEATS` will break anyone in the process of logging in this way; but since those cookies last only 20 mins and are currently only staff, I think not worrying about backwards incompatibility is okay.

We only want one caveat at a time right now, but some of the code is written in a way that would allow multiple caveats. However, it will not behave in the expected manner if multiple caveats were actually used. (The use of Sets but then adding items from the set to a HMAC in an order will fail the HMAC check randomly). There are some comments about this.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
mwtrew: >Better, perhaps, would be to make calculateSessionHMAC take a Set and do the ordering itself so we don't have to worry about all places that use it?
Yep, I think that makes the most sense.

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/UserAuthenticationManager.java
mwtrew: Agreed, this shouldn't happen so I'll log it as a warning.

PR #575, opened 2024-02-01
Remove unused statistics methods and database code

jsharkey13:
I am attempting to clean up things that access `logged_events`. At the end of this PR, there are precisely three methods that touch the logged events table, reflecting the three current use-cases in the API: to log a new event; to merge anonymous log events into a registered user; and to count logs by type for the stats page.

All other code paths that used `logged_events` were unreachable from the API surface and were mostly historic analysis code. This PR removes those methods, which did school analysis and analysis of log event counts over time. For performance and separation-of-concerns reasons, we no longer do any form of analysis like this using the live API.


skyepurchase (commented):
LGTM. The deletion is safe and consistent. I have not checked whether these methods should be deleted or whether more methods could be deleted.

PR #578, opened 2024-02-08
Purge unused contentIndex parameter

jsharkey13:
We used to use this for the current content SHA, but we have not done so in a long time. Now it is merely `"live"` or `"latest"` - but only the ES related classes (`AdminFacade` and `GitContentManager`) should know what the content version is.
Using it elsewhere is likely to lead to bugs; as in `IsaacController` where it was used in a caching header despite the fact it never changes when content versions change.


skyepurchase (approved):
LGTM. Safe to remove and delete the contentIndex identified. Whether more items can be deleted was not reviewed

PR #583, opened 2024-02-23
Improve warnings on creating duplicate account

jsharkey13:
Some user soft-deletion has caused the existing code-path that checks for duplicate accounts to miss some cases.
This modifies the INSERT query to fail softly, which allows us to distinguish between database errors and (failed) duplicate inserts. It is also apparently more performant:
https://aws.amazon.com/blogs/database/hidden-dangers-of-duplicate-key-violations-in-postgresql-and-how-to-avoid-them/

We already had a `if (pst.executeUpdate() == 0)` check, so this just adjusts the messaging to be consistent with the other pathways. The error message about being unable to set a password was inaccurate, so that has been updated too.


src/main/java/uk/ac/cam/cl/dtg/segue/auth/exceptions/DuplicateAccountException.java
```diff
@@ -23,6 +23,7 @@
  */
 public class DuplicateAccountException extends SegueDatabaseException {
     private static final long serialVersionUID = 7415512495026915539L;
+    public static final String DEFAULT_MESSAGE = "An account with that email address already exists.";
```
jsharkey13, line 26: Should this be `private` instead? I nearly made it in one of the Constants files, but this seemed a more sensible home.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/users/PgUsers.java
```diff
@@ -679,10 +680,12 @@ private RegisteredUser createUser(final RegisteredUser userToCreate) throws Segu
...
-        String query = "INSERT INTO users(family_name, given_name, email, role, date_of_birth, gender," +
-                " registration_date, school_id, school_other, last_updated, email_verification_status, last_seen," +
-                " email_verification_token, email_to_verify, registered_contexts, registered_contexts_last_confirmed, country_code, teacher_account_pending)" +
-                " VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);";
+        String query = "INSERT INTO users(family_name, given_name, email, role, date_of_birth, gender,"
+                + " registration_date, school_id, school_other, last_updated, email_verification_status, last_seen,"
+                + " email_verification_token, email_to_verify, registered_contexts, registered_contexts_last_confirmed,"
+                + " country_code, teacher_account_pending)"
+                + " VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)"
+                + " ON CONFLICT DO NOTHING;";
```
jsharkey13, line 688: This is the important line. We need the `if (pst.executeUpdate() == 0)` that was already in the code below now to check an insert was actually done and error if not.

src/main/java/uk/ac/cam/cl/dtg/segue/auth/exceptions/DuplicateAccountException.java
skyepurchase:
```suggestion
    private static final String DEFAULT_MESSAGE = "An account with that email address already exists.";
```
If `DEFAULT_MESSAGE` is only accessed by this class it should be `private`. As it currently isn't being accessed outside the class I would suggest that the change be made now.

I agree that it should be placed in this class as it is related to this class.

skyepurchase (changes requested):

skyepurchase (approved):

PR #587, opened 2024-03-20
Chained hashing algorithm support

jsharkey13:
This PR adds a new abstract class for chaining password hash algorithms, and one concrete implementation wrapping `SeguePBKDF2v1` inside `SegueSCryptv1`. This allows us to protect the now-weak hashes using the older algorithm with the more secure SCrypt algorithm, _without_ knowing the user's plaintext password to create a new hash by itself. If a user does return and log in, we will compute the inner hash and then the outer hash and compare it to the stored hash; at that point we will upgrade the password automatically as we currently do. This may make login a little slower once for these users, but it is a one-off for much more secure storage.

It also adds a basic admin endpoint to perform in-place upgrades of hashes, so that upgrades do not need to be done by hand or externally to the live service.

This PR is likely easiest to review commit by commit.

(The copyright dates of 2022 are correct; the core chained algorithms part was written back then).

---

**Pull Request Check List**


skyepurchase (changes requested):
LGTM! CodeQL warnings need to be fixed

skyepurchase (approved):

PR #589, opened 2024-03-28
Expire unused accounts

jsharkey13:
This PR adds the scripts required to automatically archive accounts which have been unused for more than 4 years.

The first migration, `2024-02-archived-users.sql` is just a copy of the script that has already run for reference. The cleanup migration is still required.


src/main/resources/db_scripts/scheduled/archive-users-transaction.sql
```diff
@@ -0,0 +1,36 @@
+-- This is NOT scheduled via Quartz, but intended to be run regularly using cron.
```
mwtrew, line 1: Out of curiosity, why not?

src/main/resources/db_scripts/scheduled/archive-users-transaction.sql
```diff
@@ -0,0 +1,36 @@
...
+
+-- Repeatable read is required to ensure the same users are moved into archived as are deleted.
+BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;
```
mwtrew, line 4: I'm not super familiar with [how this works](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-REPEATABLE-READ), but I take it only concurrent updates to the target rows (i.e. users who haven't used the platform in 4 years) will cause a serialisation error? That would be pretty unlikely, but if it's any old update to `users` I could see this failing a lot.

mwtrew (approved):

src/main/resources/db_scripts/scheduled/archive-users-transaction.sql
jsharkey13: Partly a separation of concerns, but mostly so that we have scope for moving the ownership of the archived table to a new Postgres user in the future. If we did that, the API would be unable to read or write it and so couldn't access it via Quartz, which would then protect the archival data from a whole class of vulnerabilities (SQL injection) in the web server.

That was too complex to build right away, but could be dropped in later without changing this script.

src/main/resources/db_scripts/scheduled/archive-users-transaction.sql
jsharkey13: Yes, you are correct. However, in the trial deletion runs we had on average one account a day come back from the older-than-4-years archived ones! It is extraordinarily unlikely that they would come back _during_ the specific deletion run for their account though.

Also, since the initial writing of the query, I moved to an atomic internal method too; we select into a temporary table and then do things on those user IDs. So I think the isolation level is now unnecessary, but I am confident it does no harm (and we may still want to fail if something modifies one of the users we were about to delete).

PR #591, opened 2024-04-04
Return correct response when deleting a user from a group

jacbn:


mwtrew (approved):
Looks good, my only concern was that the original usage of this endpoint (group manager) would no longer work as expected without an updated list of users in the response, but I've tried it and it works fine.

Out of interest, [apparently](https://stackoverflow.com/a/50794127) a `204` response would be conventional here (not worth changing IMO).

jsharkey13:
We do already use `204 No Content` elsewhere for this sort of empty response, though I agree it's not important enough to alter at this point.
And I imagine that RTK Query wasn't using the list of users in the response, but invalidating its cache when making the delete request and then just reloading the list on a response. So previously we were doubling the work, and this is cleaner.

PR #595, opened 2024-04-12
Remove support for user badges

jsharkey13:
We haven't made any use of user badges in the 5 or so years they have been in the codebase; but they have added a performance and maintenance cost. This PR removes the badges and the overhead, returning log management directly to the PgLogManager and removing the slow badge updating which happens during common teacher activities.
It makes no attempt to clean up or remove historic badge data.

Probably wants [the front-end PR](https://github.com/isaacphysics/isaac-react-app/pull/948) merging first, too.


skyepurchase (approved):
LGTM. All the deletions are safe and the API behaves correctly on local. Whether more code could be removed was not checked

PR #597, opened 2024-04-12
Feature/change gameboard completion meaning

mlt47:
The definition of a completed gameboard or assignment is changing from having correctly answered every question on the gameboard to having attempted every question on the gameboard.
As completion is used for filtering and sorting of some paginated endpoints we need to compute this in the back-end.

---

**Pull Request Check List**
- [ ] ~Unit Tests & Regression Tests Added (Optional)~
- [ ] ~Added enough Logging to monitor expected behaviour change~
- [ ] ~Security - Data Exposure - PII is not stored or sent unencrypted~
- [ ] ~Security - Access Control - Check authorisation on every new endpoint~
- [ ] ~Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)~
- [x] Peer-Reviewed


jacbn (approved):

mwtrew (changes requested):
Looking at this together with https://github.com/isaacphysics/isaac-react-app/pull/839. 

The`gameboards/user_gameboards` endpoint will return an error when the [new sort instructions](https://github.com/isaacphysics/isaac-react-app/pull/839/files#diff-ea1c4613584f576f503e71960c25a83756b590f4cd59fa6cef49c4a8c9fdbc18R318) (`attempted` and `correct`) are used, as it still only expects `completion` - you can reproduce this in My Gameboards by setting 'Sort By' to 'Completion...' and 'Show' to anything except 'ALL'. I'm not sure if that's intentional since we're trying to do most of the sorting in the front-end. 

At any rate I'd like to avoid ending up with `attempted`, `correct` and `completion`, where `completion` happens to equal `attempted` - that'll get confusing quickly.

mlt47:
In rebasing my changes on such a long-running branch I have created a merging headache so the meaningful changes have been cherry picked onto the linked PR to avoid the conflicts: https://github.com/isaacphysics/isaac-api/pull/620.
I will close this PR to record comments.

PR #598, opened 2024-04-25
Allow page-level types to be deprecated and superseded

jsharkey13:
This is a change from _any_ content being able to be deprecated, to just top-level page types (i.e. those that extend SeguePage). We did not do anything with deprecated set on any other content, so this change is safe to make and reflects the actual behaviour.
This also moves supersededBy to the SeguePage and subclasses, so that more than just question pages can be superseded.
It also tidies up other classes that had one or both of these properties but where we were not using them. A similar change to the frontend types exists to demonstrate that we only used these for SeguePage level objects.

I have tested that this has not broken any existing content, and also verified that ContentSummaryDTOs still get the two properties mapped across as expected.


jsharkey13:
This is probably where the properties should have been added in the first place. I am still a little worried about odd cases like wildcards or past question attempts where `deprecated` might have gotten into the database and now be unable to be un-mapped. I believe wildcards are the only problematic case where we used an unconfigured ObjectMapper which included null values, however!

jsharkey13:
This goes with https://github.com/isaacphysics/isaac-react-app/pull/963, although neither depends on the other! They both confirm that the move is safe to make, however.

skyepurchase (changes requested):
LGTM. But, please fix the CodeQL warnings

skyepurchase (approved):

PR #599, opened 2024-04-26
Load question attempts via page_id index

jsharkey13:
Before we were using prefix matching on the question_id column, which sometimes meant that the index could not be used, or caused issues with questions like "book_c1_q1" where there are dozens of other questions which start with the same prefix ("book_c1_q10", etc) that were loaded
unnecessarily.
The repeated "OR question_id LIKE ..." did not perform well for large numbers of question IDs, either.
The page_id column is easy to match with an equals match of the page IDs, and the new index is much smaller than the question_id index which should help it perform better too.

---

This cannot be merged until all the existing question attempts have the `page_id` value backfilled and the index is created, which I am working on.


jsharkey13:
The data is backfilled, the index has been made. We are good to go 🚀

mwtrew (approved):

PR #607, opened 2024-05-15
Support pagination and attempt data in question finder

jacbn:
- Augment search results for the new question finder with question attempt information;
- Support pagination, looping up to 5 times to try to pull `SEARCH_RESULTS_PER_PAGE = 30` questions in an attempt to match filters which cannot otherwise be checked for exclusively in ES.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
@@ -36,6 +36,7 @@ import uk.ac.cam.cl.dtg.isaac.dto.IsaacPageFragmentDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacQuestionPageDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.IsaacTopicSummaryPageDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.ResultsWrapper;
+import uk.ac.cam.cl.dtg.isaac.dto.SearchResultsWrapper;
 import uk.ac.cam.cl.dtg.isaac.dto.SegueErrorResponse;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentBaseDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentDTO;
@@ -54,6 +55,7 @@ import uk.ac.cam.cl.dtg.segue.dao.content.ContentManagerException;
 import uk.ac.cam.cl.dtg.segue.dao.content.GitContentManager;
 import uk.ac.cam.cl.dtg.util.AbstractConfigLoader;
 
+import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.ws.rs.DefaultValue;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
```diff
@@ -388,40 +381,75 @@ public final Response getQuestionList(@Context final Request request,
...
+                    for (ContentSummaryDTO result : summarizedResults) {
+                        augmentContentSummaryWithAttemptInformation(result, questionAttempts);
+                    }
+
+                    if (hideCompleted) {
+                        summarizedResults = summarizedResults.stream()
+                                .filter(q -> q.getCorrect() == null || !q.getCorrect())
+                                .collect(Collectors.toList());
+                    }
+                }
```
jsharkey13, line 433: This is the sort of logic that is supposed to live in Manager classes, not Facade classes. But it's currently not clear to me which manager this is best suited to be in, and there was already a lot of logic in this method beforehand :/

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -842,19 +842,18 @@ private GameboardDTO augmentGameboardWithQuestionAttemptInformation(final Gamebo
...
         return questions.stream()
                 .map(q -> this.gameboardPersistenceManager.convertToGameboardItem(
                         q, new GameboardContentDescriptor(q.getId(), QUESTION_TYPE, AudienceContext.fromFilter(gameFilter))))
-                .map(questionItem -> {
+                .peek(questionItem -> {
                     try {
                         this.augmentGameItemWithAttemptInformation(questionItem, userQuestionAttempts);
                     } catch (ContentManagerException | ResourceNotFoundException e) {
                         log.error("Unable to augment '" + questionItem.getId() + "' with user attempt information");
                     }
```
skyepurchase, line 856:
```suggestion
                .map(questionItem -> {
                    try {
                        this.augmentGameItemWithAttemptInformation(questionItem, userQuestionAttempts);
                    } catch (ContentManagerException | ResourceNotFoundException e) {
                        log.error("Unable to augment '" + questionItem.getId() + "' with user attempt information");
                    }
                    return questionItem;
```
`peek` is mainly used for debugging purposes "peeking" at the contents of the pipeline as they go past and can be optimised away. [The javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#peek(java.util.function.Consumer)) and a [discussion on stack overflow](https://stackoverflow.com/questions/44370676/java-8-peek-vs-map).

I think IntelliJ thinks the original map just returns the object and so assumes this is some peek operation. We are augmenting the object (as the function is helpfully named) and so map is more sensible (see [non interference referenced in the doc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/package-summary.html#NonInterference)). IntelliJ wants peek because it is "more efficient" than map in that it doesn't create a new object in the pipeline, but that is precisely what we are doing.

skyepurchase (changes requested):
LGTM. I think the `peek` change is an erroneous IntelliJ suggestion

jsharkey13 (commented):
There are a few things here I think could be clearer, and a few things to tidy up.

There are bigger questions raised about how much of this logic should be in a manager class, and also about the _many_ different but equivalent augmentation functions, but I don't necessarily think this is the time to address those!

jacbn:
Changes above should all be made; I have moved the user attempt business logic into a new class, `UserAttemptManager`; name / file path change suggestions appreciated!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
```diff
@@ -388,40 +389,72 @@ public final Response getQuestionList(@Context final Request request,
...
-                    "Error locating the content requested", e1);
-            log.error(error.getErrorMessage(), e1);
-            return error.toResponse();
+                summarizedResults = extractContentSummaryFromList(c.getResults());;
+
+                if (summarizedResults.isEmpty()) {
+                    break;
+                }
+
+                nextSearchStartIndex += summarizedResults.size(); // add the total number of returned results, including any we will filter out
```
jsharkey13, line 425: This can do the wrong thing. Say there are 65 questions matching the search and we have a page size of 30. If the user has answered 5 of the first page, we will load the next 30 and increment `nextSearchStartIndex` from 30 to 60. But we will then throw away 25 of these, since we only need 5 to fill the list. The next request will start at an index of 60, and those middle 25 will never be able to be loaded.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/UserAttemptManager.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Augment a list of content summary objects with user attempt information.
+     * @param user the user to augment the content summary list for.
+     * @param summarizedResults the content summary list to augment.
+     * @param hideCompleted whether to filter out completed content.
+     * @return the augmented content summary list.
+     * @throws SegueDatabaseException if there is an error retrieving question attempts.
+     */
+    public List<ContentSummaryDTO> augmentContentSummaryListWithAttemptInformation(RegisteredUserDTO user, List<ContentSummaryDTO> summarizedResults, boolean hideCompleted) throws SegueDatabaseException {
```
jsharkey13, line 92: The filtering feels a bit odd to be inside this function, since it is not obviously related to the rest of the method and can be (and is) done separately after the function with the results.
I think it is fine for now, though!

jsharkey13 (commented):
The new `UserAttemptManager` looks okay to me (barring one comment about one of the methods); I also cannot think of a better name!

skyepurchase (approved):

PR #608, opened 2024-05-22
Add lock on anonymous user ID when attempting questions

jacbn:
Since anonymous accounts' question attempts are stored as a single row in postgres, if two threads simultaneously try to write (in this case, caused by simultaneous submission of inline parts), a race condition occurs and writes can be lost. This meant all attempt values returned correctly to the frontend on submission, but since they were not saved in the database some updates were lost on reload.

This commit adds a `pg_advisory_xact_lock` on the anonymous user ID to prevent the race.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
@@ -78,8 +78,10 @@ public class PgQuestionAttempts implements IQuestionAttemptManager {
     public void registerAnonymousQuestionAttempt(final String userId, final String questionPageId,
             final String fullQuestionId, final QuestionValidationResponse questionAttempt)
             throws SegueDatabaseException {
+
+        // Anonymous question attempts are stored in a JSONB object, need to atomically read-and-update this object:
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
```diff
@@ -78,7 +78,15 @@ public PgQuestionAttempts(final PostgresSqlDb ds, final ContentMapper objectMapp
...
             final String fullQuestionId, final QuestionValidationResponse questionAttempt)
             throws SegueDatabaseException {
+
+        // Anonymous question attempts are stored in a JSONB object, need to atomically read-and-update this object:
         try (Connection conn = database.getDatabaseConnection()) {
+            conn.setAutoCommit(false);  // Start a transaction to hold a lock open during.
+            try (PreparedStatement pst = conn.prepareStatement("SELECT pg_advisory_xact_lock(?)")) {
+                pst.setLong(1, userId.hashCode());
+                pst.executeQuery();
+            }
```
mlt47, line 88: As grabbing a lock is blocking I think we should explicitly set a lock timeout: https://stackoverflow.com/questions/20963450/controlling-duration-of-postgresql-lock-waits/20963803#20963803

PR #610, opened 2024-06-05
Remove fuzzy search from Question Finder

skyepurchase:


src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -229,7 +229,10 @@
 
             if (null == searchResults || searchResults.isEmpty()) {
                 if (!failQuietly) {
-                    log.error(String.format("Failed to locate content with ID '%s' in the cache for content SHA (%s)", id, getCurrentContentSHA()));
+                    log.error(String.format(
+                            "Failed to locate content with ID '%s' in the cache for content SHA (%s)",
+                            id, getCurrentContentSHA()
+                    ));
```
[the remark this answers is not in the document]
skyepurchase: This is not a new alert, it will be reviewed in future work!

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -326,7 +330,7 @@ public final ResultsWrapper<ContentDTO> searchForContent(
             @Nullable final Set<String> topics, @Nullable final Set<String> books,
             @Nullable Set<String> levels, @Nullable Set<String> stages, @Nullable Set<String> difficulties,
             @Nullable Set<String> examBoards, final Set<String> contentTypes, final Integer startIndex,
-            final Integer limit, final boolean showNoFilterContent
+            final Integer limit, final boolean showNoFilterContent, final boolean fuzzySearch
```
mwtrew, line 333: This is making me wonder if having global search and the finder/builder use the same method (`searchForContent`) is still the right thing to do, because it's getting quite cyclomatically complex based on the differing needs of each path (prior to [this commit](https://github.com/isaacphysics/isaac-api/commit/1e14c07341c3581417ed12c620bcf1247fd54a7d) they were separate). I can see them diverging further in future.

I think the value of `IsaacSearchInstructionBuilder` is being able to tell at a high level what we're searching for, and having a lot of dynamic stuff detracts from its' readability. What do you think @skyepurchase @mlt47 ?

mwtrew (changes requested):
Apart from the below query about our general approach (not explicitly to do with this PR but it is unfortunately related, sorry), it would be nice to see a test case that illustrates what we're trying to achieve by removing fuzzy search and whether this does the job. Some old/new search results on the Trello card would do.

(P.S thanks for fixing all the formatting!)

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
skyepurchase: This sounds sensible. The plan would be to create a `siteWideSearch` function for `\search` endpoints and a `questionSearch` for `\questions` endpoints similar to the commit you shared. This should reduce the complexity of the functions and prevent unnecessary parameters being passed.

There's potentially a case for making the `SearchInstructionBuilder` more higher level, or having a higher level abstraction on top. But I don't that is necessary at the moment.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/IsaacController.java
```diff
@@ -360,29 +374,30 @@ public final Response getDocumentByPath(@Context final Request request, @Context
...
             if (null == fileContent) {
                 String refererHeader = httpServletRequest.getHeader("Referer");
-                SegueErrorResponse error = new SegueErrorResponse(Status.NOT_FOUND, "Unable to locate the file: " + path);
+                SegueErrorResponse error = new SegueErrorResponse(
+                        Status.NOT_FOUND, "Unable to locate the file: " + path);
                 log.warn(String.format("Unable to locate the file: (%s). Referer: (%s)", path, refererHeader));
                 return error.toResponse(getCacheControl(NUMBER_SECONDS_IN_TEN_MINUTES, false), etag);
             }
 
+            RegisteredUserDTO currentlyLoggedInUser = userManager.getCurrentRegisteredUser(httpServletRequest);
```
mwtrew, line 394: It makes sense to check this early on like we did before, otherwise we're doing unnecessary work for a user who isn't logged in.

src/main/java/uk/ac/cam/cl/dtg/segue/search/IsaacSearchInstructionBuilder.java
```diff
@@ -324,24 +333,33 @@ private void addSearchesInFieldsToInstruction(final BooleanInstruction instructi
...
+                        generatedSubInstructions.add(
+                                new WildcardInstruction(searchInField.getField(), "*" + term + "*", boost));
                         // Use a multi-match instruction, and ensure multi-match instructions for a particular term are
                         // grouped together
                         multiMatchSearchesGroupedByTerm.putIfAbsent(term, Sets.newHashSet());
                         multiMatchSearchesGroupedByTerm.get(term).add(searchInField.getField());
 
                     } else if (searchInField.getStrategy() == Strategy.SIMPLE) {
-                        generatedSubInstructions.add(new MatchInstruction(searchInField.getField(), term));
+                        Long boost = searchInField.getPriority() == Priority.HIGH ? HIGH_PRIORITY_FIELD_BOOST : 1L;
```
mwtrew, line 359: I'd make the normal priority boost a constant like we have for the others.

mwtrew (approved):
Another couple of small comments, but looking great overall and very happy to see some integration tests!

PR #614, opened 2024-06-12
Correctly capitalise LLM acronym in class and type names

mlt47:
Update naming of classes containing the LLM acronym to align with the codebase's convention.


jsharkey13 (approved):
😍

PR #615, opened 2024-06-17
Sanitise contact form

skyepurchase:
**Pull Request Check List**

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/ContactFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/ContactFacade.java
@@ -24,6 +24,8 @@ import org.apache.commons.text.StringEscapeUtils;
 import org.apache.commons.validator.routines.EmailValidator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


mwtrew (approved):

src/main/java/uk/ac/cam/cl/dtg/segue/api/ContactFacade.java
```diff
@@ -128,13 +129,18 @@ public Response contactUs(final Map<String, String> form, @Context final HttpSer
                             .put("contactUserRole", currentUserRole)
                             .put("contactEmail", form.get("emailAddress"))
                             .put("contactSubject", form.get("subject"))
-                            .put("contactMessage", message)
+                            .put("contactMessage", StringEscapeUtils.escapeHtml4(message))
```
jsharkey13, line 132: This feels like it double-escapes things, since we already call `StringEscapeUtils.escapeHtml4` in `EmailManager::sanitizeEmailParameters` on this property :/

jsharkey13:
I have fixed the underlying bug in our ticketing system that meant double-encoding was necessary, so Skye's original solution didn't quite work. We also need the newlines to be preserved and not escaped.

PR #627, opened 2024-08-12
Add migration to set default assignment email preference

mwtrew:
---

**Pull Request Check List**
- ~~Security - Data Exposure - Test any altered or created endpoints using [swagger](http://localhost:8080/isaac-api/api-docs/)~~
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/resources/db_scripts/migrations/2024-08-set_default_email_preference_ada.sql
+++ b/src/main/resources/db_scripts/migrations/2024-08-set_default_email_preference_ada.sql
@@ -1,0 +1,7 @@
+INSERT INTO public.user_preferences (user_id, preference_type, preference_name, preference_value, last_updated)
+SELECT DISTINCT user_id, 'EMAIL_PREFERENCE' AS preference_type, 'ASSIGNMENTS' AS preference_name, true AS preference_value, now() as last_updated
+FROM public.user_preferences
+WHERE user_id NOT IN (
+    SELECT user_id FROM public.user_preferences
+    WHERE preference_type = 'EMAIL_PREFERENCE'
+);


src/main/resources/db_scripts/migrations/2024-08-set_default_email_preference_ada.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO public.user_preferences (user_id, preference_type, preference_name, preference_value, last_updated)
+SELECT DISTINCT user_id, 'EMAIL_PREFERENCE' AS preference_type, 'ASSIGNMENTS' AS preference_name, true AS preference_value, now() as last_updated
+FROM public.user_preferences
```
jsharkey13, line 3: This will only select users who already have some other preference on their account. Users who do not have any form of preference set won't be included.
You can swap the `FROM user_preferences` to `FROM users` to fix this, and change the column name selected to `id` not `user_id` - which has the added benefit of being unique now and so no `SELECT DISTINCT` horribleness necessary!

I.e. the query boils down to:
```sql
INSERT INTO user_preferences
SELECT id, 'EMAIL_PREFERENCE', 'ASSIGNMENTS', true, CURRENT_TIMESTAMP
FROM users
WHERE users.id NOT IN (SELECT user_id FROM user_preferences WHERE preference_type='EMAIL_PREFERENCE');
```
when you also skip the `public` scoping unnecessary on live (there's only one schema with these names in on live).

mwtrew: Ah, thanks for catching that!

skyepurchase (approved):

PR #628, opened 2024-08-15
Update dependencies

mwtrew:
Also removes Twitter authenticator, which is now unused and had an out of date SDK.


pom.xml
```diff
@@ -11,25 +11,25 @@
...
+        <jackson.version>2.17.2</jackson.version>
         <jackson-databind.version>2.17.0</jackson-databind.version>
         <powermock.version>2.0.9</powermock.version>
         <junit-4.version>4.13.2</junit-4.version>
-        <junit-5.version>5.10.2</junit-5.version>
+        <junit-5.version>5.10.3</junit-5.version>
         <swagger-ui-version>4.13.2</swagger-ui-version>
         <prometheus.version>0.16.0</prometheus.version>
-        <jetty-version>11.0.20</jetty-version>
+        <jetty-version>11.0.22</jetty-version>
```
jsharkey13, line 25: This change needs a corresponding change to the two Dockerfiles: `Dockerfile-api` and `Dockerfile-etl` otherwise things might break when deployed.

pom.xml
```diff
@@ -451,7 +445,7 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcprov-jdk18on</artifactId>
-            <version>1.78.1</version>
+            <version>1.78</version>
```
jsharkey13, line 448: Is this not a downgrade? `1.78.1` was released after `1.78` and seems to [fix some issues](https://www.bouncycastle.org/resources/java-release-1-78-1-is-now-available-for-download/).

jsharkey13 (changes requested):

jsharkey13 (approved):

PR #629, opened 2024-08-16
Implement question completion status filtering

skyepurchase:
ContentSummaryDTOs now use an enum for a new `state` attribute rather than a boolean `correct` attribute.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
@@ -76,6 +76,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
```diff
@@ -381,6 +382,21 @@ public final Response getQuestionList(@Context final Request request,
             }
         }
 
+        try {
+            if (Objects.equals(statuses, "")) {
```
jsharkey13, line 386: If statuses is `null`, this won't catch it and it will cause an unhandled NPE in the else when it tries to split it. 
```suggestion
            if (null == statuses || Objects.equals(statuses, "")) {
```

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/UserAttemptManager.java
```diff
@@ -78,7 +82,15 @@ private void augmentContentSummaryWithAttemptInformation(
...
         }
-        contentSummary.setCorrect(questionAnsweredCorrectly);
+        if (attempted) {
+            if (questionAnsweredCorrectly) {
+                contentSummary.setState(ALL_CORRECT);
+            } else {
+                contentSummary.setState(IN_PROGRESS);
+            }
+        } else {
+            contentSummary.setState(NOT_ATTEMPTED);
```
jsharkey13, line 92: I think these would be better scoped, especially since we have another Enum with some of the same item names and it's best to be clear which is which.
`CompletionState.ALL_CORRECT`, etc, and change the import to remove the `.*` on the end.

jsharkey13 (approved):

PR #630, opened 2024-08-20
Fix inline question marking for content summaries

jsharkey13:
We (still) have two different ways of working out whether a question has been answered correctly. This removes some of the duplication, to fix an issue where inline regions were overlooked when marking a question using the content summary way of marking.

The gameboard-based methods had already been updated to include inline regions, but `collateQuestionPartIds(...)` did not include them in its collation.


mlt47 (approved):
This should achieve the same result as the removed code (but also account for inline questions).

PR #631, opened 2024-08-22
Add char limit to QF search

axlewin:
Apply the 1000 character limit from the main search to the new question finder search.


mwtrew (approved):
Looks good - for future reference a test proving that it works as expected (e.g. in `PagesFacadeIT`) would make reviewing a bit quicker.

PR #634, opened 2024-08-23
Support anonymous users in question finder

jsharkey13:
This PR is mostly to add support for anonymous users in question finder augmentation and filtering behaviour.

There are some performance optimisations (don't load the user multiple times, avoid filtering if the filter is equivalent "everything"), and one key behavior change (don't augment results if `statuses` is not specified, which restores the old behaviour useful for the gameboard builder).

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
@@ -75,7 +75,11 @@ public final class Constants {
     public enum CompletionState {
         ALL_CORRECT, IN_PROGRESS, NOT_ATTEMPTED;
 
-        public static final Set<CompletionState> ALL_STATES = Set.of(CompletionState.values());
+        private static final Set<CompletionState> allStates = Set.of(CompletionState.values());
+
+        public static Set<CompletionState> getAllStates() {
+            return allStates;
+        }
     }
 
     public enum QuestionPartState {
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
@@ -73,7 +73,9 @@ public final class Constants {
     }
 
     public enum CompletionState {
-        ALL_CORRECT, IN_PROGRESS, NOT_ATTEMPTED
+        ALL_CORRECT, IN_PROGRESS, NOT_ATTEMPTED;
+
+        public static final Set<CompletionState> ALL_STATES = Set.of(CompletionState.values());
     }
 
     public enum QuestionPartState {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
```diff
@@ -73,7 +73,9 @@ public enum GameboardItemState {
     }
 
     public enum CompletionState {
-        ALL_CORRECT, IN_PROGRESS, NOT_ATTEMPTED
+        ALL_CORRECT, IN_PROGRESS, NOT_ATTEMPTED;
+
+        public static final Set<CompletionState> ALL_STATES = Set.of(CompletionState.values());
```
jsharkey13, line 78: I cannot decide if this is 👌 or 🤢. Clearly the compiler can tell the difference between the enum values and this static attribute, but in the code it's not at all obvious. Should I make it private (I don't want to create it again and again) and use a getter?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/PagesFacade.java
```diff
@@ -388,28 +388,29 @@ public final Response getQuestionList(@Context final Request request,
         }
 
         try {
-            if (null == statuses || Objects.equals(statuses, "")) {
-                // If no statuses apply assume all statuses
-                filterByStatuses = Set.of(CompletionState.values());
+            if (null == statuses) {
+                filterByStatuses = null;
+            } else if (statuses.isEmpty()) {
+                filterByStatuses = CompletionState.ALL_STATES;
```
jsharkey13, line 394: Does this need explaining, or is it obvious?

mlt47: I think a comment to explain that this is used to determine whether or not to augment the results with question attempt data would be worth it. I wouldn't say that would be expected/obvious!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/Constants.java
mlt47: Tricky. To reduce the ambiguity a little while reading it could maybe be called `ALL_STATES_SET` (although I don't generally like having the name dictate the type)? Like you said compiler/IDE will help make sure it doesn't lead to bugs.
I've also heard that EnumSet is even cooler than Set!

mlt47: A static getter method would probably be the way to go if you want to eliminate ambiguity.

mlt47 (approved):
Looks good.
For your Enum question I'd go with a static method that returns the set.

mlt47 (approved):

PR #635, opened 2024-08-23
Allows equivalent answers with different "requiresExactMatch" values to have different feedback

sjd210:
Currently a `SYMBOLIC` match to an `EXACT` answer will supersede all later matches to all other `logicFormulaChoice`s (therefore always providing the default feedback). 

This change allows for a special case with custom feedback to be given on questions with both `"requiresExactMatch": true` and `"requiresExactMatch": false` choices with the same answer.

e.g. [Regression Test Page](https://staging.isaacphysics.org/questions/_regression_test_?stage=all) (Part H)


jsharkey13 (approved):
Without a lot of test cases, it's hard to be sure this is always fine; but it seems reasonable! Let's try it out!

PR #636, opened 2024-08-27
Hide superseded content in QF for non-teachers

jsharkey13:
This is a compromise, since the endpoint is also used for the gameboard builder where superseded questions are desired. We may review who sees superseded questions where later.


src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -438,6 +438,9 @@ public final ResultsWrapper<ContentDTO> questionSearch(
         IsaacSearchInstructionBuilder searchInstructionBuilder = new IsaacSearchInstructionBuilder(
                 searchProvider, this.showOnlyPublishedContent, this.hideRegressionTestContent, !showNoFilterContent)
```
mwtrew, line 439: The other similar options e.g. `hideRegressionTestContent` are passed in as parameters to the constructor, I'd lean towards doing the same for consistency (or change the lot if there's a good reason to).

mwtrew (approved):
Looks good, added a couple of tests to confirm.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
jsharkey13: Yes, I didn't like the endless boolean constructor params, especially given the builder pattern the rest of the class uses. So I made this one consistent with the builder pattern and it would be nice to remove the rest from the constructor but that didn't seem important enough now.

PR #638, opened 2024-09-03
Add LLM marking formula AST classes

mlt47:
This PR will need to go in at least one release cycle before we add the classical computing method of summing LLM question marks so that our two running APIs and ETL can process the new content schema.

If there are concerns with the approach, I could write an ADR and we can discuss as a group sometime this week. The question type is live and functioning, it would just be an improved experience once we take away the LLM's requirement to do basic maths!

---

**Pull Request Check List**
- [ ] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
@@ -36,6 +36,7 @@ import uk.ac.cam.cl.dtg.isaac.dos.content.ContentBase;
 import uk.ac.cam.cl.dtg.isaac.dos.content.DTOMapping;
 import uk.ac.cam.cl.dtg.isaac.dos.content.Item;
 import uk.ac.cam.cl.dtg.isaac.dos.content.JsonContentType;
+import uk.ac.cam.cl.dtg.isaac.dos.content.LLMMarkingExpression;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentBaseDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentSummaryDTO;
@@ -343,6 +344,9 @@ public class ContentMapper {
      * @return ObjectMapper that has been configured to handle the segue recursive object model.
      */
     public ObjectMapper generateNewPreconfiguredContentMapper() {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+
         ContentBaseDeserializer contentDeserializer = new ContentBaseDeserializer();
         contentDeserializer.registerTypeMap(jsonTypes);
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
```diff
@@ -353,20 +357,20 @@ public ObjectMapper generateNewPreconfiguredContentMapper() {
...
+                new IsaacQuestionBaseDeserializer(contentDeserializer);
 
         SimpleModule contentDeserializerModule = new SimpleModule("ContentDeserializerModule");
         contentDeserializerModule.addDeserializer(ContentBase.class, contentDeserializer);
         contentDeserializerModule.addDeserializer(IsaacQuestionBase.class, isaacQuestionBaseDeserializer);
         contentDeserializerModule.addDeserializer(Choice.class, choiceDeserializer);
         contentDeserializerModule.addDeserializer(Item.class, itemDeserializer);
         contentDeserializerModule.addDeserializer(QuestionValidationResponse.class, validationResponseDeserializer);
+        contentDeserializerModule.addDeserializer(
+                LLMMarkingExpression.class, new LLMMarkingExpressionDeserializer(objectMapper));
```
mlt47, line 373: The ability to pass in a shared object mapper to all of these deserializers looks to be a pattern that could remove the need to order these deserializers in a special way to try to handle our recursive content model. We are only registering the deserializers for classes at this stage, the object mapper will handle the recursive content just fine when it comes to actually deserializing.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/ContentMapper.java
```diff
@@ -353,20 +357,20 @@ public ObjectMapper generateNewPreconfiguredContentMapper() {
...
```
jsharkey13, line 372: Why does this need its own `ObjectMapper`? The other deserialisers all just use the jsonParser provided as first argument to the method calls:
```java
ObjectMapper mapper = (ObjectMapper) jsonParser.getCodec();
```

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/LLMMarkingExpressionDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+    private final ObjectMapper objectMapper;
+
+    public LLMMarkingExpressionDeserializer(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public LLMMarkingExpression deserialize(final JsonParser jsonParser, final DeserializationContext deserializationContext)
+            throws IOException {
+        ObjectNode root = objectMapper.readTree(jsonParser);
```
jsharkey13, line 26: I.e. why not just do `
```suggestion
        ObjectMapper mapper = (ObjectMapper) jsonParser.getCodec();
        ObjectNode root = objectMapper.readTree(jsonParser);
```
like all the other deserialisers do?

jsharkey13 (approved):

PR #639, opened 2024-09-11
Allow valid permutations of molecules in api

sjd210:
Passes the `allowPermutations` flag from the Content Editor to the Chemistry Checker

---

Companion PR to [Allow valid permutations of molecules](https://github.com/isaacphysics/chemistry-checker-js/pull/1) for the Chemistry Checker.

(This is based on the unmerged [new-chemistry-checker-server](https://github.com/isaacphysics/isaac-api/tree/feature/new-chemistry-checker-server) branch, so that should probably be approved and merged in before this is).


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java
```diff
@@ -366,22 +368,30 @@
 
                 // Weak match to a correct answer.
 
-                if (!((String) closestResponse.get("expectedType")).contains("nuclear")
-                        && closestResponse.get("sameState").equals(false)) {
+                if (closestResponse.get("sameElements").equals(false)) {
```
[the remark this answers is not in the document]
jsharkey13: We check `closestMatch`, but I am not convinced that `closestResponse` is always non-null here; I think the static analysis is right?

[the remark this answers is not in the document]
jsharkey13: Also, it is not at all clear why we care about the state of `closestResponse`; I can believe we care about comparing to the currently-best-matched choice to see if we're better changing it, but I think it needs some explanation if it is right?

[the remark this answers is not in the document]
sjd210: Ah, I see. This seems to be a pre-existing problem then (the line I'm replacing was already using `closestResponse`).

We do always set `closestResponse` when setting `closestMatch` and I don't think there's a way for that response to be null from the checker - but I also don't see the harm in adding a check and explanation. I'll get on that.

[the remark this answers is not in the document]
jsharkey13: [This case](https://github.com/isaacphysics/isaac-api/blob/8c7dd64f37bab1ec56acd2b847c82300540b0af9/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java#L297-L299) was the one I was concerned about; it sets `closestMatch` but not `closestResponse`. Is that in fact a mistake too?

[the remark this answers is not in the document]
sjd210: In that case, we are always caught in [this if statement](https://github.com/isaacphysics/isaac-api/blob/8c7dd64f37bab1ec56acd2b847c82300540b0af9/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java#L331) later and so never reach the "problem" line later in the `else if` chain, so it's not currently possible for it to create an issue (and there is no need to set `closestResponse` as we don't have any automatic feedback to give on an exactly correct answer).

It's probably still suitable to add the check for the sake of future-proofing e.g. maybe at some point we DO want to be able to give feedback on correct answers - but shouldn't be an issue either way right now.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacSymbolicChemistryQuestion.java
```diff
@@ -31,6 +31,8 @@
 public class IsaacSymbolicChemistryQuestion extends IsaacSymbolicQuestion {
     @JsonProperty("isNuclear")
     private boolean isNuclear;
+    @JsonProperty("allowPermutations")
```
jsharkey13, line 34: This isn't necessary! We need it for the line above since Jackson tries to do clever name things with "standard" getters and setters; `isABCDE` is assumed to be the getter for a JSON property called `ABCDE`, whereas we want the JSON property to actually be `isNuclear`.
I'd like to avoid these annotations generally so they don't proliferate unnecessarily like this . . .

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/IsaacSymbolicChemistryQuestionDTO.java
```diff
@@ -25,6 +25,8 @@
 public class IsaacSymbolicChemistryQuestionDTO extends IsaacSymbolicQuestionDTO {
     @JsonProperty("isNuclear")
     private boolean isNuclear;
+    @JsonProperty("allowPermutations")
```
jsharkey13, line 28: Ditto, I believe this in unnecessary!
```suggestion
```

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacSymbolicChemistryQuestion.java
jsharkey13:
```suggestion
```

jsharkey13: (I wondered about "allow" being special, but reading between the lines of https://github.com/FasterXML/jackson-annotations/wiki/Jackson-Annotations#property-inclusion, I think that `get...`, `set...` and `is...` are the only "special" prefixes Jackson cares about).

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java
```diff
@@ -191,6 +191,8 @@ public QuestionValidationResponse validateQuestionResponse(final Question questi
                     req.put("target", formulaChoice.getMhchemExpression());
                     req.put("test", submittedFormula.getMhchemExpression());
                     req.put("description", chemistryQuestion.getId());
+                    req.put("allowPermutations", String.valueOf(chemistryQuestion.allowPermutations()));
```
jsharkey13, line 194: What happens if this is null? Will this produce a string `"null"` rather than `null`?

jsharkey13: (I think we decided that [the checker won't notice either way](https://github.com/isaacphysics/chemistry-checker-js/pull/1/files#diff-0313b3dba60dab2976642eac726c4a1ce1c8d222ad198a13097fbc0c9aa038f5R29), since it expects the string `"true"`).

jsharkey13 (approved):

PR #641, opened 2024-09-13
Improve glossary term caching

jsharkey13:
Glossary terms are expensive to load, send and process. There are many of them, and writing the JSON takes tens of ms.

The HTTP caching wasn't working correctly because whilst we did set a max-age and an ETag, we never used this ETag value if it was sent back.

Separately, loading the (static per content version, identical per user) glossary terms from ElasticSearch each time was a lot of unnecessary computational overhead. Add a cache, similar to the GitContentManager cache, to store glossary terms in. Since the cache key contains the content version, this is always safe to use. After a content push, we will never want the old version again, so the cache lifetime can be quite low (and using expireAfterAccess renews it for each successful load).
Monitor this new cache with Prometheus as for the GCM cache.

These changes do not affect the first user to load the glossary terms per content version, but each subsequent new user will load them ~10x faster. If the ETag remains valid past the max-age, we now use 304 responses to reduce the data sent by 1000x.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/GlossaryFacade.java
@@ -123,7 +123,7 @@ public class GlossaryFacade extends AbstractSegueFacade {
 
         if (null != startIndex) {
             startIndexOfResults = Integer.parseInt(startIndex);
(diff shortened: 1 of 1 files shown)


jacbn (approved):
Looks good. It would be nice to reduce the number of terms we send one day...

PR #643, opened 2024-09-23
Avoid double-loading full gameboard each request

jsharkey13:
We load the gameboard initially not just to check it exists but also to load the question IDs; these are used to speed up loading question attempts a lot.
However, the full gameboard requires going to ElasticSearch and then mapping a bunch of objects between classes. Since we never use anything from these objects that is not in a Lite (unaugmented) gameboard this is entirely wasted work.
The end effect is minor since loading question attempts and the full gameboard still dominate, but I still think worthwhile.


sjd210:
LGTM! We're just getting item IDs and calculating an ETag with this `unAugmentedGameboard` so a lite one will do, and everything seems to work on the page 👍

PR #644, opened 2024-09-25
Avoid unnecessarily reregistering Segue Jobs

jsharkey13:
Currently we reregister jobs on every API start, even if they are unchanged from the version existing in the database. (The code of the job can change without the _job_ changing; the job records the class to execute and the cron record).
Doing so overwrites the last run time of the job, which prevents us from doing sensible things if the API has been offline for a while.

This change only reregisters the job if the job details or cron string change.

It also tests out the IntelliJ recommendation for parameterised log messages, too.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/SegueJobService.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/SegueJobService.java
@@ -35,6 +35,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 
 import static org.quartz.CronScheduleBuilder.cronSchedule;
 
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/SegueJobService.java
```diff
@@ -142,13 +142,18 @@ public void registerScheduleJob(final SegueScheduledJob jobToRegister) throws Sc
...
+        if (!scheduler.checkExists(job.getKey()) || !scheduler.checkExists(cronTrigger.getKey())) {
             scheduler.scheduleJob(job, cronTrigger);
-            log.info(String.format("Registered job (%s) to segue job execution service. Current jobs registered (%s): ", jobToRegister.getJobKey(), localRegisteredJobs.size()));
+            log.info("Registered Quartz job ({}). Current jobs registered ({}): ", jobToRegister.getJobKey(), localRegisteredJobs.size());
         } else {
-            // FIXME - this does not update the job details, e.g. if the SQL file name changes.
-            scheduler.rescheduleJob(cronTrigger.getKey(), cronTrigger);
-            log.info(String.format("Re-registered job (%s) to segue job execution service. Current jobs registered (%s): ", jobToRegister.getJobKey(), localRegisteredJobs.size()));
+            CronTrigger existingTrigger = (CronTrigger) scheduler.getTrigger(cronTrigger.getKey());
+            if (!existingTrigger.equals(cronTrigger)) {
```
mwtrew, line 150: I changed the `cronString` of a job already registered, but `existingTrigger.equals(cronTrigger)` still resolves `true`, so I don't think this is working as expected.

jsharkey13: Thanks for spotting this! 🙌

mwtrew (approved):

PR #649, opened 2024-10-11
Add utm_source parameter to notification emails

jsharkey13:
It might have been possible to do this in the content template, but this feels more reliable. This will help judge traffic from assignment, test, and group join notification emails


sjd210:
LGTM! 👍

PR #651, opened 2024-10-16
Re-word Chemistry warning texts

sjd210:
According to Andrea's feedback:
- “Your answer contains invalid syntax!” → "Your answer is not chemically valid!"
- “Check your elements!” → “Check that you have all the correct atoms present and in the right place!"
- “What type of equation is this?” → “Check your reaction arrow!”

& add exclamation marks to the end of those that do not already have them, for unity e.g. “Your equation is unbalanced.”

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java
@@ -58,6 +58,7 @@ public class IsaacSymbolicChemistryValidator implements IValidator {
 
     private final Set<String> VALID_ERROR_FEEDBACK = Set.of(
             "Division by zero is undefined!",
+            "Check that all atoms have a mass and atomic number!",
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacSymbolicChemistryValidator.java
```diff
@@ -324,9 +324,7 @@ public QuestionValidationResponse validateQuestionResponse(final Question questi
             if (containsError) {
 
                 // User input contains error terms.
-                // FIXME: This currently clashes with determining whether the submitted answer was the wrong type
-                // Inequality should be changed to not allow Nuclear syntax in Chemistry questions and vice versa
-                feedback = new Content("Your answer contains invalid syntax!");
+                feedback = new Content("Your answer is not chemically valid!");
```
jsharkey13, line 327: I've always thought this warning was a bit "in your face" and certain. It's now even more certain, and yet I think that it's just as likely we don't support whatever they are trying to do. For numeric questions we use a softer tone: "Your answer is not in a format we recognise".

We could do something similar here: "Your answer is not in a format we recognise; it might not be chemically valid."?

PR #653, opened 2024-11-06
Wrap exceptions in correct HTTP status responses

jsharkey13:
Since we have `ExceptionMapper<Exception>`, this handler catches almost all exceptions that reach the top-level. However, many of these are things that Jetty would have correctly turned into HTTP responses with the correct status codes and headers. Jetty would not respond with JSON, and so we still want to catch these, but this moves to responding with an appropriate error status code for invalid requests (things that ought to have been 4xx's not 500's).

Instead of adding multiple specific handlers, instead just use this one catch-all since it already existed and already caught everything. This reduces a lot of boilerplate too. It will also correctly handle `405 Method Not Allowed` exceptions; the old handler did not correctly add the required Allow header, so it has been removed.

There are other types of Jackson exception beyond `InvalidFormatException` that Jackson can throw _before_ hitting our facade level code. These do not end up processed by the `UnhandledExceptionMapper`, and instead get returned to the user as unformatted errors. Jackson has a top-level exception `JsonMappingException` that all of its exceptions subclass. Handle that instead, to ensure all are processed.

Notably, this suppresses stack traces in the logs for all these uncaught exceptions. Whilst this is useful to avoid malicious activity from overwhelming our logging, it may make debugging errors more difficult. We could add back stack trace logging to the base-case for unknown and unexpected errors if we find it necessary.


sjd210 (commented):
Other than a couple of typos, I'm quite confident this works. The code makes sense and I've got it working locally :+1:

PR #656, opened 2024-11-12
Handle partial postcodes in admin user search

axlewin:
If a valid first part of a postcode (e.g. CB3) is provided, treat the centroid of this region as the postcode location.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/util/locations/PostCodeIOLocationResolver.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/util/locations/PostCodeIOLocationResolver.java
@@ -199,7 +199,7 @@ public class PostCodeIOLocationResolver implements PostCodeLocationResolver {
             }
         }
 
-        if (completePostCodes.size() > POSTCODEIO_MAX_REQUESTS) {
+        if (completePostCodes.size() + outCodes.size() > POSTCODEIO_MAX_REQUESTS) {
             throw new IllegalArgumentException(String.format("Number of postcodes cannot be bigger than %d!",
                     POSTCODEIO_MAX_REQUESTS));
         }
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/cam/cl/dtg/util/locations/PostCodeIOLocationResolver.java
```diff
@@ -185,40 +186,69 @@ private List<PostCode> carryOutExternalPostCodeServiceRequest(final List<String>
...
+        for (String postCode : unknownPostCodes) {
+            if (postCode.length() > 4) {
+                completePostCodes.add(postCode);
+            }
+            else {
+                outCodes.add(postCode);
+            }
+        }
+
+        if (completePostCodes.size() > POSTCODEIO_MAX_REQUESTS) {
```
jacbn, line 202: We should certainly be checking the number of outCodes against this too.
```suggestion
        if (completePostCodes.size() + outCodes.size() > POSTCODEIO_MAX_REQUESTS) {
```

jacbn (commented):
Looks good. I am not a fan of sending many individual requests but there really is no bulk endpoint equivalent so 🤷. We do need to fix the `MAX_REQUESTS` check though.

src/main/java/uk/ac/cam/cl/dtg/util/locations/PostCodeIOLocationResolver.java
axlewin: Given that `POSTCODEIO_MAX_REQUESTS` is based on the max. number of postcodes that can be batch-requested from the `/postcodes` endpoint in one request, comparing the combined number of postcodes and outcodes to this limit feels fairly arbitrary - but there should definitely be a limit, and this works as well as any.

jacbn: Ah, okay. From a user perspective this keeps things consistent with how they were before which I prefer, though admittedly I struggle to see anyone using 100 different postcodes in one request anyway.

jacbn (approved):

PR #658, opened 2024-11-15
Only clean up ES indices when single indexer running

jsharkey13:
If more than one indexing process is happening, cleaning up unaliased indices is not a safe operation. There are several different ways the threads can conflict!
Much of the problem is caused by the fact each "version" is a collection of 5 indices {content, content_error, metadata, published_unit, unit} which each must get an alias. Each of these indices is created, and then aliased, separately.

1) Job A is in the process of creating a new version's collection of ES indices; none of them have aliases yet. Job B concurrently reaches the cleanup stage, finds the indices Job A is adding data to, sees they are unaliased and deletes them. This causes Job A to fail, but this failure should be noticed and fail safe(ish).

2) Job A has created all of the collection of indices for a new version, and is starting to move the various alias pointers. Job B reaches the cleanup stage, finds some mixture of two versions' indices without aliases and deletes them. This may cause Job A to fail, but will also leave the not-yet-aliased-as previous version in an invalid state; this version is being used by the API and the live site is broken.

This code now uses a thread-safe counter to skip running the deletion if another indexing job is running. However, this operation is not itself thread safe; it may take some time between the check and the response from ElasticSearch listing the indices and their aliases, in which time a new indexing job may have started creating indices. So we check again after getting the index list request that there are still no other concurrent jobs. This should catch all plausible cases.

Without truly serialising the whole ETL mechanics, I do not think we can make it perfect - but I believe this is a net improvement.

---

The first commit contains all the logic changes, later commits clean up some logging issues spotted during the work.


mwtrew (commented):
If I'm understanding this, we don't mind whether the indexing step happens concurrently across multiple threads, but we want to stop any deletion step happening while others are indexing. If so, I think we can (mis?)use [ReadWriteLock](https://www.baeldung.com/java-concurrent-locks#2-reentrantreadwritelock) for this and it will be a bit simpler. Read locks are not mutually exclusive, so can be used for the indexing step. Write locks can't be acquired while any read or write locks are held, so can be used for the deletion step. 

`tryLock()` for the write lock means it won't wait if the lock can't be acquired, it will just continue.

For example:
```java
    private static ReadWriteLock indexLock = new ReentrantReadWriteLock();

(...)

    void setNamedVersion(final String alias, final String version) throws Exception {
        log.info("Requested new aliased version: {} - {}", alias, version);

        try {
            indexLock.readLock().lock();
            indexer.loadAndIndexContent(version);
            log.info("Indexed version {}. Setting alias '{}'.", version, alias);
            indexer.setNamedVersion(alias, version);
            // Store the alias to file so that we can recover after wiping ElasticSearch.
            this.contentIndicesStore.saveProperty(alias, version);
        } finally {
            indexLock.readLock().unlock();
        }

        if (indexLock.writeLock().tryLock()) {
            indexer.deleteAllUnaliasedIndices();
            indexLock.writeLock().unlock();
        }
    }
```

mwtrew:
> I guess we also need to release the write lock in a try-finally too, right?

On reflection yes, no harm in doing so and we wouldn't want the write lock to be held forever if something went wrong!

mwtrew (approved):

PR #669, opened 2025-01-15
Use lightweight validation responses for progress stats

jsharkey13:
Prior to this commit, we loaded the full attempt JSON data from the database for every question attempt the user had made - but at no point do we need this data. We can avoid a great deal of deserialising if we instead load just the lightweight attempts, which we can do using existing queries (plus some plumbing).

One bug was revealed; although the database query sorts the lightweight attempts identically to the full attempts, and the structure of the nested Maps is identical - the Lightweight code path used HashMaps but the full validation response code path used LinkedHashMaps. The former do not preserve insertion order, and so were removing this sorting. The existing uses of the data clearly did not _need_ sorted attempts, but the My Progress stats page does. The cost of switching to the LinkedHashMap is small and only on creation; iterating over it as we do is much faster. This now preserves the (costly) sorting done in the DB.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
@@ -393,7 +393,7 @@ public class QuestionManager {
     }
 
     /**
-     * getQuestionAttemptsByUser. This method will return all of the question attempts for a given user as a map.
+     * getLightweightQuestionAttemptsByUser. This method will return all of the question attempts for a given user as a map.
      *
      * Attempts will not be augmented with the full attempt JSON data.
      *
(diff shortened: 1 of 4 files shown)


jsharkey13:
In my testing, this was a little over 50% faster for users with large numbers of question attempts (because _finding_ them in the database is still slow), but saves a (very) large amount of CPU work doing unnecessary deserialising!

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
```diff
@@ -392,6 +392,23 @@ public Map<String, Map<String, List<QuestionValidationResponse>>> getQuestionAtt
         }
     }
 
+    /**
+     * getQuestionAttemptsByUser. This method will return all of the question attempts for a given user as a map.
```
axlewin, line 396:
```suggestion
     * getLIghtweightQuestionAttemptsByUser. This method will return all of the question attempts for a given user as a map.
```

axlewin (approved):

PR #670, opened 2025-01-16
Add external validator timeout

jsharkey13:
Without a timeout specified here, the .send() call would block indefinitely. I believe 1 second is a safe enough limit; most validator requests take around 50-100ms (and trivial requests much less).
However, we should keep an eye on this when deployed.

This should prevent the validator from queuing too many requests and exhausting the Jetty threadpool.

For symbolic questions, the most common, explicitly log the timeout as an error by itself, to aid review of the timeout duration.


sjd210:
Tested this with adding sleep() to the symbolic validator locally and it works 👍 

1 second seems like a reasonable cutoff point. This should stop the Jetty threads being the bottleneck, so I'll be curious to see how the validators themselves handle an influx of requests if/when we get that again.

PR #678, opened 2025-02-21
Remove deprecated visibleToStudents flag on tests

jsharkey13:
The hiddenFromRoles is far more flexible than this check, and the redundancy between them was risking mistakes.


jsharkey13:
This PR now also removes some odd special-casing that the `TUTOR` role had. Tutors can now see everything not marked `"hiddenFromRoles": ["TUTOR" ...]`. Existing content has been updated to match the old behaviour.
(Tutors were previously prevented from previewing things hidden from students and things hidden from teachers).

sjd210:
Seems straightforward enough, and importantly it successfully makes practice tests actually practice-able once again 👍

PR #679, opened 2025-02-27
Question Finder improvements

jsharkey13:
This does not use the existing but deprecated GitContentManager method, but updates the questionSearch method to allow passing a random seed which will be used if no sort order is provided. The random sorting is done in ElasticSearch in exactly the same way the old gameboard builder randomisation was.

I reordered the arguments of the function to better logically group them, and removed some unused arguments from all the methods involved.

Add a hard limit to the question search endpoint, and also improve logging of the usage of the limit parameter.

This is entirely backwards compatible.


sjd210:
Nice. Tested this with and without sort order and it's all working as intended 👍 We'll have to be careful to ensure that the random seed is generated only once per page (or re-generated with a shuffle button) to avoid duplicates, but that's expected and fine.

PR #684, opened 2025-03-26
Improved gameboard loading

jsharkey13:
At present when gameboards are loaded in bulk and augmented by `getGameboardsByIds` in the `GameboardPersistenceManager`, we make separate requests to ElasticSearch for each gameboard. This is because the DO to DTO mapping is run once per board, rather than all together, and each call makes a query to ES.
But `getGameboardItemMap`, which that method implicitly used, had been written to allow bulk loading of data in much larger groups than a single gameboard. So instead of making the bulk `convertToGameboardDTOs` method use the single `convertToGameboardDTO` method, the reverse is now true; single boards are augmented via a singleton list and the bulk loading is the only code path.
That enables DO to DTO mapping to vastly reduce the number of ES queries that are made, and also reduced the spaghetti code down to a single code path. From there, all the augmentation methods can be combined and the duplicate or redundant code removed.

Doing this also allows us to fix a bug where existing bulk augmentation of gameboards would lose the creationContext for a question if it was included on multiple boards. This was because there was one map of question ID -> gameboard item (`gameboardReadyQuestions`), which could only keep just one context per question.
This rewrites the code to use a map of question ID to ContentDTO (which is always the same), and does the mapping inline in the renamed `augmentGameboardItemsWithContentData` method using the correct context for each question/gameboard combination.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
@@ -553,7 +553,7 @@ public class GameManager {
         List<GameboardDTO> sublistOfGameboards = resultToReturn.subList(startIndex, toIndex);
 
         // fully augment only those we are returning.
-        this.gameboardPersistenceManager.augmentGameboardItems(sublistOfGameboards);
+        this.gameboardPersistenceManager.augmentGameboardItemsWithContentData(sublistOfGameboards);
 
         return new GameboardListDTO(sublistOfGameboards, (long) resultToReturn.size(),
                 totalNotStarted, totalInProgress, totalAllAttempted);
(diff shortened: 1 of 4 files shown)


barna-isaac:
I wrote up our findings about user impact to the ticket: https://trello.com/c/Rbuzl2DR#comment-67e6b7a4ebb85ff98aa1e9b2
1) I understood that you've fixed the augmentation bug by leaving the original creationContext on the questions from postgres, and never augmenting these properties from elasticsearch
2) I understood that the performance improvement to `getGameboardsByIds` is due to the fact that the previous implementation performed augmentation for each gameboard, but we now augment items at once at the very end, in bulk
3) I understood that the reason these two changes happened as part of the same commit is that the `getGameboardItemMap` function is now removed.

I think both changes are great, so we should merge them as soon as possible.

PR #686, opened 2025-03-31
Add an endpoint to load a detailed quiz summary

jsharkey13:
Currently there is no way to see the rubric of a quiz (to work out if it is relevant to you, say) before freely attempting it. However, if you freely attempt it, you must complete it to remove it from your account.

This endpoint allows previewing the rubric for non-teacher users for all freely-attemptable quizzes.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
@@ -53,6 +53,7 @@ import uk.ac.cam.cl.dtg.isaac.dto.UserGroupDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ChoiceDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentBaseDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.content.ContentSummaryDTO;
+import uk.ac.cam.cl.dtg.isaac.dto.content.DetailedQuizSummaryDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.users.RegisteredUserDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.users.UserSummaryDTO;
 import uk.ac.cam.cl.dtg.segue.api.ErrorResponseWrapper;
@@ -295,7 +296,7 @@ public class QuizFacade extends AbstractIsaacFacade {
     }
 
     /**
-     * Preview a quiz. Only available to tutors and above.
+     * Preview the full contents of a quiz. Only available to tutors and above.
      *
      * @param request            - so we can deal with caching.
      * @param httpServletRequest - so that we can extract user information.
@@ -306,7 +307,8 @@ public class QuizFacade extends AbstractIsaacFacade {
     @Path("/{quizId}/preview")
     @Produces(MediaType.APPLICATION_JSON)
     @GZIP
-    @Operation(summary = "Preview an individual test.")
+    @Operation(summary = "Preview the full contents of a test.",
+            description = "Students may not preview tests.")
     public final Response previewQuiz(@Context final Request request,
                                       @Context final HttpServletRequest httpServletRequest,
                                       @PathParam("quizId") final String quizId) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/QuizFacade.java
```diff
@@ -348,6 +350,66 @@ public final Response previewQuiz(@Context final Request request,
...
+     * Students cannot preview tests, but they may wish to use the rubric to decide whether to attempt
+     * the quiz or not. The lightweight summaries provided by the overview endpoint
+     * {@link QuizFacade#getAvailableQuizzes(Request, HttpServletRequest)} do not contain the rubric because
+     * it can be quite large.
+     *
+     * @param httpServletRequest - so that we can extract user information.
+     * @param quizId             - the ID of the quiz the user wishes to attempt.
+     * @return a DetailedQuizSummaryDTO
+     */
+    @POST
```
barna-isaac, line 366:
```suggestion
    @GET
```

jsharkey13: 🤦 
Good spot, this is not a state-changing request. The perils of copy-pasting. Thanks!

barna-isaac:
Apart from the HTTP verb, looks all good, and it's great that the implementation for `getQuizSummary` in manager was so easy. Would have taken me a lot of time going through the code, and may still have missed that we already had something ready to use -- the PR shows the kind of effortless confidence that comes from knowing service very well. I'm also glad the PR is quite compact, which made it easier to review. I think this is good to merge.

Still, because this is one my first reviews, and @jsharkey13 has earlier asked explicitly that I document all impressions, here are a few ideas for improvement.
- we use two separate names for what is essentially the same thing. The endpoint says that it returns a `rubric` (path: `/quizId/rubric`, name: `viewQuizRubric`), whilst the type being returned is `DetailedQuizSummaryDTO`. I guess the question is: is it just an accident that `DetailedQuizSummaryDTO` currently consist of a rubric, or are these things fundamentally the same?
- logic about ensuring that a user has the rights to access the summary for a quiz could be pushed down into the manager. That way, we'd ensure the manager only returns information to users who are entitled, lowering the chance that we forget to perform the check somewhere else.
- does this endpoint not have tests because most of its behaviour is tested elswehere? I still need to understand how we make decisions about omitting or adding tests.

Finally, a general observation about the entire facade:
- it'd be nice if could reuse logic about caching and logging. Something like: `@CachedEndpoint(<how_to_obtain_key>, <expiration>)`, and `@LoggedEndpoint(<what_to_log>)` would eliminate a lot of boilerplate.

jsharkey13:
Responding to your points in order:

1) I used `DetailedQuizSummaryDTO` as the name for consistency with our other objects that extend a base object (`DetailedEventBooking`). But I realise that we renamed `DetailedUserSummaryDTO` to use a clearer name, although that was more because there were then 3 different summary objects and it was important for security to use the correct one.
   I think the facade should still mention rubrics because it makes it clear to the consumer of the API what the endpoint is for. But I am not against `QuizSummaryWithRubricDTO` for consistency with e.g. `UserSummaryWithGroupMembershipDTO` and `UserSummaryWithEmailAddressDTO`?

2) Our general rule is that permissions checks only occur at the facade level; because of the spaghetti interconnection of the managers, it reduces the proliferation of permissions checks to just one place. It's super easy to review every endpoint (I have done them all by hand in an afternoon); it is much harder to trace every code path and ensure that the permissions checks are done consistently. (This is often not this neat; there are both "who can use this endpoint" checks, but also "can this user see _this particular bit_ of this other user's data" checks. The latter do tend to be in managers).

3) For an endpoint I don't envision changing, which just loads a POJO and returns it to the client with minimal logic, it's not clear to me what utility a test has? If this endpoint was changed in future, the test would break and have to be fixed. But I cannot see (and we have never had) a case where a simple "load a static object" endpoint has broken accidentally and so the only time the test would fail would be when we have to replace it anyhow. If there was any business logic here, or augmentation of data, I'd want a test because those things _do_ break.
   Given that there already are IT tests for the QuizFacade, alongside all the content required in the test repo and ES data, it would be quick to add tests for this, though! What do you think would be useful?

4) I don't really see how any of these are less boilerplate than we have now? The caching stuff is very concisely wrapped up already; almost every endpoint has a different combination of how to generate an ETag (if at all), what Cache-Control header to add, whether the data is public, at what point you have enough data to check the ETag, etc. The verification boilerplate is also different every time; the messaging for each parameter is customised, and in a priority order (in some cases the order is critical for platform security). I don't see how you could reduce this with annotations that would need all this same data inside them.

jsharkey13:
Thanks for adding the tests we discussed, they look good to me!

PR #688, opened 2025-04-03
Send teacher welcome email on email verification

mwtrew:
---

**Pull Request Check List**
- [x] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/segue/api/EmailFacade.java
@@ -222,7 +222,7 @@ public class EmailFacade extends AbstractSegueFacade {
      * @param token  - A password reset token
      * @return Success if the token is valid, otherwise returns not found
      */
-    @GET
+    @POST
     @Path("/users/verifyemail/{userid}/{token}")
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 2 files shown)


barna-isaac (approved):
I've checked locally, using the university's SMTP servers, and the email works (I get the teacher welcome email when I verify the address). I've also checked and we won't start accidentally sending duplicate emails to teachers on Physics who are manually promoted before they verify their email address (the `ALLOW_DIRECT_TEACHER_SIGNUP_AND_FORCE_VERIFICATION` property is `false` there).

Two observations about the endpoint:
- even before this PR, `verifyEmail` has been a `GET` endpoint, that causes database changes (slightly bad). Now it's a `GET` endpoint that causes database changes and sends emails (slightly worse). I don't mind too much, and this PR didn't introduce this problem.
- the endpoint has a race condition, where it might double-post emails (which, in the case of the university's SMTP server caused no emails to go out at all -- at least I think this is what happened). I noticed this because React sends duplicate requests because of Strict Mode. Because both requests arrive at the same time, this piece of code
```java
 if (evStatus == EmailVerificationStatus.VERIFIED
                && user.getEmail().equals(user.getEmailToVerify())) {
            log.warn(String.format("Received a duplicate email verification request for (%s) - already verified",
                    user.getEmail()));
            return this.convertUserDOToUserDTO(user);
        }
``` 
clears both requests. This has been fine earlier, when both endpoints made the same changes to the database (so the endpoint was still idempotent), but now that the endpoint also sends emails, it might lead to duplication / emails getting dropped depending on the SMTP server's behaviour. For most users, requests won't be duplicated (React Strict Mode was the culprit in my case), but `GET` requests might be retried any time.

I still don't think we should worry about any of this, and would like to go ahead with the merge. @mwtrew please confirm that you also agree. @mwtrew please also let me know if you think anybody else should review this before I merge, or if there's anything you wanted feedback on but I overlooked.

jsharkey13:
Also, good spot Barna on the GET/POST thing. It's not the end of the world; it is idempotent as written, you can keep making the request and it won't do anything except the first time - but it does no harm and tightens up the CORS protection here too.

PR #689, opened 2025-04-04
Update dependencies

jsharkey13:
In the integration tests, postgres:12 EOL-ed last year, so that also warranted an update.


sjd210:
Everything in Docker and the API seem to be running as expected on my machine. I'm happy :+1:

PR #690, opened 2025-04-10
Add LLM Unit Tests

sjd210:
Adds a series of unit tests for the LLMFreeTextValidator using several typical question styles (default + advantage/disadvantage + point/explanation). 

Responses from the OpenAI client are mocked to be the desired response to the question, except in the case of explicitly testing invalid output. e.g. `"Foo Bar"` receives marks for `"pointFoo"` and `"pointBar"` but not `"pointFizz"`.

More tests for invalid behaviour are due to be added once failure reporting has been improved for the LLM-marked questions.

---

Also flags `markingFormulaString` and `markCalculationInstructions` as deprecated fields for LLM-marked questions. `markingFormulaString` was only ever needed in the content editor to store an editable version of the formula, so doesn't need to be a part of the content model, and `markCalculationInstructions` are outdated (and unused) now that formulae are used to calculate the mark total directly. These should be able to be removed safely, but I'll make sure it is and do so fully at a later point.


mwtrew (changes requested):
Looks good. I see quite a lot of static analysis warnings locally though - largely assertion arguments being the wrong way round which can be momentarily confusing when the tests fail, so worth fixing. 

Not too bothered about the long lines, but the other Checkstyle warnings should also be addressed.

sjd210:
Aha, the default setting for Checkstyle is to have it disabled for Test files, so I wasn't seeing any of the warnings 🙃 Shouldn't be too much effort to fix.

I also notice that I had double-spaced indentation for the JSON lists and such. I can probably save at least a few of the long line warnings reducing this, but I agree I don't really mind about the rest.

mwtrew (approved):

PR #691, opened 2025-04-23
Add null checking to LLM Free Text Questions

sjd210:
`IsaacLLMFreeTextQuestion`s have many unique fields not used in other questions that must each be set manually in the content editor. 

Most of these are fine to be null - several are explicitly optional (e.g. **markingFormula**, **additionalMarkingInstructions**) and so are already null-checked, others are lists (e.g. **markScheme**, **markedExamples**) so any nulls are safely converted to empty lists during processing - however **maxMarks** is a required field so was leading to a `NullPointerException` when missing. 

We can avoid this by simply setting a default value of 0 during question input validation.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacLLMFreeTextValidator.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacLLMFreeTextValidator.java
@@ -68,6 +68,10 @@ public class IsaacLLMFreeTextValidator implements IValidator {
         Objects.requireNonNull(question);
         if (!(question instanceof IsaacLLMFreeTextQuestion)) {
             throw new IllegalArgumentException(question.getId() + " is not a LLM free-text question");
+        } else {
+            if (((IsaacLLMFreeTextQuestion) question).getMaxMarks() == null) {
+                ((IsaacLLMFreeTextQuestion) question).setMaxMarks(0);
+            }
         }
 
         // Validate answer


src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacLLMFreeTextValidator.java
```diff
@@ -68,6 +68,10 @@ private void validateInputs(final Question question, final Choice answer) {
         Objects.requireNonNull(question);
         if (!(question instanceof IsaacLLMFreeTextQuestion)) {
             throw new IllegalArgumentException(question.getId() + " is not a LLM free-text question");
+        } else {
+            if (((IsaacLLMFreeTextQuestion) question).getMaxMarks() == null) {
+                ((IsaacLLMFreeTextQuestion) question).setMaxMarks(0);
```
jsharkey13, line 73: I don't think modifying the actual content document is the right way to deal with this. It makes it harder to reason about _why_ the property has the value it has, and it will pollute the DTO cache too.

I also don't think no marks is a good way to default this. If it's such an important property, then the validator should just return early with a "This question cannot be answered correctly" message, like [numeric questions do for invalid sig fig rules](https://github.com/isaacphysics/isaac-api/blob/37d5cd29a0a1cae8d33d57f3bf838586fc59ab58/src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacNumericValidator.java#L86-L94).

sjd210: I was actually looking into alternative responses like that numeric question one you show, but the different kind of response for this question type didn't easily allow it. Our default response much of the time currently *is* to give a zero mark result with no user-visible error (when the issue is with OpenAI's response) so this is a similar case to what we already do.

That said, I very much agree that's not very helpful; I'm already working on improving those other failure reports now in another branch so it's probably easiest if I bundle this in alongside those instead.

PR #693, opened 2025-05-07
refactor `IsaacLLMFreeTextValidator` tests

barna-isaac:
My goal was to highlight just the important parts of what each case tests. To achieve, I relied on:
- parameterising test cases to reduce redundancy
- a more concise notation for mark schemes
- hiding implementation details, such as examples (these are always present on a question but don't really matter for these cases as responses are mocked)
- showcasing important inputs, such as marking formulae

### DIFF
--- a/src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacLLMFreeTextValidatorTest.java
+++ b/src/test/java/uk/ac/cam/cl/dtg/isaac/quiz/IsaacLLMFreeTextValidatorTest.java
@@ -80,6 +80,11 @@ public class IsaacLLMFreeTextValidatorTest {
             expectMark(resp, expectedResult, expectedMarkTotal, toMarkScheme(breakdown));
         }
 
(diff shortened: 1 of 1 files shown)


sjd210 (commented):
This is really nice! Of course, we've talked through previous iterations so we already reached an agreement on what would work best, but it's great to see it come together 🥳 

My only remaining ask is that I think this could still do with some comments. I've marked specific places where I think it's important with some other comments. Afterwards, I think this should be good to go in.

barna-isaac:
@sjd210 have discussed what comments would be necessary in a private message. We agreed it's a good idea to include information about the three main kinds of mark scheme. I suggested also including a link to the Handbook, but in the end we concluded that it would be too distracting.

I've also expressed concerns about overusing JavaDocs, but we concluded that the helpers are worth annotating, especially because the IDE hints in IntelliJ might help future test authors. I tried to include helpful comments that still don't contain redundant information.

PR #694, opened 2025-05-09
Let anonymous users list student-visible quizzes

jsharkey13:
This does not affect viewing the quiz itself, so it just pushes the problem along a step. We may want to allow the "/view" endpoint to have similar behaviour too, for consistency, but I think that's a separate discussion.


jacbn:
All good besides test failure.

jacbn (approved):

PR #701, opened 2025-06-17
Improve question finder logging and search results

jsharkey13:
See the three individual commits for details:

 - don't show superseded/nofilter questions on random question searches;
 - don't log details about random question searches;
 - improve title and subtitle matching when searching questions.


jsharkey13:
The test is failing because the assignment test object better matches the "Regression Test Page" search than the page with that title. This is unfortunate, because generally the fuzzy match provides better results for title searches than the substring matching did 😣

mwtrew (approved):

PR #703, opened 2025-06-19
Add accordion section logged events to anonymous dataset

mwtrew:


jsharkey13 (approved):
Ideally everything API side would be on `master`, and not `redesign-2024`. But I guess we can fast-forward the other way, for once!

PR #710, opened 2025-07-10
Improvement/rest integration tests

barna-isaac:
I think this is now done: we have one integration test that's tested through the web interface rather than a direct method call. For wins, observe
- how the tests no longer need to care whether the `authenticationCallback` method passes the response to a parameter or if it returns the response
- that we no longer need to supply mocked versions of request, response and session.

These tests are rather complex too, because they test that a session was started on the server after the user logs in -- I think this is as difficult as it gets.

I was hoping to use a lighter implementation, like the one in commit [f140a63](https://github.com/isaacphysics/isaac-api/pull/710/commits/f140a6389ce458b58dce16e748b8c1fd7019443d), but it seems we need an http layer because some endpoints use the low-level `HttpServletRequest` class. As far as I understand, this is a realistic implementation. It uses an in-memory server, attaches a RestEasy dispatcher like the one we use in production, and registers the JAX-RS facade under test.

Regardless, this runs quite quickly (~20 ms per test case after the first, which takes a ~300ms warm-up), does not depend on the availability of any specific ports, and has the advantages outlined above.

@mwtrew has also asked whether this approach would avoid bugs like the one mentioned on [this thread](https://isaacphysics.slack.com/archives/C06582NR1B4/p1748964399934699). I'm hoping the answer is yes, because the endpoints now fully execute, and this would include serialisation.

I think we still need to decide whether to now re-write all the other integration tests to be using a web api too. I think we should aim for this, but should only actually rewrite tests as we work with them, and write any new tests so they test through the web api.

Although these now run an actual web api, there are still differences from live. For the live deployment, we attach additional logic in `IsaacApplicationRegister` and the `web-api-live.xml` files (eg: `UnhandledExceptionMapper`). Tests could be even more realistic if we also attached these. I left this improvement for the future.

The new `assertJ` dependency is tangential to the original task. It let me avoid casting as I was rewriting assertions. We can decide not to use it, I only introduced it because I thought the assertions looked nicer without the casts, and thought that introducing a test dependency was fine.


mwtrew (approved):
Looks good, I checked to make sure debugging still works and it does. I'm also wondering how easy to work with POST requests will be, but I don't think that's a blocker for this.

As-is, I think it will be hard for a newcomer to tell what parts of `IsaacIntegrationTest` are there to support the existing (with mocked requests etc.) way of doing things, and what's there to enable this new approach. A suggestion to make things more clear:

1. Move the common setup code (instantiating managers etc.) to a new `AbstractIsaacIntegrationTest`
2. Create a new subclass of that called `IsaacIntegrationTest` to hold the utilities for the existing tests
3. Create a new subclass of that called e.g. `IsaacIntegrationTestWithREST` for your new utilities

That means we won't have to rename the existing references to `IsaacIntegrationTest` (except in your new tests).

mwtrew:
Looks great - lastly there are some Checkstyle warnings locally that should be addressed, and it would be nice to have a short class-level Javadoc comment for `IsaacIntegrationTestWithREST` explaining what it does.

After that feel free to merge (I don't think it will need another review!)

PR #712, opened 2025-07-21
Augment concept pages with related content attempts

jsharkey13:
This adds attempt metadata to the related contents of concept pages. However, doing so means that the concept page data is no longer cacheable or public, and adds an extra database query to load this data, which may make increase query rates and affect performance.


sjd210 (approved):
Confirmed that this works through to the front-end properly. I'm happy  👍

PR #720, opened 2025-09-03
Replace Orika mapping with MapStruct

axlewin:
This removes the old Orika converters & automappers and replaces them with MapStruct mappings.

Most of the mappings just define a source & target type and let the MapStruct auto-implementations handle the rest. The DTO <-> DO mappings are named `map(source)` (or `copy(source)` if the source & target classes are the same); any other mappings are named according to the target type.

The mappers are split according to the kind of objects they map, but the `MainMapper` interface is used to provide a mapper object in most cases. For consistency, one of the new mappers is called `ContentMapper`, hence the existing `ContentMapper` has been renamed (to `ContentSubclassMapper`).

There are two new sets of tests: one for `Content` DO -> DTO mappings and one for `QuestionValidationResponse` DO -> DTO mappings. These are the only cases where there are lots of subclasses we can easily test at the same time, but we could have case-by-case tests for some/all of the other mappings too.

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dto/content/GlossaryTermDTO.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dto/content/GlossaryTermDTO.java
@@ -38,7 +38,7 @@ public class GlossaryTermDTO extends ContentDTO {
                            @JsonProperty("examBoard") String examBoard,
                            @Nullable @JsonProperty("stage") Set<String> stages) {
         this.explanation = explanation;
-        this.examBoard = examBoard != null ? examBoard : "";
+        this.examBoard = examBoard;
         this.stages = stages;
     }
 


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -1212,7 +1215,12 @@ private IsaacWildcard getWildCardById(final String id) throws ContentManagerExce
 
         Content wildcardResults = this.contentManager.getContentDOById(id);
 
-        return mapper.map(wildcardResults, IsaacWildcard.class);
+        if (wildcardResults instanceof IsaacWildcard) {
+            // Create a copy to avoid modifying the original object
+            IsaacWildcard wildcard = (IsaacWildcard) wildcardResults;
+            return mapper.copy(wildcard);
```
axlewin: It might be ok to just return `(IsaacWildcard) wildcardResults` here? I don't think the DO actually gets modified anywhere.

PR #724, opened 2025-09-25
Add pre-filtered table for state schools

mwtrew:
Adds a pre-filtered table of state schools matching our definition, to avoid repetition at the Superset dataset level.

### DIFF
--- a/src/main/resources/db_scripts/create_analysis_transforms.sql
+++ b/src/main/resources/db_scripts/create_analysis_transforms.sql
@@ -203,3 +203,13 @@ CREATE TABLE anonymous.teacher_engagement_v2_events AS
         )
     UNION
         (SELECT user_id, timestamp from anonymous.workshop_join_events);
+
+CREATE TABLE anonymous.england_scotland_state_schools_v2 AS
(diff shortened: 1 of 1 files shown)


src/main/resources/db_scripts/create_analysis_transforms.sql
```diff
@@ -202,4 +202,14 @@ CREATE TABLE anonymous.teacher_engagement_v2_events AS
             OR event_type = 'VIEW_HINT' AND event_details ->> 'hintIndex' = '0'
         )
     UNION
-        (SELECT user_id, timestamp from anonymous.workshop_join_events);
\ No newline at end of file
+        (SELECT user_id, timestamp from anonymous.workshop_join_events);
+
+CREATE TABLE anonymous.england_scotland_state_schools_v2 AS
```
jsharkey13, line 207: This is fine to create a new table, but it won't inherit the indices of its base table. Whilst we don't want the indices created unnecessarily during the initial initialisation of the anon database (before sending), we may want to add an index and primary key to this table elsewhere?

PR #729, opened 2025-10-06
Allow permissions & notes on fasttrack question pages

axlewin:
`IsaacFastTrackQuestionPage` extends `IsaacQuestionPage` without defining a new constructor, so Jackson uses the default constructor which takes no arguments and doesn't know about `permissions` or `notes`.  Therefore these fields need getters/setters in `SeguePage`.


sjd210 (approved):
Makes sense 👍

PR #772, opened 2026-04-02
Add "Exact answers only" option to coordinate questions

axlewin:
Adds an option for "Exact answers only" (`disregardSignificantFigures`) for coordinate questions. This works the same way as it does for numeric questions, i.e. validates that two values are numerically equal without considering significant figures, even if significant figures have been set on the question.

This enables coordinate questions to be used for cases where different values in the choice need to be given to different significant figures (e.g. (1, 2.34)).

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacCoordinateQuestion.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacCoordinateQuestion.java
@@ -29,6 +29,7 @@ public class IsaacCoordinateQuestion extends IsaacQuestionBase {
     private String[] suffixes;
     private String buttonText;
 
+    private Boolean disregardSignificantFigures;
     private Integer significantFiguresMin;
     private Integer significantFiguresMax;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacCoordinateQuestion.java
```diff
@@ -29,6 +29,7 @@ public class IsaacCoordinateQuestion extends IsaacQuestionBase {
     private String[] suffixes;
     private String buttonText;
 
+    private Boolean disregardSignificantFigures;
```
axlewin: For numeric questions we include this in the question DTO too, but I can't see that we need to?

jsharkey13 (approved):

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/IsaacCoordinateQuestion.java
jsharkey13: I agree, no need for DTO changes.

PR #777, opened 2026-04-21
Add bookmark functionality

axlewin:
Adds endpoints to allow users to bookmark/unbookmark question/concept pages and get a list of their bookmarked content. Also augments the `/questions` endpoint to include information about when a question was bookmarked, if at all.

Note that this doesn't augment the `/concepts` endpoint to include bookmarked status; this may be required depending on how we use this in the frontend.


src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java
```diff
@@ -0,0 +1,166 @@
...
+        RegisteredUserDTO user;
+        try {
+            user = userManager.getCurrentRegisteredUser(request);
+        } catch (final NoUserLoggedInException e) {
+            return SegueErrorResponse.getNotLoggedInResponse();
+        }
+
+        if (null != contentType && !(contentType.equals("isaacQuestionPage") || contentType.equals("isaacConceptPage"))) {
+            SegueErrorResponse error = new SegueErrorResponse(Status.BAD_REQUEST, "Invalid content type for bookmarks query: " + contentType);
+            log.warn(error.getErrorMessage());
```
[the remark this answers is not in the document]
jsharkey13: We actually should fix this, with a parameterised log message: do not use `error.getErrorMessage()` inside the log call.
I would keep the log.warn displaying the user provided thing, but make the HTTP error message just be "Only question and concept pages can be bookmarked!" without reference to the type they tried?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Adds a bookmark for the current user.
+     *
+     * @param request
+     *            - so we can find the current user.
+     * @param contentId
+     *           - the id of the content to bookmark.
+     */
+    @POST
+    @Path("/")
```
jsharkey13: Should this not be `/bookmark/{pageId}`? I think a path param is more sensible here than a query param. Then the delete endpoint can be the same resource path, but a DELETE instead.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java
```diff
@@ -0,0 +1,178 @@
...
+                SegueErrorResponse error = new SegueErrorResponse(Status.BAD_REQUEST,
+                        "Invalid content type for bookmark: " + contentType);
+                log.error(error.getErrorMessage());
+                return error.toResponse();
+            }
+            bookmarksDbManager.addBookmarkForUser(user.getId(), contentId, contentType);
+        } catch (final ContentManagerException | NullPointerException e) {
+            SegueErrorResponse error = new SegueErrorResponse(Status.NOT_FOUND,
+                    "Failed to create bookmark, could not find content with ID: " + contentId, e);
+            log.error(error.getErrorMessage(), e);
```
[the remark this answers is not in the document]
jsharkey13: Same here, we should be parameterising the log (and maybe not including the user-provided value back to them in the error message).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/BookmarksManager.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param bookmarks the list of bookmarks to map.
+     * @return the list of content summaries corresponding to the bookmarks.
+     */
+    public List<ContentSummaryDTO> mapBookmarkListToContentSummaryList(final List<BookmarkDO> bookmarks) {
+
+        List<ContentSummaryDTO> contentSummaries = new ArrayList<>();
+
+        for (BookmarkDO bookmark : bookmarks) {
+            try {
+                ContentDTO content = this.contentManager.getContentById(bookmark.contentId());
```
jsharkey13: This function call will go off to ElasticSearch potentially 100 times, which will be extraordinarily slow. It _might_ be able to use the GCM cache more effectively than a single request, but probably `GCM::getUnsafeCachedContentDTOsMatchingIds` outside the loop is a safer call here.
Maybe `GameboardPersistenceManager::augmentGameboardItemsWithContentData` is a good reference here (but if there are only ever 100 things, batching is likely unnecessary so that code is more complex than necessary here).

src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+@Path("/bookmarks")
+@Tag(name = "BookmarksFacade", description = "/bookmarks")
+public class BookmarksFacade {
+    private static final Logger log = LoggerFactory.getLogger(BookmarksFacade.class);
+
+    private final UserAccountManager userManager;
+    private final GitContentManager contentManager;
+    private final BookmarksManager bookmarksManager;
+    private final IBookmarks bookmarksDbManager;
```
jsharkey13: The facade layer shouldn't know anything about this layer; it should all go via the BookmarksManager!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java
```diff
@@ -0,0 +1,178 @@
...
+
+        if (null != contentType && !contentType.isEmpty()
+                && !(contentType.equals("isaacQuestionPage") || contentType.equals("isaacConceptPage"))) {
+            SegueErrorResponse error = new SegueErrorResponse(Status.BAD_REQUEST, "Invalid content type for bookmarks query: " + contentType);
+            log.warn(error.getErrorMessage());
+            return error.toResponse();
+        }
+
+        List<BookmarkDO> bookmarks = bookmarksDbManager.getBookmarksForUser(user.getId(), contentType);
+        return Response.ok(bookmarksManager.mapBookmarkListToContentSummaryList(bookmarks)).build();
```
jsharkey13: This logic should live in the BookmarksManager. Probably `BookmarksManager::getBookmarksForUser` could do both these steps in one?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java
```diff
@@ -0,0 +1,178 @@
...
+            return new SegueErrorResponse(Status.BAD_REQUEST, "Cannot delete bookmark without content ID.").toResponse();
+        }
+
+        List<BookmarkDO> currentBookmarks = bookmarksDbManager.getBookmarksForUser(user.getId());
+        if (currentBookmarks.stream().noneMatch(b -> b.contentId().equals(contentId))) {
+            return new SegueErrorResponse(Status.BAD_REQUEST, "You have not bookmarked this content.").toResponse();
+        }
+
+        bookmarksDbManager.removeBookmarkForUser(user.getId(), contentId);
+        return Response.ok().build();
```
jsharkey13: This could be a `204 No Content`.
```suggestion
        return Response.status(Status.NO_CONTENT).build();
```

I think the POST could be, too, to be honest, but I think we only use 204's for DELETEs at present.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/PgBookmarks.java
```diff
@@ -0,0 +1,117 @@
...
+
+        boolean filterByContentType = false;
+        if (null != contentType && !contentType.isEmpty()) {
+            if (contentType.equals("isaacQuestionPage") || contentType.equals("isaacConceptPage")) {
+                query += " AND content_type = ?";
+                filterByContentType = true;
+            } else {
+                // Should have already been caught at facade level
+                log.warn("Invalid content type provided for bookmarks query: {}", contentType);
+                throw new IllegalArgumentException("Invalid content type for bookmarks query: " + contentType);
```
jsharkey13: I don't think this filtering belongs here; if there is a string, just pass it transparently into Postgres and ask no questions of it. The validation belongs at the Facade layer, partly to give feedback quicker to the user, and partly because then this code is future-proof too.

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/PgBookmarks.java
```diff
@@ -0,0 +1,117 @@
...
+            pst.setLong(1, userId);
+            pst.setString(2, contentId);
+            pst.setString(3, contentType);
+            pst.setTimestamp(4, created);
+            if (pst.executeUpdate() == 0) {
+                throw new SegueDatabaseException("Unable to save bookmark.");
+            }
+
+        } catch (final SQLException | SegueDatabaseException e) {
+            e.printStackTrace();
```
jsharkey13: Please don't do `e.printStackTrace()`. You can do `log.error("Database error saving bookmark!", e);` if you have nothing better.

Won't this be hit on duplicate bookmarks? Or have I missed logic elsewhere to check for that?

jsharkey13: You could make this idempotent, if we don't care, by adding an `ON CONFLICT` clause that does nothing?

src/test/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacadeIT.java
```diff
@@ -0,0 +1,224 @@
...
+    public void tearDown() throws Exception {
+        // Restore ALICE_STUDENT bookmarks to original state
+        PreparedStatement pstDelete = postgresSqlDb.getDatabaseConnection().prepareStatement(
+                "DELETE FROM user_bookmarks WHERE user_id = ? AND content_id = ?");
+        pstDelete.setLong(1, ITConstants.ALICE_STUDENT_ID);
+        pstDelete.setString(2, ITConstants.FUZZY_MATCH_TEST_PAGE_ID);
+        pstDelete.executeUpdate();
+
+        PreparedStatement pstInsert = postgresSqlDb.getDatabaseConnection().prepareStatement(
+                "INSERT INTO user_bookmarks (user_id, content_id, content_type, created) VALUES (?, ?, ?, NOW()) ON CONFLICT DO NOTHING");
```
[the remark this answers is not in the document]
jsharkey13: Why is this #wontfix @axlewin?

jsharkey13 (changes requested):

src/main/java/uk/ac/cam/cl/dtg/isaac/dos/PgBookmarks.java
axlewin: Duplicates should be caught by the facade: https://github.com/isaacphysics/isaac-api/blob/feature/bookmarks/src/main/java/uk/ac/cam/cl/dtg/isaac/api/BookmarksFacade.java#L121-L123

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/BookmarksManager.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Augment a list of content summaries with bookmark information.
+     *
+     * @param bookmarks the bookmarks to augment the content summaries with.
+     * @param contentSummaries the content summary list to augment.
+     * @return the augmented content summary list.
+     */
+    public List<ContentSummaryDTO> augmentContentSummaryListWithBookmarkInformation(final List<BookmarkDO> bookmarks,
+                                                                                    final List<ContentSummaryDTO> contentSummaries) {
+        Map<String, Date> bookmarkMap = new HashMap<>();
```
jsharkey13: Could we short circuit here if bookmarks is empty?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -438,6 +440,8 @@ private void configureApplicationManagers() {
 
         bind(IUserStreaksManager.class).to(PgUserStreakManager.class);
 
+        bind(IBookmarks.class).to(PgBookmarks.class);
```
jsharkey13, line 443: Not that it functionally matters, but I think the intent was that data managers would live inside `configureDataPersistence` instead of `configureApplicationManagers`. I know that this isn't the case because there's a mess right now and at least 3/4 of these are data persistence managers. Feel free not to fix this now!

jsharkey13 (approved):

jsharkey13 (approved):

PR #783, opened 2026-04-27
Reject contact form messages over 5000 characters

axlewin:
This should be prevented by the frontend (see https://github.com/isaacphysics/isaac-react-app/pull/2120) when the contact form is used normally, but we should reject the request if this endpoint is accessed another way with a message that is too long.


jsharkey13 (approved):
Probably this limit isn't the same as the frontend one due to Unicode character vs codepoint fun (plus newline encoding questions), but it will do for now!

PR #787, opened 2026-05-05
Add date filter to assignment progress CSV endpoints

axlewin:
Adds an optional `to_date` URL parameter to the assignment progress CSV endpoints to allow question attempts to be filtered to only those occurring before a given date. The date must be provided as a Unix timestamp in ms, as is the case [elsewhere](https://github.com/isaacphysics/isaac-api/blob/main/src/main/java/uk/ac/cam/cl/dtg/segue/api/QuestionFacade.java#L237-L238).


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -499,7 +502,12 @@ public Response getAssignmentProgressDownloadCSV(@Context final HttpServletReque
                 questionPageIds.add(questionPage.getId());
             }
             Map<Long, Map<String, Map<String, List<LightweightQuestionValidationResponse>>>> questionAttempts;
-            questionAttempts = this.questionManager.getMatchingLightweightQuestionAttempts(groupMembers, questionPageIds);
+
+            if (null != toDate) {
+                questionAttempts = this.questionManager.getMatchingLightweightQuestionAttempts(groupMembers, questionPageIds, new Date(toDate));
+            } else {
+                questionAttempts = this.questionManager.getMatchingLightweightQuestionAttempts(groupMembers, questionPageIds);
+            }
```
jsharkey13, line 509: I don't mind this, but I feel like making a new `Date cutoffDate = null` (and then setting it if `toDate` is not null) which always gets passed into the function call makes the control flow easier to see? It is the same method in both cases (indeed the bottom one calls the top one with a `null`)

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
```diff
@@ -260,15 +261,21 @@ public Map<String, Map<String, List<QuestionValidationResponse>>> getQuestionAtt
...
         if (userIds.isEmpty()) {
             return Collections.emptyMap();
         }
 
-        String query = "SELECT id, user_id, question_id, correct, marks, timestamp FROM question_attempts"
-                     + " WHERE user_id = ANY(?) ORDER BY \"timestamp\" ASC";
+        String query;
+        if (null != toDate) {
+            query = "SELECT id, user_id, question_id, correct, marks, timestamp FROM question_attempts"
+                    + " WHERE user_id = ANY(?) AND timestamp < ? ORDER BY \"timestamp\" ASC";
```
jsharkey13, line 274: Should this be `<=`? I think so long as we document it clearly. `to_date` isn't as obviously a cutoff as `before` or `deadline` are, so it's not obvious to me what the right answer is!

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
```diff
@@ -307,16 +318,23 @@ public Map<String, Map<String, List<LightweightQuestionValidationResponse>>> get
...
                 = userIds.stream().collect(Collectors.toMap(Function.identity(), k -> Maps.newHashMap()));;
 
         try (Connection conn = database.getDatabaseConnection()) {
-            String query = "SELECT id, user_id, question_id, correct, marks, timestamp FROM question_attempts"
-                         + " WHERE user_id = ANY(?) AND page_id = ANY(?)"
-                         + " ORDER BY \"timestamp\" ASC";
+            String query;
+            if (null != toDate) {
+                query = "SELECT id, user_id, question_id, correct, marks, timestamp FROM question_attempts"
+                        + " WHERE user_id = ANY(?) AND page_id = ANY(?) AND timestamp < ?"
```
jsharkey13, line 331: Whatever we decide about `<` vs `<=` applies here too!

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
```diff
@@ -417,17 +417,35 @@ public Map<String, Map<String, List<QuestionValidationResponse>>> getQuestionAtt
...
+     *
+     * @param users who we are interested in.
+     * @param questionPageIds we want to look up.
+     * @return a map of user id to question page id to question_id to list of attempts.
+     * @throws SegueDatabaseException if there is a database error.
+     */
+    public Map<Long, Map<String, Map<String, List<LightweightQuestionValidationResponse>>>> getMatchingLightweightQuestionAttempts(
+            final List<RegisteredUserDTO> users, final List<String> questionPageIds)
+            throws SegueDatabaseException {
+        return getMatchingLightweightQuestionAttempts(users, questionPageIds, null);
```
jsharkey13, line 448: (If we swap the control flow that calls this to remove the need for it, we can remove this. But if we stick with it as-is, that's fine!)

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
axlewin: Hmm. The `to_date` is given to the nearest millisecond, and in practice I expect it will usually be exactly midnight on a given date (or at least an exact hour). In this case I think it does make sense to include attempts made at 23:59:59.999, but not at 00:00:00.000 the next day? The Javadocs on the endpoint methods imply that the cutoff is exclusive, but perhaps this should be more explicit.

(In practice it's unlikely to ever make a difference so I don't feel strongly either way!)

axlewin: If we were to surface this in the frontend with a date picker or similar we'd need to be clearer, but I don't think we're planning to do that?

src/main/java/uk/ac/cam/cl/dtg/segue/api/managers/QuestionManager.java
axlewin: I agree it's neater not to use this for the CSV endpoints, but we're using this in a few different places, e.g. [here](https://github.com/isaacphysics/isaac-api/blob/improvement/assignment-progress-date-cutoff/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java#L651), where `toDate` is irrelevant. Would you propose replacing all the usages with the (`users, questionPageIds, null`) overload?

jsharkey13: Ah, I see. No then, I think keeping this is the right thing to do!

jsharkey13 (approved):

src/main/java/uk/ac/cam/cl/dtg/isaac/quiz/PgQuestionAttempts.java
jsharkey13: No, I think we're not planning to surface this to users any time soon, and it is consistent across usages so let's keep it as-is.

PR #789, opened 2026-05-08
Augment gameboards with "saved to account" info on book pages & My Assignments

axlewin:


src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
```diff
@@ -184,7 +184,8 @@ public Response getAssignments(@Context final HttpServletRequest request) {
 
             // Gather all gameboards we need to augment for the assignments in a single query
             List<String> gameboardIds = assignments.stream().map(AssignmentDTO::getGameboardId).collect(Collectors.toList());
-            Map<String, GameboardDTO> gameboardsMap = this.gameManager.getGameboardsWithAttempts(gameboardIds, currentlyLoggedInUser)
```
axlewin, line 187: This was the only usage of `getGameboardsWithAttempts`, but I haven't removed the method because it seems like it could be useful again at some point.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -338,6 +338,62 @@ public final List<GameboardDTO> getGameboardsWithAttempts(final List<String> gam
...
+     */
+    public final List<GameboardDTO> getGameboardsWithUserSavedInformation(final List<String> gameboardIds, final RegisteredUserDTO user)
+            throws SegueDatabaseException {
+        if (null == gameboardIds || gameboardIds.isEmpty()) {
+            return new ArrayList<>();
+        }
+
+        List<GameboardDTO> gameboardsByIds = this.gameboardPersistenceManager.getGameboardsByIds(gameboardIds);
+        for (GameboardDTO gameboard : gameboardsByIds) {
+            augmentGameboardWithUserSavedInformation(gameboard, user);
```
jsharkey13: This loop, like the one below it, will make many database queries. For book pages, this might be limited to 4 or 5, but for My Assignments it could be hundreds for long-term users (and those in our year-long programmes or events with weekly assignments).

A new method that does this lookup in bulk has been on my mind for a while. 
```sql
SELECT gameboard_id FROM user_gameboards WHERE user_id = ? AND gameboard_id = ANY(?);
```
would work, although the performance of this will drop off rapidly once the list goes about maybe 50 elements.

Perhaps the right approach is then _two_ new public methods on the `GameboardPersistenceManager`, and to replace `isBoardLinkedToUser`. One takes a list of board IDs and returns a set of those the user is connected to, the other just returns all gameboard IDs connected to the user. We then use the second method whenever the first method is provided with too many gameboard IDs. This is the same approach `PgQuestionAttempts::getMatchingLightweightQuestionAttempts` takes.

```java
public Set<String> getGameboardIdsLinkedToUser(final Long userId, final Collection<String> gameboardIds)

public Set<String> getGameboardIdsLinkedToUser(final Long userId)
```
The only gotcha is documenting in the JavaDoc for the first one that the returned set may contain _more_ IDs than the filtered list, so set membership must still be checked.


Then `isBoardLinkedToUser` just becomes calling the first method with a singleton list and then checking whether the ID is in the returned set. A quick look at the query plans with `EXPLAIN ANALYSE` suggests that using the query above for a single board ID produces exactly the same query plan and duration as the `count(*)` approach it would replace.

jsharkey13: We use a mix of `Collection<String>` and `List<String>` for `gameboardIds` in the data manager at the moment, so I don't mind which you pick. Since we're returning a set, order can't matter, so flexibility seems best? But there may be something I have missed.

That `PgQuestionAttempts` method has a good overview of how to use SQL arrays for the `= ANY(?)` business. We do use `conn.createArrayOf(...)` in `getGameboardsByIds` in this file, but unlike the other places we don't call `.free()` afterwards and I'm not sure if that's bad but I think it might be.

jsharkey13: Then once these exist, you will need to use them. I think only one method should add user information; separate out the concerns of "augment with questions" and "augment with user link data". (You can both of these with a singleton list for the "get one gameboard fully augmented" case, and one or the other for the other cases).

I.e. I think the method `augmentGameboardWithQuestionAttemptInformationAndUserInformation` needs to go.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -338,6 +338,62 @@ public final List<GameboardDTO> getGameboardsWithAttempts(final List<String> gam
...
+            return new ArrayList<>();
+        }
+
+        List<GameboardDTO> gameboardsByIds = this.gameboardPersistenceManager.getGameboardsByIds(gameboardIds);
+        List<String> questionPageIds = gameboardsByIds.stream().map(GameboardDTO::getContents).flatMap(Collection::stream)
+                .map(GameboardItem::getId).collect(Collectors.toList());
+        Map<String, Map<String, List<LightweightQuestionValidationResponse>>> userQuestionAttempts =
+                questionManager.getMatchingLightweightQuestionAttempts(user, questionPageIds);
+        for (GameboardDTO gameboard : gameboardsByIds) {
+            augmentGameboardWithQuestionAttemptInformationAndUserInformation(gameboard, userQuestionAttempts, user);
```
jsharkey13: This for loop has the same issue.

jsharkey13 (changes requested):
I am sure this would work fine, but I'd like to both make it more efficient, and try and simplify the sheer number of different methods here.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/GameManager.java
```diff
@@ -388,7 +388,15 @@ public final List<GameboardDTO> getGameboardsWithAttemptsAndUserSavedInformation
         Map<String, Map<String, List<LightweightQuestionValidationResponse>>> userQuestionAttempts =
                 questionManager.getMatchingLightweightQuestionAttempts(user, questionPageIds);
         for (GameboardDTO gameboard : gameboardsByIds) {
-            augmentGameboardWithQuestionAttemptInformationAndUserInformation(gameboard, userQuestionAttempts, user);
+            augmentGameboardWithQuestionAttemptInformation(gameboard, userQuestionAttempts);
+        }
+
+        Set<String> savedBoardIds = this.gameboardPersistenceManager.getGameboardIdsLinkedToUser(user.getId(), gameboardIds);
+        // getGameboardIdsLinkedToUser may return IDs that weren't in the original list, so take the intersection:
+        savedBoardIds.retainAll(gameboardIds);
```
jsharkey13: 🤔 
Don't _think_ this is necessary 😅

src/main/java/uk/ac/cam/cl/dtg/isaac/api/AssignmentFacade.java
jsharkey13: We have now removed the method, because we think that any time you want a user's question attempts, you probably want whether they are connected too (and the overhead of that is minimal compared to loading question attempts).

jsharkey13 (approved):

PR #790, opened 2026-05-18
Add warning for URL invalid media sources

machacekj0:
Looking at filenames which Jetty rejected as illegal, for example `kinetics_initial_rate_with_[b].svg`. This adds a content error for the first invalid character in any bad filenames.


sjd210 (approved):
Looks good, and works well when tested locally 👍 :)

PR #791, opened 2026-05-19
Stop LLMFreeTextQuestions from generating choices-based content warnings

sjd210:
`isaacLLMFreeTextQuestion`s do not require a `choices` field, since they use a `markScheme` as its equivalent. 
We should therefore not be giving a warning when the `choices` field is missing.


barna-isaac (approved):
"Files with errors" is down from 445 to 350 on my local machine with these changes.
Errors for the file `sqa_2025_q5` also nicely demonstrate that the changes worked well. It previously had 3 warnings about missing items, and now just has the single warning about the markScheme missing (not a published question, don't worry.)

Approved.

PR #793, opened 2026-05-21
Add endpoint to get sitemap from content

axlewin:


src/main/java/uk/ac/cam/cl/dtg/segue/api/InfoFacade.java
```diff
@@ -122,6 +129,44 @@ public final Response getSegueEnvironment(@Context final Request request) {
...
+
+            if (null == fileContent) {
+                String refererHeader = httpServletRequest.getHeader("Referer");
+                log.warn("Unable to locate sitemap file. Referer: ({})", refererHeader);
+                return new SegueErrorResponse(Response.Status.NOT_FOUND, "Unable to locate sitemap file.").toResponse();
+            }
+
+            return Response.ok(fileContent.toByteArray()).type(APPLICATION_XML)
+                    .cacheControl(getCacheControl(NUMBER_SECONDS_IN_ONE_DAY, true))
+                    .header("Content-Disposition", "attachment")  // Do not show this file in the browser.
```
jsharkey13:
```suggestion
```
I think we don't need this for a `sitemap.xml` file. We only wanted it for the documents so that they got downloaded by users rather than displayed.

src/main/java/uk/ac/cam/cl/dtg/segue/api/InfoFacade.java
```diff
@@ -122,6 +129,44 @@ public final Response getSegueEnvironment(@Context final Request request) {
...
+    /**
+     * Gets the sitemap from the content database.
+     *
+     * @param httpServletRequest
+     *            - used for the Referer header for helpful error messages.
+     * @return a Response containing the sitemap file contents as Content-Disposition: attachment, or a SegueErrorResponse.
+     */
+    @GET
+    @Produces(APPLICATION_XML)
+    @Path("/sitemap")
```
jsharkey13: Given this only serves one file, and it must be XML if found, maybe this can be
```suggestion
    @Path("/sitemap.xml")
```
You would have to change your Nginx config if we change this, though. The 404 will return JSON at present, but maybe that's fine even with the `@Produces` as-is.

We could always
```java
    return Response.status(Response.Status.NOT_FOUND).type(APPLICATION_XML).build();
```
to return a completely empty response if we wanted to always return XML. Since we have no useful error message beyond the 404 itself, that doesn't seem terrible.

jsharkey13: The other errors would remain JSON, though, and I am not sure fixing those is worth the effort.

src/main/java/uk/ac/cam/cl/dtg/segue/api/InfoFacade.java
```diff
@@ -122,6 +129,44 @@ public final Response getSegueEnvironment(@Context final Request request) {
...
+    @Operation(summary = "Get the sitemap file from the content database.")
+    public final Response getSitemap(@Context final HttpServletRequest httpServletRequest) {
+        String path = "sitemap.xml";
+        try {
+            ByteArrayOutputStream fileContent = this.contentManager.getFileBytes(path);
+
+            if (null == fileContent) {
+                String refererHeader = httpServletRequest.getHeader("Referer");
+                log.warn("Unable to locate sitemap file. Referer: ({})", refererHeader);
```
jsharkey13: Do we think that the `Referer` header on a request to the sitemap might contain anything meaningful? It is helpful for embedded images, but I can't imagine sitemap requests are anything but direct accesses?
If not helpful, we can ditch the `httpServletRequest` from the method params, and just log a warning that a sitemap was requested but none was found.

jsharkey13 (approved):

PR #794, opened 2026-06-01
Remove deprecated Elasticsearch query framework

axlewin:
Removes all usage of `BooleanSearchClause` and `AbstractFilterInstruction`-based searching in favour of the newer `AbstractInstruction`-based pattern. This removes a lot of duplicated logic; there should now be a single,  consistent approach to building ES queries across the whole project.

Also removes the `generateTemporaryGameboard` endpoint, which has been unsupported for over a year. This change is technically unrelated but was needed in order to remove some of the ES framework.


src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -252,12 +248,10 @@ public Content getContentDOById(final String id, final boolean failQuietly) thro
         try {
             ResultsWrapper<Content> result = contentDOcache.get(k, () -> {
 
-                ResultsWrapper<String> rawResults = searchProvider.termSearch(
-                        contentIndex,
-                        CONTENT_INDEX_TYPE.CONTENT.toString(),
-                        id,
-                        Constants.ID_FIELDNAME + "." + Constants.UNPROCESSED_SEARCH_FIELD_SUFFIX, 0, 1,
-                        getBaseFilters());
```
jsharkey13, line 260: Have we not lost the base filters here, now?

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -300,24 +294,16 @@ public ResultsWrapper<ContentDTO> getUnsafeCachedContentDTOsMatchingIds(final Co
...
 
-                Map<String, AbstractFilterInstruction> finalFilter = Maps.newHashMap();
-                finalFilter.putAll(new ImmutableMap.Builder<String, AbstractFilterInstruction>()
-                        .put(ID_FIELDNAME + "." + UNPROCESSED_SEARCH_FIELD_SUFFIX,
-                                new TermsFilterInstruction(ids))
-                        .build());
+                BooleanInstruction searchInstruction = new BooleanInstruction();
 
-                if (getBaseFilters() != null) {
-                    finalFilter.putAll(getBaseFilters());
```
jsharkey13, line 310: And here too?

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -723,19 +622,21 @@ public ContentDTO populateRelatedContent(final ContentDTO contentDTO)
...
         List<String> relatedContentIds = Lists.newArrayList();
         for (ContentSummaryDTO summary : contentDTO.getRelatedContent()) {
             relatedContentIds.add(summary.getId());
         }
 
-        fieldsToMap.add(new BooleanSearchClause(
-                Constants.ID_FIELDNAME + '.' + Constants.UNPROCESSED_SEARCH_FIELD_SUFFIX,
-                Constants.BooleanOperator.OR, relatedContentIds));
+        // build query the db to get full content information
+        BooleanInstruction searchInstruction = new BooleanInstruction();
```
jsharkey13: Is this not recreating the logic of `getUnsafeCachedContentDTOsMatchingIds`? And does this new implementation not have the same bug that the default filters that `findByFieldNames` used to apply are now being skipped?

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
```diff
@@ -811,7 +712,7 @@ public String getCurrentContentSHA() {
      */
     public IsaacSearchInstructionBuilder getBaseSearchInstructionBuilder() {
         return new IsaacSearchInstructionBuilder(
-                searchProvider, this.showOnlyPublishedContent, this.hideRegressionTestContent, true);
+                searchProvider, this.showOnlyPublishedContent, this.hideRegressionTestContent, false);
```
jsharkey13: I don't think that changing the default here is the right approach. I think the problem is the lack of setter on the builder for this parameter? I think the right thing to do is to add that setter, revert this, and then work out which places you needed it changed for and make them use that.

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/DeleteEventAdditionalBookingInformationJob.java
```diff
@@ -57,20 +52,19 @@ public void execute(final JobExecutionContext context) throws JobExecutionExcept
...
         Map<String, Constants.SortOrder> sortInstructions = Maps.newHashMap();
-        Map<String, AbstractFilterInstruction> filterInstructions = Maps.newHashMap();
-        fieldsToMatch.put(TYPE_FIELDNAME, Collections.singletonList(EVENT_TYPE));
         sortInstructions.put(DATE_FIELDNAME, SortOrder.DESC);
-        DateRangeFilterInstruction anyEventsToNow = new DateRangeFilterInstruction(null, new Date());
-        filterInstructions.put(ENDDATE_FIELDNAME, anyEventsToNow);
+
+        BooleanInstruction instruction = new BooleanInstruction();
+        instruction.must(new MatchInstruction(TYPE_FIELDNAME, EVENT_TYPE));
+        instruction.must(new RangeInstruction<Long>(ENDDATE_FIELDNAME).lessThanOrEqual(new Date().getTime()));
```
jsharkey13: Is there a bug here? Was there a bug here before too?

Below, there's a comment I think I wrote about "end date (if present) > 30 days ago, else ...", but will this not _only_ match events with end dates set, and thus _never_ load events where we need to do the other logic?
I have a feeling events are supposed to have `end_dates` (my 10 year old snake case typo haunts me) these days, but we might miss removing PII if they don't?

jsharkey13 (commented):
A few questions about subtle filtering behaviour where I think we were, deep under the hood, doing some filtering which this no longer does - because it does not use the new-style builder that does that by default.

src/main/java/uk/ac/cam/cl/dtg/segue/scheduler/jobs/DeleteEventAdditionalBookingInformationJob.java
axlewin: ...yes, I think you're right, and this bug has been around for as long as this job has existed, possible copy & pasted from [here](https://github.com/isaacphysics/isaac-api/blob/6e05f2ae05ec19d0dead1ab0b725c9a869d8864f/src/main/java/uk/ac/cam/cl/dtg/isaac/api/EventsFacade.java#L178-L179) 😔

That said, creating new events without end dates now causes a content error and I [bulk-added end dates](https://github.com/isaacphysics/rutherford-content/pull/547) for older events a while ago so I think this is actually ok now, but we could have a `DATE_FIELDNAME` backup here to be safe. At the very least the comment ought to be consistent with the actual behaviour.

src/main/java/uk/ac/cam/cl/dtg/segue/dao/content/GitContentManager.java
axlewin: You're correct that the reason I did this was to avoid adding a setter. However, it turns out there _is_ a setter, it's just not called `setExcludeNofilterContent` so I didn't see it before.

I'm tempted to rename the setter, but it does negate its input so maybe not (I suppose it's also not _strictly_ a setter in the JavaBeans sense if we're returning `this`...).

Anyway, yes, I'll revert this and use the setter where necessary.

jsharkey13 (approved):

PR #798, opened 2026-06-10
Skills question endpoint

barna-isaac:
This adds a new endpoint, meant for Anvil apps to record new question attempts.

For now, the new endpoint:
- is not rate limited
- does not create entries in the `logged_events` table
I plan to implement this functionality later, when the feature is more mature.

## Testing
I've manually tested that attempts are recorded on my own machine, by visiting http://localhost:8004/pages/app_page_mental_maths_overall, opening the JS console, and intercepting one of the messages emitted by the app:
```js
window.addEventListener('message', e => {
    if(e.origin === 'https://cuta66namfmobkjs.anvil.app' && e.data.type === 'SUBMISSION_MARKED') {
        console.log('got anvil message', e.data);
    }
});
``` 

Then, a request can be posted with the following curl:
```
curl --location 'http://localhost:8080/isaac-api/api/skills/app_page_mental_maths_overall_recorded/answer' \
--header ' __Host-SEGUE_AUTH_COOKIE=<your_cookie_here>\
--header 'Content-Type: application/json' \
--data '{
    "payload": "<from_message>",
    "hmac": "<from_message>"
}'
```

## Overview of changes
To help with the review, here is a high-level list of changes:
17 changed files in total.
- 11 changes to the implementation
  - 2 new DTO's 1 new Exception. `InvalidAnvilMarkingRequestException.java`, `AnvilMarkingRequestDTO.java`, `AnvilPayloadDTO.java`. Two DTO's because the HMAC'ed payload is first validated, and only then parsed.
  - For the implementation, 4 files: `SkillsFacade.java`, `SkillsManager.java`, `ISkillsAttemptManager.java`, `PgSkillsAttemptManager.java`, mirroring the Manager <-> IPersistenceManager <-> PersistenceManager pattern for question attempts
  - 4 configuration files, `Constants.java`, `IsaacApplicationRegister.java`, `SegueGuiceConfigurationModule.java`, and a migration in `2026-06-skills_question_attempts.sql`

- 6 changes to testing
  - 3 files for testing the new endpoint: `SkillsFacadeIT.java`, `postgres-rutherford-create-script.sql` to apply the new db schema for testing, and `segue-integration-test-config.yaml` so the new config value is set for test.
  - 4 files for changes to the test framework. `ElasticSearchTestHelper.java`, `QuestionFacadeIT.java`, `IsaacIntegrationTestWithRest.java` because I've extracted a helper method from QuestionFacadeIT, so the same method can be reused in the new tests. The method lets us add content to elastic search. Further, the test server now applies the same exception handlers as the live app.
 
---

**Pull Request Check List**
- [ ] ~Security - New dependency - configured sensibly not relying on defaults~
- [ ] ~Security - New dependency - Searched for any know vulnerabilities~
- [ ] ~Security - New dependency - Signed up team to mailing list~
- [ ] ~Security - New dependency - Added to dependency list~
- [ ] Peer-Reviewed


src/test/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacadeIT.java
```diff
@@ -0,0 +1,200 @@
...
+                    "Payload user_id does not match session"),
+                Arguments.of("expired timestamp", validPayload(p -> p.put("timestamp", "2022-01-01T13:00:00Z")),
+                    "Payload timestamp is outside the allowed window")
+            );
+        }
+
+        @ParameterizedTest(name = "{0}")
+        @MethodSource("invalidPayloads")
+        public void invalidPayload_Returns400(
+            final String name, final JSONObject payload, final String expectedMessage
```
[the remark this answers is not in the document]
barna-isaac: It's used in the annotation

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsManager.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param payloadStr - the payload string from the marking request
+     * @param userId     - the ID of the currently authenticated user
+     * @param appId      - the app ID from the URL, which must match the payload's skill_id
+     * @throws InvalidAnvilMarkingRequestException if the payload is malformed or any validation fails
+     */
+    public AnvilPayloadDTO parsePayload(
+        final String payloadStr, final long userId, final String appId
+    ) throws InvalidAnvilMarkingRequestException {
+        try {
+            if (payloadStr.length() > 10 * 1024) {
```
jacbn: This ought to be a constant, if for no other reason than you can explain in a comment where this comes from more easily there!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsManager.java
```diff
@@ -0,0 +1,89 @@
...
+                throw new InvalidAnvilMarkingRequestException("Payload too large", null);
+            }
+
+            AnvilPayloadDTO dto = objectMapper.readValue(payloadStr, AnvilPayloadDTO.class);
+            if (dto.getUserId() != userId) {
+                throw new InvalidAnvilMarkingRequestException("Payload user_id does not match session", null);
+            }
+            if (dto.getTimestamp().before(new Date(System.currentTimeMillis() - 300_000L))) {
```
jacbn: Another magic number, ought to be a constant

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
```diff
@@ -0,0 +1,119 @@
...
+            var error = new SegueErrorResponse(Status.NOT_FOUND, "Error locating the version requested", e);
+            log.error(error.getErrorMessage(), e);
+            return error.toResponse();
+        }
+    }
+
+    private boolean hasAnvilApp(final Content content) {
+        return content instanceof AnvilApp
+            || (content != null
+                && content.getChildren().stream().anyMatch(c -> c instanceof Content cc && hasAnvilApp(cc)));
```
jacbn: Why is this recursive and not simply `isAnvilApp`? Beyond having the potential to be incredibly slow, we later check that the ID of the content object originally passed in to this method matches the payload's `dto.getSkillId()`, which surely implies said original object is expected to be an app in the first place? Why make this pass if it _contains_ an anvil app, when later on such an object would cause an exception?

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/AnvilPayloadDTO.java
```diff
@@ -0,0 +1,109 @@
...
+            @JsonSetter(nulls = Nulls.FAIL) @JsonProperty(value = "skill_id", required = true) final String skillId,
+            @JsonSetter(nulls = Nulls.FAIL) @JsonProperty(value = "subskill_id", required = true)
+            final String subskillId,
+            @JsonSetter(nulls = Nulls.FAIL) @JsonProperty(value = "question", required = true) final JsonNode question,
+            @JsonSetter(nulls = Nulls.FAIL) @JsonProperty(value = "question_attempt", required = true)
+            final JsonNode questionAttempt,
+            @JsonSetter(nulls = Nulls.FAIL) @JsonProperty(value = "marks", required = true) final Number marks,
+            @JsonSetter(nulls = Nulls.FAIL) @JsonProperty(value = "timestamp", required = true) final Date timestamp) {
+        Validate.isTrue(userId != 0);
+        Validate.isTrue(skillAssignmentId == null);
```
jacbn, line 58: Out of interest, what happens if these fail? The [docs](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/Validate.html#isTrue(boolean)) suggest this throws an `IllegalArgumentException`, which I can't find being caught anywhere. I understand this will only happen if an anvil app is broken and is our problem and we could fix it, but having an uncaught exception around these things feels messy? What do you think?

jacbn (commented):
Logically, this looks okay to me. I'm not really in a position where I can comment on the testing side of things, so perhaps someone else could have a look over those parts?

I'm not keen on, particularly in the maths app, how 20 separate HTTP requests are fired for each answer. Each has its own overhead r.e. user authentication etc, so there is a lot of repeated work done. However, I understand that with time constrains, combining them all into a single request might not be feasible. Have you thought about this? How much work would it be?

That aside, I have spotted a few things that I'm not sure why you've done or that could use e.g. transforming into a constant. On the whole though the logic here looks good!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsManager.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Verifies that the HMAC in the DTO matches the expected signature of the payload.
+     *
+     * @param dto - the parsed marking request
+     * @return whether the hmac was valid
+     */
+    public boolean isHmacValid(final AnvilMarkingRequestDTO dto) {
+        String expected = UserAuthenticationManager.calculateHMAC(hmacSecret, dto.getPayload());
+        return expected.equals(dto.getHmac());
```
jsharkey13: This is vulnerable to timing attacks, which maybe we don't care about here since we're doing this more to discourage trivial cheating and there are far simpler ways to cheat than forging attempts via timing attacks!

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
```diff
@@ -0,0 +1,95 @@
...
+    @Path("/{appId}/answer")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response answerQuestion(@Context final HttpServletRequest request,
+                                   @PathParam("appId") final String appId,
+                                   final AnvilMarkingResponseDTO markingResponse) {
+        try {
+            RegisteredUserDTO currentUser = userManager.getCurrentRegisteredUser(request);
+            if (!(this.contentManager.getContentDOById(appId) instanceof AnvilApp)) {
+                var error = new SegueErrorResponse(Status.NOT_FOUND, "No app found for given id: " + appId);
+                log.warn(error.getErrorMessage());
```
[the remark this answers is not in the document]
jsharkey13: Whilst I normally dismiss these alerts, in this case this _is_ more dangerous because the concatenation happened inside the error constructor line, and the log message is thus not parameterised.
```suggestion
                log.warn("No app found for given id: {}", appId);
```

I would also make the error returned to the user not include a user-provided value, either. Just make it say "No app with this ID found!".

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
```diff
@@ -0,0 +1,119 @@
...
+
+            return Response.ok(payloadDTO).build();
+        } catch (final NoUserLoggedInException e) {
+            return SegueErrorResponse.getNotLoggedInResponse();
+        } catch (final InvalidAnvilMarkingRequestException e) {
+            var error = new SegueErrorResponse(Status.BAD_REQUEST, e.getMessage());
+            log.warn(error.getErrorMessage() + ", " + e.getDetailedProblem());
+            return error.toResponse();
+        } catch (final SegueDatabaseException e) {
+            if ("Duplicate attempt ID".equals(e.getMessage())) {
```
jsharkey13: Should this not be a specific exception subtype? This seems very fragile, even if you moved it to use the same string constant in both places.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
```diff
@@ -0,0 +1,119 @@
...
+                var error = new SegueErrorResponse(Status.CONFLICT, "Duplicate attempt ID");
+                log.warn(error.getErrorMessage());
+                return error.toResponse();
+            }
+            var error = new SegueErrorResponse(Status.INTERNAL_SERVER_ERROR, "Something went wrong");
```
jsharkey13: Here and other places are missing punctuation, we should be consistent and end with a full stop or exclamation mark.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
jsharkey13: Anvil apps do not exist in the content with their app IDs being their content object IDs. So we cannot be loading the content by app ID. But this comment does highlight an issue; what IDs are we using in which places?

I.e. an Anvil app in the content looks like:
```json
{
  "type": "anvilApp",
  "appId": "AAABBBCCCDDDEEEFFF",
  "appAccessKey": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
  "author": "jsharkey13"
}
```
it does not have a content object ID and cannot be loaded by ID.

If we use the app ID as the attempt identifier, this has the advantage that the same app on two different pages share attempts, but then we have no way of loading it with our existing ElasticSearch methods (though we could create ones). 
If we use some other identifier, like the page ID, that could contain multiple Anvil apps and we will muddle attempts together.
We could add IDs to Anvil app content objects, but then these IDs will be scoped by the ETL and API process; they will become `page_id|accordion_id|app_content_object_id` which will restrict apps to pages and won't match what the Anvil app itself thinks its ID is?

src/main/java/uk/ac/cam/cl/dtg/isaac/configuration/IsaacApplicationRegister.java
```diff
@@ -123,6 +124,9 @@ public final Set<Object> getSingletons() {
             this.singletons.add(injector.getInstance(EmailFacade.class));
             this.singletons.add(injector.getInstance(QuizFacade.class));
             this.singletons.add(injector.getInstance(BookmarksFacade.class));
+            if (injector.getInstance(AbstractConfigLoader.class).getProperty(SKILLS_HMAC_SECRET) != null) {
```
jsharkey13: This is not nice, and I'm not actually sure it will prevent the facade from being registered, I fear it may just prevent it from being a pre-initialised singleton.

It would be better for the facade just to return `501 Not Implemented`s if no secret is configured?

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgSkillsAttemptManager.java
```diff
@@ -0,0 +1,46 @@
...
+            pst.setString(3, attempt.getSkillAssignmentId());
+            pst.setString(4, attempt.getSkillId());
+            pst.setString(5, attempt.getSubskillId());
+            pst.setString(6, attempt.getQuestion().toString());
+            pst.setString(7, attempt.getQuestionAttempt().toString());
+            pst.setInt(8, (Integer) attempt.getMarks());
+            pst.setTimestamp(9, new Timestamp(attempt.getTimestamp().getTime()));
+            pst.executeUpdate();
+        } catch (final SQLException e) {
+            if ("23505".equals(e.getSQLState())) {
```
jsharkey13: This feels horrible. I suppose this is a Postgres-specific class. Maybe using `PSQLState.UNIQUE_VIOLATION.getState()` would be at least a little easier to understand, and force an import of a Postgres-specific driver, which means this can't silently fail one day?

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgSkillsAttemptManager.java
```diff
@@ -0,0 +1,46 @@
...
+import uk.ac.cam.cl.dtg.isaac.dto.AnvilPayloadDTO;
+import uk.ac.cam.cl.dtg.isaac.quiz.ISkillsAttemptManager;
+import uk.ac.cam.cl.dtg.segue.dao.SegueDatabaseException;
+import uk.ac.cam.cl.dtg.segue.database.PostgresSqlDb;
+
+import java.sql.SQLException;
+import java.sql.Timestamp;
+
+/** PostgreSQL-backed persistence for Anvil skills question attempts. */
+public class PgSkillsAttemptManager implements ISkillsAttemptManager {
```
jsharkey13: I would have expected `ISkillsAttemptManager` to be the `SkillsManager`, not the persistence layer manager? I think we use `Data` in these names for the persistence ones?

src/main/java/uk/ac/cam/cl/dtg/segue/configuration/SegueGuiceConfigurationModule.java
```diff
@@ -874,6 +877,25 @@ private IQuestionAttemptManager getQuestionManager(final PostgresSqlDb ds, final
...
+     * SkillAttemptManager.
+     *
+     * @param ds - postgres data source
+     * @return a singleton for question persistence.
+     */
+    @Inject
+    @Provides
+    @Singleton
+    private ISkillsAttemptManager getSkillAttemptManager(final PostgresSqlDb ds) {
+        // this needs to be a singleton as it provides a temporary cache for anonymous question attempts.
```
jsharkey13: The actual `PgSkillsAttemptManager` is entirely stateless and very simple. It does not need to be a singleton at all, but is simple enough that Guice can make it one with `@Singleton` annotated on the class itself, but I don't even think that is necessary. I would get rid of this, and leave Guice to make it as needed.

src/main/resources/db_scripts/postgres-rutherford-create-script.sql
```diff
@@ -1542,7 +1542,24 @@ ALTER TABLE ONLY public.user_streak_freezes
 ALTER TABLE ONLY public.user_streak_targets
     ADD CONSTRAINT user_streak_targets_user_id_fkey FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE;
 
+--
+-- Name: skill_question_attempts; Type: TABLE; Schema: public; Owner: rutherford
+--
+CREATE TABLE public.skills_question_attempts (
```
jsharkey13: This will work, but it isn't ideal. It's better to modify the script in the `db_scripts` directory to dump the schema only and then fix up that resulting file. That ensures that any future changes will have simpler diffs when they are done in that way too.
See e.g. 102b97dc5a1913eb5fcc65782e952ab0ae4aa487 for the command required. Make sure to only commit relevant changes; any other local modifications should be excluded.

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsManager.java
jsharkey13: We use `MessageDigest.isEqual(..., ...)` elsewhere!

barna-isaac: Thanks for the finding, @jsharkey13! A very good catch :).

src/main/java/uk/ac/cam/cl/dtg/isaac/dto/AnvilPayloadDTO.java
barna-isaac: We return 404 bad request, and the body says "Invalid payload" (for proof, see, eg. https://github.com/isaacphysics/isaac-api/blob/feature/skills-questions-endpoint/src/test/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacadeIT.java#L157). The mapper throws `JsonProcessingException` if anything goes wrong as it tries to create the required objects, and this is caught at  https://github.com/isaacphysics/isaac-api/blob/feature/skills-questions-endpoint/src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsManager.java#L79

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    @POST
+    @Path("/{appId}/answer")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response answerQuestion(@Context final HttpServletRequest request,
+                                   @PathParam("appId") final String appId,
+                                   final AnvilMarkingRequestDTO markingRequest) {
+        if (getProperties().getProperty("SKILLS_HMAC_SECRET") == null) {
+            return new SegueErrorResponse(Status.NOT_IMPLEMENTED, "Endpoint has not been implemented.").toResponse();
```
jsharkey13:
```suggestion
            return SegueErrorResponse.getNotImplementedResponse();
```
I made this factory method specifically for this case.

jsharkey13 (approved):
Very minor, but good to go either way I hope

PR #799, opened 2026-06-12
Allow using user bookmarks list as question source

jacbn:
In doing so, reworks the ID field in the `/questions` endpoint.

- The list of question IDs to search from has been removed as a parameter (it was entirely unused beforehand, except in the ITs). 
- A new `bookmarks?: boolean` param has been added, which repurposes the _functionality_ of the old ID param to send to ES a list of content IDs – namely those that have been bookmarked – to search within. This allows us to reuse the filtering logic we have for other uses of the `/questions` endpoint inside a bookmarks search.

The test that is entirely commented out I was not sure what to do with – perhaps we can repurpose it, but given that it requires specifically the list functionality of the old ID param which has no equivalent any longer, perhaps we can also just remove it.


axlewin (approved):
LGTM 👍 

R.E. the test that was commented out: I've made some minor changes to repurpose it to test the `searchBookmarks` param instead. If you're happy with that, this can be merged.

PR #804, opened 2026-07-06
Remove unused event map data endpoint

jsharkey13:
We have not used this endpoint in some time, and we do not believe we will require it again.


sjd210 (approved):
Just removing unused code, so nothing _should_ be able to go wrong 👍

PR #808, opened 2026-07-10
Endpoint for skills on my progress

barna-isaac:
This adds an endpoint to show skills on My Progress.

It's just the one app for now, and given how close the symposium is, the following are hardcoded:
- this only ever shows attempts from the Mental Maths Practice App. The UI is hardcoded to only show this (we don't have app discovery yet), but I thought it might be a privacy concern having the endpoint return more data than is visible to users on the UI. In the privacy policy, we tell users to expect their teachers seeing the contents of their My Progress page, so I thought it important that this data is indeed what's accessible to a teacher.
- for now, always shows results from the past year. In the future, these will probably be parameters, but I didn't want to introduce them just yet (it'd be extra work validating these)
- For now, this is not cached, but if @jsharkey13 thinks this would important, I can get it done quickly!
- Not a concern for now as there's just the one app, but I wondered if @jsharkey13 has a preference whether to always return results from all apps, or for the app to be a parameter and return results just for the requested app. The second is simpler, while the first might be more performant.
---

**Pull Request Check List**
- [] Security - Data Exposure - Test any altered or created endpoints using [swagger](http://localhost:8080/isaac-api/api-docs/)
- ~[ ] Security - New dependency - configured sensibly not relying on defaults
- ~[ ] Security - New dependency - Searched for any know vulnerabilities
- ~[ ] Security - New dependency - Signed up team to mailing list
- ~[ ] Security - New dependency - Added to dependency list
- ~[ ] DB schema changes - postgres-rutherford-create-script updated
- ~[ ] DB schema changes - upgrade script created matching create script
- ~[ ] Updated Release Procedure & Documentation (& Considered Implications to Previous Versions)
- [x] Peer-Reviewed

### DIFF
--- a/src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
+++ b/src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
@@ -12,7 +12,10 @@ import uk.ac.cam.cl.dtg.isaac.dto.AnvilMarkingRequestDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.AnvilPayloadDTO;
 import uk.ac.cam.cl.dtg.isaac.dto.SegueErrorResponse;
 import uk.ac.cam.cl.dtg.isaac.dto.users.RegisteredUserDTO;
+import uk.ac.cam.cl.dtg.isaac.dto.users.UserSummaryDTO;
 import uk.ac.cam.cl.dtg.segue.api.managers.UserAccountManager;
+import uk.ac.cam.cl.dtg.segue.api.managers.UserAssociationManager;
+import uk.ac.cam.cl.dtg.segue.auth.exceptions.NoUserException;
 import uk.ac.cam.cl.dtg.segue.auth.exceptions.NoUserLoggedInException;
 import uk.ac.cam.cl.dtg.segue.dao.ILogManager;
 import uk.ac.cam.cl.dtg.segue.dao.SegueDatabaseException;
@@ -42,6 +45,8 @@ public class SkillsFacade extends AbstractIsaacFacade {
     private final UserAccountManager userManager;
     private final GitContentManager contentManager;
     private final SkillsAttemptManager skillsAttemptManager;
+    private final UserAssociationManager userAssociationManager;
+
 
     /**
      * Constructor.
@@ -49,11 +54,13 @@ public class SkillsFacade extends AbstractIsaacFacade {
     @Inject
     public SkillsFacade(final AbstractConfigLoader properties, final UserAccountManager userManager,
                         final ILogManager logManager, final GitContentManager contentManager,
-                        final SkillsAttemptManager skillsAttemptManager) {
+                        final SkillsAttemptManager skillsAttemptManager,
+                        final UserAssociationManager userAssociationManager) {
         super(properties, logManager);
         this.userManager = userManager;
         this.contentManager = contentManager;
         this.skillsAttemptManager = skillsAttemptManager;
+        this.userAssociationManager = userAssociationManager;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsAttemptManager.java
```diff
@@ -23,7 +25,7 @@
 public class SkillsAttemptManager {
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private static final int MAX_PAYLOAD_LENGTH = 30 * 1024; // about 30 kilobytes for English payloads, even Unicode
-    public static final long FIVE_MINUTES_IN_MILLIS = 300_000L;
+    public static long FIVE_MINUTES_IN_MILLIS = 300_000L;
```
jsharkey13, line 28: I think this should be `final` even if that makes testing it impossible. On the other hand, if you want a variable timeout for something, make it a final non-static argument that gets passed to the constructor? And make that default to 5 mins using a constant but then in tests replace it?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/managers/SkillsAttemptManager.java
```diff
@@ -94,4 +96,16 @@ public void recordAttempt(final List<AnvilPayloadDTO> attempt)
...
     }
+
+    /**
+     * Returns a user's attempts in the mental maths skills app.
+     *
+     * @param from - return attempts starting from this date (inclusive)
+     *
+     * @param to - return attempts until this date (exclusive)
+     */
+    public Map<LocalDate, Long> getMentalMathsAttempts(final Long userId, final LocalDate from, final LocalDate to)
```
jsharkey13, line 107: Maybe it's not worth fixing now given the tight timing, but I feel that this could have been usefully generic, as could the database layer, with the ID just hardcoded in the facade layer 🤷

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
```diff
@@ -113,4 +124,49 @@ public Response answerQuestion(@Context final HttpServletRequest request,
...
+
+            // decide if the user is allowed to view this data. If user isn't viewing their own data, user viewing
+            // must have a valid connection with the user of interest and be at least a teacher.
+            if (!currentUser.getId().equals(userIdOfInterest)
+                    && !userAssociationManager.hasTeacherPermission(currentUser, userOfInterestSummaryObject)) {
+                return SegueErrorResponse.getIncorrectRoleResponse();
+            }
+
+            HashMap<String, Map<LocalDate, Long>> resultsMap = new HashMap<>();
+            resultsMap.put("mental_maths_overall", skillsAttemptManager.getMentalMathsAttempts(
```
jsharkey13, line 156: Is it sensible to have the ID of the page-level object here? Probably this needs so much refactoring to work in the general case that it's fine now, but I was wondering how we'd go from page ID to app ID without loading data from ES unnecessarily.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgSkillsAttemptPersistenceManager.java
```diff
@@ -55,4 +62,43 @@ INSERT INTO skills_question_attempts (
...
+        try (Connection conn = database.getDatabaseConnection();
+             PreparedStatement pst = conn.prepareStatement("""
+                WITH dates(dt) AS (
+                    SELECT generate_series(?, ?, INTERVAL '1' MONTH)::DATE
+                ), attempts AS (
+                    SELECT
+                        DATE_TRUNC('month', timestamp::DATE) AS dt,
+                        COUNT(1) AS cnt
+                    FROM skills_question_attempts
+                    WHERE user_id = ? AND timestamp >= ? AND skill_id = ? AND LENGTH(question_attempt->>'result') > 0
```
jsharkey13, line 78: This query is complex!

Why do we need the length verification of attempts? I thought each attempt was stored separately, and surely we don't store empty attempts since even sending them to the API would waste bandwidth, server resources and time?

src/test/java/uk/ac/cam/cl/dtg/isaac/api/ElasticSearchTestHelper.java
```diff
@@ -43,7 +43,9 @@ public <T extends Content> T persist(final T content) throws Exception {
 
     /** Indexes a raw JSON content object with a fixed ID of {@code "i1"} and returns it. */
     public JSONObject persistJSON(final JSONObject contentJSON) throws Exception {
-        contentJSON.put("id", "i1");
+        if (!contentJSON.has("id")) {
+            contentJSON.put("id", "i1");
```
jsharkey13, line 47: Probably the JavaDoc comment above should change too?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
barna-isaac: The "full" id format `app_page_mental_maths_overall|0e184f9d-b619-4225-ac12-3c96d3c74046` contains the page-level object's id, and this is already what we store in the database. We can certainly hardcode this instead of the shorter string.

Depending on how this will work when we support multiple apps, we may or may not need to fetch these from Elasticsearch. If we always return all skill attempts, just grouping by what's in postgres, we might skip es lookups entirely.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgSkillsAttemptPersistenceManager.java
barna-isaac: @jacbn has also raised this! https://github.com/isaacphysics/isaac-api/pull/798#issuecomment-4834897740. I wanted to record empty attempts because I wanted us to later have the option to be strict about also counting these. What the strict thing is depends will depend on the target. If it's something like "number of attempts", where the target is in the numerator, the strict thing to do is to drop these. If it's something like "percentage of correct attempts", where the total number is in the denominator, the strict thing to do is to keep these.

Are you worried about how this query might perform or whether the empty attempts will take up too much space in the database?

src/main/java/uk/ac/cam/cl/dtg/isaac/api/SkillsFacade.java
barna-isaac: We've agreed to come back to this once we have more apps.

src/main/java/uk/ac/cam/cl/dtg/isaac/dao/PgSkillsAttemptPersistenceManager.java
barna-isaac: @jsharkey13 and I have agreed to stop recording empty attempts. The Anvil app should stop sending these, and the API shouldn't allow any. Recording empty attempts is not an effective way to protect against a user that only wants to submit easy answers, as they could just refresh the page.

PR #814, opened 2026-07-29
Restore random seed functionality

axlewin:
Only add the new tie-break sort order clauses if there is no random seed set, otherwise it should be null (as before).


barna-isaac (approved):
Looks good, thanks for the fix.

